PR #29, opened 2019-12-23
start done

oksanaDaniluk:
https://oksanaDaniluk.github.io/layout_stars/
https://oksanaDaniluk.github.io/layout_stars/report/html_report/

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -8,6 +8,47 @@
   <link rel="stylesheet" href="./style.css">
 </head>
 <body>
-  <h1>Stars</h1>
+  <div class="stars--0 general block">
+    <div class="star"></div>
+    <div class="star"></div>
+    <div class="star"></div>
+    <div class="star"></div>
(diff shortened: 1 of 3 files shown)


src/index.html
```diff
@@ -8,6 +8,47 @@
   <link rel="stylesheet" href="./style.css">
 </head>
 <body>
-  <h1>Stars</h1>
+  <div class="stars--0 general block">
```
earthhuman41: по бему "stars stars--0"

src/index.html
```diff
@@ -8,6 +8,47 @@
...
+    <div class="star"></div>
```
earthhuman41: stars__star

src/index.html
oksanaDaniluk: ок, переробляю. Робила по тз - Add 5 more blocks with a modifier per block from `stars--1` to `stars--5

PR #48, opened 2020-02-28
Solution

KostyaPenyavskiy:
* [DEMO LINK](https://KostyaPenyavskiy.github.io/layout_stars/) <br>
* [TEST REPORT LINK](https://KostyaPenyavskiy.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,28 @@
+body {
+  margin: 0;
+}
+
+.stars__icon {
+  display: flex;
+}
+
+.stars__icon div {
+  background-image: url("./images/star.svg");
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,28 @@
+body {
+  margin: 0;
+}
+
+.stars__icon {
+  display: flex;
+}
+
+.stars__icon div {
```
vkryvytskyy: better add class to div inside stars_icon

vkryvytskyy: than styling tag

PR #93, opened 2020-03-09
Develop

coastin:
- [DEMO LINK](https://coastin.github.io/layout_stars/) <br>
    - [TEST REPORT LINK](https://coastin.github.io/layout_stars/report/html_report/)


vkryvytskyy:
image
links are not clickable

src/style.css
```diff
@@ -0,0 +1,24 @@
...
+.stars-block {
+  background-image: url("images/star.svg");
+  background-repeat: no-repeat;
+  background-position: center;
+  width: 15px;
+  height: 15px;
+  margin-right: 4px;
+}
+  
+.stars-first div:nth-child(-n+1),
```
vkryvytskyy:
```suggestion
.stars-first .stars-block:nth-child(-n+1),
```
Try to avoid stylizing tags. Always use classes for this. Tags stylizing is OK when you reset default browser styles, like margin: 0 for BODY tag, which has 8px margins. Setting width and height for all DIV tags is bad practice

PR #144, opened 2020-07-05
soluted, but there are question

Andrii256:
- [DEMO LINK](https://andrii256.github.io/layout_stars/) <br>
- [TEST REPORT LINK](https://andrii256.github.io/layout_stars/report/html_report/)

question: why we must use n-th child if can just use `` and `` ?

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,41 @@
+body {
+  margin: 0;
+}
+
+.stars {
+  display: flex;
+}
+
+.stars__star {
+  --side: 15px;
+  height: var(--side);
+  width: var(--side);
+  margin-right: 4px;
+  background: url(images/star.svg) center no-repeat;
+}
+
+.stars--1 .stars__star:nth-child(1) {
+  background-image: url(images/star-active.svg);
+}
+
+.stars--2 .stars__star:nth-child(1),
+.stars--2 .stars__star:nth-child(2) {
+  background-image: url(images/star-active.svg);
+}
+
+.stars--3 .stars__star:nth-child(1),
+.stars--3 .stars__star:nth-child(2),
+.stars--3 .stars__star:nth-child(3) {
+  background-image: url(images/star-active.svg);
+}
+
+.stars--4 .stars__star:nth-child(1),
+.stars--4 .stars__star:nth-child(2),
+.stars--4 .stars__star:nth-child(3),
+.stars--4 .stars__star:nth-child(4) {
+  background-image: url(images/star-active.svg);
+}
+
+.stars--5 .stars__star:nth-child(1n) {
+  background-image: url(images/star-active.svg);
+}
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,41 @@
...
+}
+
+.stars--2 .stars__star:nth-child(1),
+.stars--2 .stars__star:nth-child(2) {
+  background-image: url(images/star-active.svg);
+}
+
+.stars--3 .stars__star:nth-child(1),
+.stars--3 .stars__star:nth-child(2),
+.stars--3 .stars__star:nth-child(3) {
```
vkryvytskyy, line 28: зображення

src/style.css
```diff
@@ -0,0 +1,41 @@
...
+.stars--4 .stars__star:nth-child(1),
+.stars--4 .stars__star:nth-child(2),
+.stars--4 .stars__star:nth-child(3),
+.stars--4 .stars__star:nth-child(4) {
+  background-image: url(images/star-active.svg);
+}
+
+.stars--5 .stars__star:nth-child(1n) {
+  background-image: url(images/star-active.svg);
+}
```
vkryvytskyy, line 41: this are the same rules, write it like that:

```
 .stars--4 .stars__star:nth-child(-n+),
 .stars--5 .stars__star:nth-child(-n+) {
}
```

vkryvytskyy:
> question: why we must use n-th child if can just use  and  ?

answer: cause images that has no sense(lika rating stars) should be placed on page through css. Such way screenreader won't focus on them, and don't try to read images with no sense.

src/style.css
Andrii256: fixed

src/style.css
Andrii256: fixed

src/style.css
```diff
@@ -14,28 +14,10 @@ body {
...
-
-.stars--4 .stars__star:nth-child(1),
-.stars--4 .stars__star:nth-child(2),
-.stars--4 .stars__star:nth-child(3),
-.stars--4 .stars__star:nth-child(4) {
-  background-image: url(images/star-active.svg);
-}
-
-.stars--5 .stars__star:nth-child(1n) {
+.stars--1 .stars__star:nth-child(1),
```
vkryvytskyy, line 17:
```suggestion
.stars--1 .stars__star:first-child,
```

src/style.css
```diff
@@ -14,28 +14,10 @@ body {
...
+.stars--2 .stars__star:nth-child(-n+2),
+.stars--3 .stars__star:nth-child(-n+3),
+.stars--4 .stars__star:nth-child(-n+4),
+.stars--5 .stars__star:nth-child(-n+5) {
```
vkryvytskyy, line 21:
```suggestion
.stars--5 .stars__star {
```
since there only five stars - no need to provide `nth-child` in this case

PR #146, opened 2020-07-06
solution

ArtemiiKab:
- [DEMO LINK](https://ArtemiiKab.github.io/layout_stars/) <br>
  - [TEST REPORT LINK](https://ArtemiiKab.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,23 @@
+body {
+  margin: 0;
+}
+
+.stars {
+  height: 15px;
+}
+
+.stars_item {
+  display: inline-block;
+  width: 13px;
+  height: 12px;
+  margin: 0 1px;
+  content: url(images/star.svg);
+}
+
+.stars--1 :nth-child(-n + 1),
+.stars--2 :nth-child(-n + 2),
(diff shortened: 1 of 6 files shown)


src/style.css
```diff
@@ -0,0 +1,23 @@
...
+
+.stars_item {
+  display: inline-block;
+  width: 13px;
+  height: 12px;
+  margin: 0 1px;
+  content: url(images/star.svg);
+}
+
+.stars--1 :nth-child(-n + 1),
```
vkryvytskyy, line 17:
```suggestion
.stars--1 :first-child,

```

src/style.css
```diff
@@ -0,0 +1,23 @@
...
+.stars--2 :nth-child(-n + 2),
+.stars--3 :nth-child(-n + 3),
+.stars--4 :nth-child(-n + 4),
+.stars--5 :nth-child(-n + 5) {
```
vkryvytskyy, line 21:
```suggestion
.stars--5 .stars-item {
```
since there is only five stars, every div in this block should be styled

PR #166, opened 2020-07-07
Develop

DmJSCourse:
[DEMO LINK](https://dmjscourse.github.io/layout_stars/)
[TEST REPORT LINK](https://dmjscourse.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,30 @@
+body {
+  margin: 0;
+}
+
+.stars {
+  display: flex;
+  align-items: center;
+}
+
+.stars--0 div,
+.stars--1 div,
+.stars--2 div,
+.stars--3 div,
+.stars--4 div,
+.stars--5 div {
+  height: 15px;
+  width: 15px;
+  background-image: url(images/star.svg);
+  background-repeat: no-repeat;
+  background-position: center;
+  margin-right: 4px;
+}
+
+/* .stars--1 > div:nth-child(1),
+.stars--2 > :nth-child(-n+2),
+.stars--3 > :nth-child(-n+3),
+.stars--4 > :nth-child(-n+4),
+.stars--5 {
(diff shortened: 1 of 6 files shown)


src/style.css
```diff
@@ -0,0 +1,30 @@
+body {
+  margin: 0;
+}
+
+.stars div {
+  display: flex;
+}
+
+.stars--0 div,
```
vkryvytskyy, line 9: add class, don't style div's 
`.stars__item` maybe

src/style.css
```diff
@@ -0,0 +1,30 @@
...
+.stars--5 div {
+  height: 15px;
+  width: 15px;
+  background-image: url(images/star.svg);
+  background-repeat: no-repeat;
+  background-position: center;
+  margin-right: 4px;
+}
+
+.stars--1 > :nth-child(1),
```
vkryvytskyy, line 23:
```suggestion
.stars--1 > stars__item:first-child,
```

src/style.css
```diff
@@ -0,0 +1,30 @@
...
+.stars--2 > :nth-child(-n+2),
+.stars--3 > :nth-child(-n+3),
+.stars--4 > :nth-child(-n+4),
+.stars--5 div {
+  background-image: url("images/star-active.svg");
+  background-repeat: no-repeat;
```
vkryvytskyy, line 29: combine into one rule

src/style.css
DmJSCourse: >Each block should have a given number of active stars without adding any special class for an individual star. 
https://github.com/mate-academy/layout_stars

vsm2k17: it means not to set different classes for stars, but 1 for all needs to be created. because  styling divs that's a bad idea

PR #203, opened 2020-08-31
finished

Sadman9973:
- [DEMO LINK](https://Sadman9973.github.io/layout_stars/)
- [TEST REPORT LINK](https://Sadman9973.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -8,6 +8,47 @@
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Stars</h1>
+    <div class="starRow stars--0">
+      <div class="star"></div>
+      <div class="star"></div>
+      <div class="star"></div>
+      <div class="star"></div>
+      <div class="star"></div>
+    </div>
+    <div class="starRow stars--1">
+      <div class="star"></div>
+      <div class="star"></div>
+      <div class="star"></div>
+      <div class="star"></div>
+      <div class="star"></div>
+    </div>
+    <div class="starRow stars--2">
+      <div class="star"></div>
+      <div class="star"></div>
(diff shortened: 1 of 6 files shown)


src/index.html
```diff
@@ -8,6 +8,47 @@
...
   <body>
-    <h1>Stars</h1>
+    <div class="starRow stars--0">
+      <div class="star"></div>
+      <div class="star"></div>
+      <div class="star"></div>
+      <div class="star"></div>
+      <div class="star"></div>
+    </div>
+    <div class="starRow stars--1">
```
YuriiHoliuk, line 18: add empty lines between blocks of HTML

src/index.html
```diff
@@ -8,6 +8,47 @@
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Stars</h1>
+    <div class="starRow stars--0">
```
YuriiHoliuk, line 11: classes should be in kebab-case or BEM, not in camelCase

YuriiHoliuk: BEM modifier should only be used on element which has the corresponding block class

src/style.css
```diff
@@ -0,0 +1,26 @@
...
+}
+
+.star {
+  width: 15px;
+  height: 15px;
+  margin-right: 4px;
+  background: url("images/star-active.svg") no-repeat 50%;
+}
+
+.stars--0 .star,
```
YuriiHoliuk, line 19: BEM block(`stars`) shouldn't style another BEM block (`star`)

YuriiHoliuk (changes requested):

PR #220, opened 2020-09-01
Stars task

yehor1959:
- [DEMO LINK](https://yehor1959.github.io/layout_stars/)
- [TEST REPORT LINK](https://yehor1959.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -8,6 +8,51 @@
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Stars</h1>
+    <div class="columns stars--0">
+      <div class="columns_part"></div>
+      <div class="columns_part"></div>
+      <div class="columns_part"></div>
+      <div class="columns_part"></div>
+      <div class="columns_part"></div>
+    </div>
+    <div class="columns stars--1">
+      <div class="columns_part"></div>
+      <div class="columns_part"></div>
+      <div class="columns_part"></div>
+      <div class="columns_part"></div>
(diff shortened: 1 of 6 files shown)


src/style.css
```diff
@@ -0,0 +1,26 @@
...
+}
+
+.columns_part {
+  background: center url(images/star.svg) no-repeat;
+  width: 15px;
+  height: 15px;
+  margin-right: 4px;
+}
+
+.stars--1 .columns_part:first-child,
```
YuriiHoliuk, line 20: one BEM block shouldn't style another BEM block
Checklist

src/style.css
```diff
@@ -0,0 +1,26 @@
+* {
+  box-sizing: border-box;
+}
+
+body {
+  margin: 0;
+}
+
+.columns {
```
YuriiHoliuk, line 9: bad classname here

YuriiHoliuk (changes requested):

src/index.html
```diff
@@ -8,6 +8,51 @@
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Stars</h1>
+    <div class="stars-block stars--0">
```
YuriiHoliuk, line 11: BEM modifier should only be used on element which has the corresponding block class
Checklist

YuriiHoliuk (changes requested):

YuriiHoliuk (approved):
👍

PR #236, opened 2020-09-03
stars

oleksiiKondrashov:
- [DEMO LINK](https://oleksiiKondrashov.github.io/layout_stars/)
- [TEST REPORT LINK](https://oleksiiKondrashov.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -8,6 +8,52 @@
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Stars</h1>
+    <div class="row stars--0">
+      <div class="row__star"></div>
+      <div class="row__star"></div>
+      <div class="row__star"></div>
+      <div class="row__star"></div>
+      <div class="row__star"></div>
+    </div>
+
+    <div class="row stars--1">
+      <div class="row__star"></div>
+      <div class="row__star"></div>
+      <div class="row__star"></div>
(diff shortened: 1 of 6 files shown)


src/index.html
```diff
@@ -8,6 +8,52 @@
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Stars</h1>
+    <div class="row stars--0">
```
YuriiHoliuk, line 11: you cannot use modifier without block class
Checklist

YuriiHoliuk (changes requested):

src/index.html
```diff
@@ -8,52 +8,52 @@
...
+      <div class="rating__star"></div>
     </div>
 
-    <div class="row stars--1">
-      <div class="row__star"></div>
-      <div class="row__star"></div>
-      <div class="row__star"></div>
-      <div class="row__star"></div>
-      <div class="row__star"></div>
+    <div class="rating stars--1">
```
YuriiHoliuk, line 19: you cannot use block modifier without corresponding block

YuriiHoliuk (changes requested):

src/index.html
oleksiiKondrashov: i don't understand my mistake about modifier(

PR #241, opened 2020-11-03
fixed

Layrovell:
[DEMO LINK](https://Layrovell.github.io/layout_stars/)
   [TEST REPORT LINK](https://Layrovell.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -4,10 +4,58 @@
     <meta charset="UTF-8">
     <meta name="viewport"
           content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0">
-    <title>Product cards</title>
+    <title>Stars</title>
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Stars</h1>
+
+    <div class="stars--0">
+      <div class="star__primary"></div>
+      <div class="star__primary"></div>
+      <div class="star__primary"></div>
+      <div class="star__primary"></div>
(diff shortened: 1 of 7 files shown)


src/index.html
```diff
@@ -4,10 +4,58 @@
...
           content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0">
-    <title>Product cards</title>
+    <title>Stars</title>
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Stars</h1>
+
+    <div class="stars--0">
+      <div class="star__primary"></div>
```
irenhh, line 13: again wrong BEM
https://en.bem.info/methodology/quick-start/
read about blocks and elements
you can't use element name with `star` prefix since you don't have a parent block class `star`

src/style.css
```diff
@@ -0,0 +1,32 @@
+* {
+  margin: 0;
+}
+
+.stars--0,
+.stars--1,
+.stars--2,
+.stars--3,
+.stars--4,
+.stars--5 {
```
irenhh, line 10: there could be a general class, one for each similar element

irenhh (changes requested):

src/index.html
Layrovell: fixed

src/style.css
Layrovell: fixed

irenhh (approved):

PR #357, opened 2021-03-21
stars all tests passed

Burunduk003:
- [DEMO LINK](https://Burunduk003.github.io/layout_stars/)
- [TEST REPORT LINK](https://Burunduk003.github.io/layout_stars/report/html_report/)


aionova (changes requested):
Screenshot 2021-03-22 at 15 59 09
You don't need to hardcode the number of active stars. Each block of stars should have a modifier '.stars--1 ', '.stars--2' etc., and if you change the class, the number of active stars should change.

src/style.css
```diff
@@ -0,0 +1,39 @@
...
+  background-image: url(images/star-active.svg);
+}
+
+.stars--4 .stars__star:nth-child(-n + 4) {
+  background-image: url(images/star-active.svg);
+}
+
+.stars--5 .stars__star:nth-child(n) {
+  background-image: url(images/star-active.svg);
+}
```
aionova, line 39:
```suggestion
.stars--1 :first-child,
.stars--2 :nth-child(-n+2),
.stars--3 :nth-child(-n+3),
.stars--4 :nth-child(-n+4),
.stars--5 :nth-child(n) {
  background-image: url(images/star-active.svg);
}
```

aionova: looks better, doesn't it?

aionova (changes requested):

aionova (approved):
Awesome!

PR #358, opened 2021-03-22
Solution

IvanGrekov:
DEMO LINK - https://IvanGrekov.github.io/layout_stars/
TEST REPORT LINK - https://IvanGrekov.github.io/layout_stars/report/html_report/

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -8,6 +8,136 @@
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Stars</h1>
+    <div class="stars stars--0">
+      <div class="stars__star">
+        <img src="./images/star.svg" alt="Star">
+      </div>
+
+      <div class="stars__star">
+        <img src="./images/star.svg" alt="Star">
+      </div>
+
+      <div class="stars__star">
+        <img src="./images/star.svg" alt="Star">
+      </div>
+
+      <div class="stars__star">
+        <img src="./images/star.svg" alt="Star">
+      </div>
+
+      <div class="stars__star">
+        <img src="./images/star.svg" alt="Star">
+      </div>
+    </div>
+
+    <div class="stars stars--1">
+      <div class="stars__star">
+        <img src="./images/star-active.svg" alt="Active-star">
+      </div>
+
+      <div class="stars__star">
+        <img src="./images/star.svg" alt="Star">
+      </div>
+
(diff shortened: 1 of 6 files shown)


VinnieJ-2k20 (changes requested):
Result is correct, but solution is not.

You need to have six blocks of stars with only one difference between them - modifier stars--0, stars--1 etc.
All single star elements should have the same class.
You should style the stars using CSS depending on parent block modifier.
Little hint:

VinnieJ-2k20 (changes requested):
You are still using different images for your star elements. They should be the same.
Try using `div` elements with background image instead of `img`.
https://www.w3schools.com/cssref/pr_background-image.asp
And then change the background image when necessary depending on parent block modifier (`stars--0`, `stars--1` etc.)

src/index.html
```diff
@@ -8,6 +8,76 @@
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Stars</h1>
+    <div class="stars--0">
+      <div class="stars__star"></div>
+
+      <div class="stars__star"></div>
```
VinnieJ-2k20, line 14: Structure is not overloaded with text, so we don't have to add empty line between sibling elements here.

src/style.css
```diff
@@ -0,0 +1,28 @@
...
+  box-sizing: border-box;
+}
+
+body {
+  margin: 0;
+}
+
+div:nth-child(n) {
+  display: flex;
+}
```
VinnieJ-2k20, line 11: Don't style using tags, use classes.

src/style.css
```diff
@@ -0,0 +1,28 @@
...
+  background-position: 1px 2px;
+}
+
+.stars--1 div:nth-child(-n + 1),
+.stars--2 div:nth-child(-n + 2),
+.stars--3 div:nth-child(-n + 3),
+.stars--4 div:nth-child(-n + 4),
+.stars--5 div:nth-child(n) {
+  background-image: url(images/star-active.svg);
+}
```
VinnieJ-2k20, line 28: Don't style using tags, use classes.

VinnieJ-2k20 (changes requested):
Almost there.
Need to fix some things - details in comments.

VinnieJ-2k20 (approved):
Good job 👍

PR #360, opened 2021-03-22
Make layout

victor-arutiunov:
- [DEMO LINK](https://victor-arutiunov.github.io/layout_stars/) <br>
- [TEST REPORT LINK](https://victor-arutiunov.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,24 @@
+body {
+  margin: 0;
+}
+
+.stars {
+  width: 100px;
+  height: 15px;
+  display: flex;
+  justify-content: space-between;
+}
+
+.stars__star {
+  object-fit: contain;
+  object-position: center;
+  width: 13px;
+}
+
+.stars :last-of-type {
+  margin-right: 10px;
+}
+
+.stars :first-of-type {
+  margin-left: 1px;
+}
--- a/.bemlintrc.json
+++ b/.bemlintrc.json
@@ -1,0 +1,16 @@
+{
+  "elementDivider": "__",
+  "modifierDivider": "_",
+  "ignore": [
+    "node_modules",
+    "dist"
+  ],
+  "rules": {
+    "one-block": true,
+    "one-element": true,
+    "element-inside-parent-block": true,
+    "no-double-element": true,
+    "no-neighbour-parent-block": true,
+    "modifiable-class": true
+  }
+}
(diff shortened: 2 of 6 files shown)


VinnieJ-2k20 (changes requested):
Result is correct, but solution is not.

You need to have six blocks of stars with only one difference between them - modifier stars--0, stars--1 etc.
All single star elements should have the same class.
You should style the stars using CSS depending on parent block modifier.
Little hint: 

src/style.css
```diff
@@ -0,0 +1,48 @@
...
+  height: 15px;
+  display: flex;
+  justify-content: space-between;
+}
+
+.stars > * {
+  width: 13px;
+  background-repeat: no-repeat;
+  background-position: center;
+}
```
VinnieJ-2k20, line 16: It is better not to use `*` selector for anything other than setting border-box property.
You have `stars_star` class. Use it here.

src/style.css
```diff
@@ -0,0 +1,48 @@
...
+  background-image: url("images/star.svg");
+}
+
+.stars--3 :nth-child(n+4) {
+  background-image: url("images/star.svg");
+}
+
+.stars--4 :nth-child(n+5) {
+  background-image: url("images/star.svg");
+}
```
VinnieJ-2k20, line 40: You can combine it like this: 
```suggestion
.stars--0 :nth-child(n),
.stars--1 :nth-child(n+2),
.stars--2 :nth-child(n+3),
.stars--3 :nth-child(n+4),
.stars--4 :nth-child(n+5) {
  background-image: url("images/star.svg");
}
```

src/style.css
```diff
@@ -0,0 +1,48 @@
...
+
+.stars > * {
+  width: 13px;
+  background-repeat: no-repeat;
+  background-position: center;
+}
+
+.stars :nth-child(n) {
+  background-image: url("images/star-active.svg");
+}
```
VinnieJ-2k20, line 20: You can move it to `stars_star` class

VinnieJ-2k20 (changes requested):
Almost there.
Need to fix couple of things. Details in comments.

Also, it is not crucial, but the logic is reversed.
Default star should be inactive and then changed to active if necessary.
You can also change it, but it's not necessary.

VinnieJ-2k20 (approved):
Good job 👍

PR #364, opened 2021-03-22
Develop

herkoss:
https://herkoss.github.io/layout_stars/
https://herkoss.github.io/layout_stars/report/html_report/

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,36 @@
+body {
+  margin: 0;
+}
+
+.stars {
+  display: flex;
+}
+
+.stars__star {
+  height: 15px;
+  width: 15px;
+  margin-right: 4px;
+  background-image: url("images/star.svg");
+  background-repeat: no-repeat;
+  background-position: center;
+}
+
+.stars--1 :nth-child(-n+1) {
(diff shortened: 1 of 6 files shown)


src/style.css
```diff
@@ -0,0 +1,36 @@
...
+  background-image: url("images/star-active.svg");
+}
+
+.stars--4 :nth-child(-n+4) {
+  background-image: url("images/star-active.svg");
+}
+
+.stars--5 :nth-child(-n+5) {
+  background-image: url("images/star-active.svg");
+}
```
VinnieJ-2k20, line 36: You can combine it like this:
```suggestion
.stars--1 :nth-child(-n+1),
.stars--2 :nth-child(-n+2),
.stars--3 :nth-child(-n+3),
.stars--4 :nth-child(-n+4),
.stars--5 :nth-child(-n+5) {
  background-image: url("images/star-active.svg");
}
```

VinnieJ-2k20 (approved):
Good job 👍 
One remark is in the comment.

src/style.css
herkoss: Ok, understood

PR #373, opened 2021-03-22
Solution

Taras-Konovalov:
[DEMO LINK](https://Taras-Konovalov.github.io/layout_stars/)
[TEST REPORT LINK](https://Taras-Konovalov.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,28 @@
+* {
+  margin: 0;
+}
+
+body {
+  box-sizing: border-box;
+}
+
+.container {
+  display: flex;
+}
+
+.stars__star {
+  height: 15px;
+  width: 15px;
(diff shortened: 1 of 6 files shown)


src/style.css
```diff
@@ -0,0 +1,28 @@
+* {
+  margin: 0;
```
aionova, line 2: Set box-sizing: border-box to all the elements with * selector, and don't set margins in * selector.
In the future, we will use reset.css for that. For now just set it for all the classes separately where you need it.

aionova (changes requested):
Good job! Almost there!

aionova (approved):
Nice!
Just don't set margin: 0 for all the elements pls. Use "*" selector for box-sizing only

PR #376, opened 2021-03-23
Develop

J-o-n-at-h-a-n:
[DEMO LINK](https://J-o-n-at-h-a-n.github.io/layout_stars/)
[TEST REPORT LINK](https://J-o-n-at-h-a-n.github.io/layout_stars/report/html_report/)
Тесты не проходят, как и во втором задании

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,57 @@
+.star {
+  height: 12px;
+  width: 13px;
+  margin-right: 6px;
+  margin-bottom: 3px;
+  background-size: 100%;
+}
+
+.stars {
+  display: flex;
+}
+
+.stars--0 .star:nth-child(n) {
+  background-image: url("images/star.svg");
+}
+
+.stars--1 :nth-child(n + 1) {
+  background-image: url("images/star.svg");
+}
+
+.stars--1 :nth-child(1) {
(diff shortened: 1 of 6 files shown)


VinnieJ-2k20 (changes requested):
You can make your styles with `nth-child` much more simple.
You'll find hints here -> http://nthmaster.com/

src/style.css
```diff
@@ -0,0 +1,51 @@
...
+  background-image: url("images/star.svg");
+}
+
+.stars--4 :nth-child(-n +  4) {
+  background-image: url("images/star-active.svg");
+}
+
+.stars--5 :nth-child(n) {
+  background-image: url("images/star-active.svg");
+}
```
VinnieJ-2k20, line 51: Let's make it even less complicated.

Apply background with empty star on default and then change it to active when necessary.
Also, you can combine selectors with same styles using `,`
```
selector1,
selector2,
selector3 {
  ...styles
}
```

VinnieJ-2k20 (changes requested):
You are on the right way.

Let's fix one more thing. Details in comment.

VinnieJ-2k20 (approved):
Good job 👍

PR #377, opened 2021-03-23
Stars-001

provolovskyi-illia:
- [DEMO LINK](https://provolovskyi-illia.github.io/layout_stars/)
- [TEST REPORT LINK](https://provolovskyi-illia.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,40 @@
+* {
+  margin: 0;
+}
+
+.stars {
+  display: flex;
+}
+
+.stars--0 :nth-child(n) {
+  background-image: url("images/star.svg");
+}
+
+.stars--1 :nth-child(n+2) {
+  background-image: url("images/star.svg");
+}
+
+.stars--2 :nth-child(n+3) {
+  background-image: url("images/star.svg");
+}
+
+.stars--3 :nth-child(n+4) {
+  background-image: url("images/star.svg");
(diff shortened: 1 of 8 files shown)


src/style.css
```diff
@@ -0,0 +1,40 @@
+* {
```
aliceblunt8, line 1: disable margins only for body

src/style.css
```diff
@@ -0,0 +1,40 @@
...
+  background-image: url("images/star.svg");
+}
+
+.stars--4 :nth-child(n+5) {
+  background-image: url("images/star.svg");
+}
+
+.stars--5 :nth-child(n+6) {
+  background-image: url("images/star.svg");
+}
```
aliceblunt8, line 31: combine this into one rule
like this

```suggestion
.stars--1 :nth-child(n+2),
.stars--2 :nth-child(n+3) {
  background-image: url("images/star.svg");
}
```

aliceblunt8 (changes requested):

aliceblunt8 (approved):
also, .stars-star styles should be above .stars--0 :nth-childs styles, read about css cascase and inheritance
https://developer.mozilla.org/en-US/docs/Learn/CSS/Building_blocks/Cascade_and_inheritance

PR #387, opened 2021-03-23
ready for cheсk

Bogdan-Kotsupey:
[DEMO LINK](https://Bogdan-Kotsupey.github.io/layout_stars/)

[TEST REPORT LINK](https://Bogdan-Kotsupey.github.io/layout_stars/report/html_report/)


aionova (changes requested):
Nice start!
Try to use nth-child(-n + smth) here and shorten your code.

PR #413, opened 2021-03-24
solution v1.0

IvanKolodiazhnyi:
- [DEMO LINK](https://IvanKolodiazhnyi.github.io/layout_stars/)
- [TEST REPORT LINK](https://IvanKolodiazhnyi.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,31 @@
+body {
+  margin: 0;
+}
+
+.stars_star {
+  width: 15px;
+  height: 15px;
+  background-position: center;
+  background-repeat: no-repeat;
+  margin-right: 4px;
+}
+
+.box {
+  display: flex;
+}
+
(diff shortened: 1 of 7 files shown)


src/style.css
```diff
@@ -0,0 +1,31 @@
...
+.box {
+  display: flex;
+}
+
+.stars--0 :nth-child(n),
+.stars--1 :nth-child(n + 2),
+.stars--2 :nth-child(n + 3),
+.stars--3 :nth-child(n + 4),
+.stars--4 :nth-child(n + 5) {
+  background-image: url(images/star.svg);
```
aionova, line 22: You could set default background-image for all the stars to be images/star.svg, and then set background for active stars separately. But your option works too

aionova (approved):
Awesome!

PR #420, opened 2021-03-24
pending remarks

amithos:
- [DEMO LINK](https://amithos.github.io/layout_stars/)
- [TEST REPORT LINK](https://amithos.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,38 @@
+* {
+  all: unset;
+  display: revert;
+}
+
+.stars {
+  display: flex;
+}
+
+.stars__star {
+  width: 15px;
+  height: 15px;
+  margin-right: 4px;
+  background-image: url("images/star-active.svg");
+  background-repeat: no-repeat;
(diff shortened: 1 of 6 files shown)


src/style.css
```diff
@@ -0,0 +1,38 @@
...
+  background-image: url("images/star.svg");
+}
+
+.stars--3 .stars__star:nth-child(n+4) {
+  background-image: url("images/star.svg");
+}
+
+.stars--4 .stars__star:nth-child(n+5) {
+  background-image: url("images/star.svg");
+}
```
VinnieJ-2k20, line 38: You can combine selectors with same styles, like this:

```
selector1,
selector2,
selector3 {
  ...styles
}
```

VinnieJ-2k20 (changes requested):
Almost done, one small remark in comments.
Also, let's fix the demo link and test link, they show page 404.

PR #428, opened 2021-03-25
Stars

BoikoEva:
- [DEMO LINK](https://boikoeva.github.io/layout_stars/)
- [TEST REPORT LINK](https://boikoeva.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,3 @@
+html {
+  margin: 0 auto;
+}
--- a/src/style.css
+++ b/src/style.css
@@ -1,3 +1,24 @@
-html {
-  margin: 0 auto;
+body {
+  margin: 0;
+  padding: 0;
+}
+
+.star {
+  display: flex;
+}
+
+.stars__star {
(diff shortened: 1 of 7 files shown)


src/style.css
```diff
@@ -0,0 +1,24 @@
...
+.star {
+  display: flex;
+}
+
+.stars__star {
+  width: 15px;
+  height: 15px;
+  margin-right: 4px;
+  background-image: url("./images/star.svg");
+  background-repeat: no-repeat;
```
VinnieJ-2k20, line 15: Try adding `background-position: center;` here.

VinnieJ-2k20 (changes requested):
Almost there.
One remark in the comments. Tests should pass after that.

src/style.css
BoikoEva: Done! Thank you:)

VinnieJ-2k20 (approved):
Good job!

PR #616, opened 2021-07-07
stars

TarasHoliuk:
- [DEMO LINK](https://TarasHoliuk.github.io/layout_stars/)
- [TEST REPORT LINK](https://TarasHoliuk.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,41 @@
+body {
+  margin: 0;
+}
+
+.stars {
+  display: flex;
+  flex-direction: row-reverse;
+  justify-content: center;
+  align-items: center;
+}
+
+.stars__star {
+  height: 16px;
+  width: 16px;
+  padding: 2px;
+  background-repeat: no-repeat;
+}
+
+.stars .stars__star {
+  background-image: url(./images/star.svg);
+}
+
+.stars--1 .stars__star:nth-child(n+5) {
+  background-image: url(./images/star-active.svg);
+}
+
+.stars--2 .stars__star:nth-child(n+4) {
+  background-image: url(./images/star-active.svg);
+}
+
+.stars--3 .stars__star:nth-child(n+3) {
+  background-image: url(./images/star-active.svg);
+}
+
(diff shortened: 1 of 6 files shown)


src/style.css
```diff
@@ -0,0 +1,41 @@
...
+  background-image: url(./images/star-active.svg);
+}
+
+.stars--4 .stars__star:nth-child(n+2) {
+  background-image: url(./images/star-active.svg);
+}
+
+.stars--5 .stars__star:nth-child(n+1) {
+  background-image: url(./images/star-active.svg);
+}
```
alex-glotov67, line 41: All good, but you can make it shorter by using , like this:
.stars--3 .stars__star:nth-child(n+3),
.stars--4 .stars__star:nth-child(n+2) {
}

It will be same but shorter)

alex-glotov67 (approved):
All great! 
I left some advice for you =)

src/style.css
TarasHoliuk: thank you, I'll fix it )

src/style.css
```diff
@@ -0,0 +1,30 @@
+body {
+  margin: 0;
+}
+
+.stars {
+  display: flex;
+  flex-direction: row-reverse;
+  justify-content: center;
+  align-items: center;
```
danheim, line 9: avoid row reverse, it looks weird, you don't need id

danheim: remove justify content and align items

danheim (changes requested):

src/style.css
```diff
@@ -0,0 +1,27 @@
...
+  height: 16px;
+  width: 16px;
+  padding: 2px;
+  background-repeat: no-repeat;
+  background-position: center;
+}
+
+.stars .stars__star {
+  background-image: url(./images/star.svg);
+}
```
danheim, line 19: почему это написано в разных селекторах?

danheim (changes requested):

danheim (approved):

PR #628, opened 2021-07-08
Develop

vova-poli:
- [DEMO LINK](https://vova-poli.github.io/layout_stars/)
- [TEST REPORT LINK](https://vova-poli.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,25 @@
+body {
+  margin: 0;
+}
+
+.stars {
+  display: flex;
+  height: 16px;
+}
+
+.stars__star {
+  background-image: url(images/star.svg);
+  display: inline-block;
+  background-repeat: no-repeat;
+  width: 16px;
+  height: 16px;
+  margin: 2px;
+}
+
+.stars--1 > :first-child,
+.stars--2 > :nth-child(-n+2),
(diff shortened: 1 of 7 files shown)


src/style.css
```diff
@@ -0,0 +1,25 @@
...
+}
+
+.stars {
+  display: flex;
+  height: 16px;
+}
+
+.stars__star {
+  background-image: url(images/star.svg);
+  display: inline-block;
```
varseniuk, line 12:
```suggestion
```
In flex container all items are inline-blocks

src/style.css
```diff
@@ -0,0 +1,25 @@
+body {
+  margin: 0;
+}
+
+.stars {
+  display: flex;
+  height: 16px;
```
varseniuk, line 7:
```suggestion
```
No need for this. You have given the heigh for your `star` element

varseniuk (approved):
good job! Left some comments

src/style.css
vova-poli: From the very beginning I tried not to use "height: 16" but my test showed an error http://joxi.ru/ZrJK3jDhbNJQlm. So then I took this solution from another students repository and somehow it got fixed

PR #634, opened 2021-07-09
stars

karolina-nad:


herkoss:
Fix please)

Снимок экрана 2021-07-09 в 15 27 41

src/index.html
```diff
@@ -8,6 +8,47 @@
...
   <body>
-    <h1>Stars</h1>
+    <div class="stars--0 stars">
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+    </div>
+    <div class="stars--1 stars">
```
VinnieJ-2k20, line 18:
```suggestion
    </div>
    
    <div class="stars--1 stars">
```

src/style.css
```diff
@@ -0,0 +1,28 @@
+body {
+  margin: 0;
+}
+
+.stars {
+  display: flex;
+  height: 16px;
+  width: 96px;
```
VinnieJ-2k20, line 8: No real need to set fixed contaner size here. It can depend on contents size.

src/style.css
```diff
@@ -0,0 +1,28 @@
...
+  display: flex;
+  height: 16px;
+  width: 96px;
+}
+
+.stars__star {
+  height: 16px;
+  width: 16px;
+  background-image: url(images/star.svg);
+  background-size: 16px 16px;
```
VinnieJ-2k20, line 15: Looks like your stars size and distance between them are different than than on mockup.
Check exact size if star image, star container dimensions in figma file.
image
image

VinnieJ-2k20 (changes requested):
Need to fix tests and sizes of the elements,
These stars blocks are later used in bigger tasks, so they should look exactly like the design.

src/style.css
```diff
@@ -0,0 +1,25 @@
...
+  background-repeat: no-repeat;
+}
+
+.stars--1 :nth-child(1),
+.stars--2 :nth-child(-n+2),
+.stars--3 :nth-child(-n+3),
+.stars--4 :nth-child(-n+4),
+.stars--5 :nth-child(-n+5) {
+  background: url("images/star-active.svg");
+  background-repeat: no-repeat;
```
VinnieJ-2k20, line 24: This style doesn't change, so no need to keep it here.

VinnieJ-2k20 (approved):

PR #642, opened 2021-07-13
Created a star's rate

StYugurt:
- [DEMO LINK](https://StYugurt.github.io/layout_stars/)
- [TEST REPORT LINK](https://StYugurt.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,24 @@
+body {
+  margin: 0;
+  padding: 0;
+}
+
+.stars {
+  display: flex;
+}
+
+.stars_inline {
+  width: 16px;
+  height: 16px;
+  margin: 2px;
+  background: url(images/star.svg) no-repeat;
+  list-style-type: none;
+}
+
+.stars--5 > .stars_inline,
+.stars--4 > .stars_inline:nth-child(-n+4),
+.stars--3 > .stars_inline:nth-child(-n+3),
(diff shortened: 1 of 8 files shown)


src/style.css
```diff
@@ -0,0 +1,24 @@
...
+  display: flex;
+}
+
+.stars_inline {
+  width: 16px;
+  height: 16px;
+  margin: 2px;
+  background: url(images/star.svg) no-repeat;
+  list-style-type: none;
+}
```
Ihor-Karpyn, line 16: You add margin to all sides, add only for the right side and remove margin in the last star

src/style.css
```diff
@@ -0,0 +1,24 @@
+body {
+  margin: 0;
+  padding: 0;
+}
+
+.stars {
+  display: flex;
+}
```
Ihor-Karpyn, line 8:
```suggestion
.stars {
}
```

redundant, rewrite all stars to span

Ihor-Karpyn (changes requested):
Nice start, let's resolve some problems.

src/style.css
StYugurt: Thanks, resolved

src/style.css
StYugurt: Thanks, resolved

Ihor-Karpyn (approved):

PR #685, opened 2021-08-18
Stars page

Sasha-Krasnoshchokov:
[DEMO LINK](https://Sasha-Krasnoshchokov.github.io/layout_stars/)
[TEST REPORT LINK](https://Sasha-Krasnoshchokov.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -8,6 +8,170 @@
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Stars</h1>
+
+    <div class="stars">
+      <div class="stars--0">
+        <img
+          src="./images/star.svg"
+          alt="Star symbol"
+          class="stars__star"
+        >
+        <img
+          src="./images/star.svg"
+          alt="Star symbol"
+          class="stars__star"
+        >
+        <img
+          src="./images/star.svg"
+          alt="Star symbol"
+          class="stars__star"
+        >
+        <img
+          src="./images/star.svg"
+          alt="Star symbol"
+          class="stars__star"
+        >
+        <img
+          src="./images/star.svg"
+          alt="Star symbol"
+          class="stars__star"
+        >
+      </div>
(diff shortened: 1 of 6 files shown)


src/index.html
```diff
@@ -8,6 +8,170 @@
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Stars</h1>
+
+    <div class="stars">
+      <div class="stars--0">
+        <img
```
Ihor-Karpyn, line 14: it's should be background-img

src/style.css
```diff
@@ -0,0 +1,20 @@
+
+body {
+  margin: 0;
+  /*padding: 22px 0 0 18px;*/
```
Ihor-Karpyn, line 4: remove comment

src/style.css
```diff
@@ -0,0 +1,20 @@
...
+}
+
+.stars > * {
```
Ihor-Karpyn, line 7:
```suggestion
.stars {
```

Ihor-Karpyn (changes requested):
Implement 6 blocks with stars class
The first should have stars--0 class
The next 5 blocks should have an extra class stars--1, stars--2 ... stars--5 (one extra class per block)
Each block should have exactly 5 stars styled as in the card design
The number of active (yellow) stars should be different for all 6 blocks depending on modifier class
The first block (with just stars class) don't have active stars
The other 5 blocks have exactly N fisrt stars active where N is a numbers in an additional css class (modifier) stars--N
Add images in CSS, instead of using the HTML "img" or "svg" tags
Each star in a block should have a class stars__star and no extra classes or inline styles
The star size is fixed (see in the design)
The distance between the stars is also fixed
Use display: flex for the stars block to avoid an issue with extra spaces between individual stars

Ihor-Karpyn (approved):
LGTM

PR #698, opened 2021-08-19
test was successful

BesConstantine:
- [DEMO LINK](https://besconstantine.github.io/layout_stars/)
- [TEST REPORT LINK](https://besconstantine.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,33 @@
+body {
+  margin: 0;
+}
+
+.block {
+  display: flex;
+  flex-direction: row;
+  flex-wrap: wrap;
+  justify-content: space-evenly;
+  align-content: center;
+  width: 100px;
+}
+
+.star {
+  display: flex;
+}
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,33 @@
...
+}
+
+.stars__star {
+  height: 16px;
+  width: 16px;
+  background-image: url("images/star.svg");
+  background-repeat: no-repeat;
+  background-position: center;
+  margin-right: 4px;
+}
```
Ihor-Karpyn, line 25: remove margin-right in last star

Ihor-Karpyn (approved):
Great

src/style.css
BesConstantine: At first, I made without margin-right in the last star but tests weren't successful. When add it in last star tests were correct.

PR #738, opened 2021-09-09
Stars

Igor-MateAcademy:
- [DEMO LINK](https://Igor-MateAcademy.github.io/layout_stars/)
- [TEST REPORT LINK](https://Igor-MateAcademy.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -8,6 +8,84 @@
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Stars</h1>
+    <div class="container">
+      <div class="stars stars--0">
+        <span class="stars__star">
+        </span>
+        <span class="stars__star">
+        </span>
+        <span class="stars__star">
+        </span>
+        <span class="stars__star">
+        </span>
+        <span class="stars__star">
+        </span>
+      </div>
+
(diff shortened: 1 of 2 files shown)


src/index.html
```diff
@@ -8,6 +8,84 @@
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Stars</h1>
+    <div class="container">
+      <div class="stars stars--0">
+        <span class="stars__star">
+        </span>
```
VinnieJ-2k20, line 14: No need to do it, it's short enough. And it's better to make them `div`s, `span` is used mostly for text.
```suggestion
        <span class="stars__star"></span>
```

src/style.css
```diff
@@ -0,0 +1,38 @@
+body {
+  margin: 0;
+}
+
+.container {
+  /* Display */
```
VinnieJ-2k20, line 6: Don't add these comments. In the example it's just a description.

VinnieJ-2k20 (changes requested):
Solution looks good, but let's make it even better.

VinnieJ-2k20 (approved):
Good job 👍

PR #791, opened 2021-10-13
layout stars

i-stanko:


AlinaMatuschak (approved):
Good but fix 1 point on [checklist](https://github.com/mate-academy/layout_stars/blob/master/checklist.md)

PR #792, opened 2021-10-13
stars-first commit

Val-space:
[DEMO LINK](https://Val-space.github.io/layout_stars/)
[TEST REPORT LINK](https://Val-space.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,30 @@
+* {
+  margin: 0;
+  padding: 0;
+}
+
+.stars {
+  display: flex;
+}
+
+.stars__star {
+  width: 16px;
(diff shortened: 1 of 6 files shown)


src/style.css
```diff
@@ -0,0 +1,30 @@
+* {
+  margin: 0;
+  padding: 0;
+}
```
illiarerun, line 4: You don't need to reset default paddings here, also * selector is only suitable for resetting box-sizing, any other properties should be put in explicit tags selector, it your case it's only the body

illiarerun (approved):
LGTM 🔥

PR #807, opened 2021-10-14
added 20211014

VolodyaBalakalchuk:
https://volodyabalakalchuk.github.io/layout_stars/
https://volodyabalakalchuk.github.io/layout_stars/report/html_report/

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,27 @@
+* {
+  margin: 0;
+  padding: 0;
+}
+
+.stars {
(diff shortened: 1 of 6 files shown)


src/style.css
```diff
@@ -0,0 +1,27 @@
+* {
+  margin: 0;
+  padding: 0;
+}
```
illiarerun, line 4: Don't reset your margins in * selector, move it to reset.css

PR #820, opened 2021-10-15
Stars card

ImIkator:
https://ImIkator.github.io/layout_stars/
https://ImIkator.github.io/layout_stars/report/html_report/


AlinaMatuschak (approved):
Awesome 🚀

PR #823, opened 2021-10-15
first attempt

Nadiyahr:
- [DEMO LINK](https://Nadiyahr.github.io/layout_stars/)
- [TEST REPORT LINK](https://Nadiyahr.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,7 +1,9 @@
 * {
   box-sizing: border-box;
+}
+
+body {
   margin: 0;
-  padding: 0;
 }
 
 .stars {


src/style.css
```diff
@@ -0,0 +1,28 @@
+* {
+  box-sizing: border-box;
+  margin: 0;
+  padding: 0;
```
illiarerun, line 4: Don't reset your styles under * selector, it's a bad practice, more so, you only need to reset margin for body in this task

illiarerun (approved):
LGTM 🔥

PR #925, opened 2021-12-21
solution stars

krasnobokydima:
- [DEMO LINK](https://krasnobokydima.github.io/layout_stars/)
- [TEST REPORT LINK](https://krasnobokydima.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,41 @@
+* {
+  margin: 0;
+  padding: 0;
+}
+
+.stars {
+  display: flex;
+  gap: 4px;
+}
+
+.stars__star {
+  background-image: url(./images/star-active.svg);
+  background-size: cover;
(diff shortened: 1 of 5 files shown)


src/style.css
```diff
@@ -0,0 +1,41 @@
...
+  background-image: url(./images/star.svg);
+}
+
+.stars--4 > :nth-child(n + 5) {
+  background-image: url(./images/star.svg);
+}
+
+.stars--5 > :nth-child(n + 6) {
+  background-image: url(./images/star.svg);
+}
```
Anna-Tkachenko, line 41: You can use coma to combine all selectors and don't repeat background-image property for each separate class

Anna-Tkachenko (approved):
Good job! 👍

PR #926, opened 2021-12-21
starsbox was made

mtctxd:
- [DEMO LINK](https://mtctxd.github.io/layout_stars/)
- [TEST REPORT LINK](https://mtctxd.github.io/layout_stars/report/html_report/)


Anna-Tkachenko (changes requested):
You should add star images as a background for blocks using css styles (don't use tag img)

Anna-Tkachenko (approved):
Just try to pass tests

PR #939, opened 2021-12-22
Develop

vsuslov29:
https://vsuslov29.github.io/layout_stars/
https://vsuslov29.github.io/layout_stars/report/html_report/


Anna-Tkachenko (approved):
Go to Q&A session to fix tests

PR #942, opened 2021-12-22
save

HeorhiiPluhovyi:
- [DEMO LINK](https://HeorhiiPluhovyi.github.io/layout_stars/)
- [TEST REPORT LINK](https://HeorhiiPluhovyi.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,40 @@
+body {
+  margin: 0;
+}
+
+.stars {
+  display: flex;
+  gap: 4px;
+}
+
+.stars__star {
+  margin: 2px 6px 2px 2px;
+  width: 12px;
+  height: 12px;
+  background-image: url(.\images\star-active.svg);
+  background-size: cover;
+}
+
+.stars--0 > :nth-child(n+1) {
+  background-image: url(.\images\star.svg);
+}
+
+.stars--1 > :nth-child(n+2) {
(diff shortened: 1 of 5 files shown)


src/style.css
```diff
@@ -0,0 +1,40 @@
...
+.stars--3 > :nth-child(n+4) {
+  background-image: url(.\images\star.svg);
+}
+
+.stars--4 > :nth-child(n+5) {
+  background-image: url(.\images\star.svg);
+}
+
+.stars--5 > :nth-child(n+6) {
+  background-image: url(.\images\star.svg);
```
Anna-Tkachenko, line 39: Check point 2 in [checklist](https://github.com/mate-academy/layout_stars/blob/master/checklist.md)

Anna-Tkachenko (changes requested):

src/style.css
```diff
@@ -0,0 +1,25 @@
...
+
+.stars {
+  display: flex;
+  gap: 4px;
+}
+
+.stars__star {
+  margin: 2px 6px 2px 2px;
+  width: 12px;
+  height: 12px;
```
Anna-Tkachenko, line 13: You shouldn't use vertical margins, only width, height and margin-right

Anna-Tkachenko (changes requested):

PR #952, opened 2021-12-23
Develop

YanaLuzanovska:
- [DEMO LINK](https://YanaLuzanovska.github.io/layout_stars/)
- [TEST REPORT LINK](https://Yanaluzanovska.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,26 @@
+body {
+  margin: 0;
+  padding: 0;
+}
+
+.stars {
+  display: flex;
+  gap: 4px;
+}
+
+.stars__star {
+  background-image: url("./images/star-active.svg");
+  background-size: cover;
+  width: 12px;
+  height: 12px;
+  margin: 2px 6px 2px 2px;
+}
+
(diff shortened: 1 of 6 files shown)


src/style.css
```diff
@@ -0,0 +1,26 @@
+body {
+  margin: 0;
+  padding: 0;
+}
+
+.stars {
+  display: flex;
+  gap: 4px;
```
Anna-Tkachenko, line 8: don't use gap

src/style.css
```diff
@@ -0,0 +1,26 @@
...
+}
+
+.stars__star {
+  background-image: url("./images/star-active.svg");
+  background-size: cover;
+  width: 12px;
+  height: 12px;
+  margin: 2px 6px 2px 2px;
```
Anna-Tkachenko, line 16: don't use vertical margins, just margin-right
center background image

Anna-Tkachenko (changes requested):

OleksiiBlezniuk (approved):
Good job! 🚀

PR #955, opened 2021-12-23
first-ver_stars

yahyafullstack01:
- [DEMO LINK](https://yahyafullstack01.github.io/layout_stars/)
- [TEST REPORT LINK](https://yahyafullstack01.github.io/layout_stars/report/html_report/)


src/index.html
```diff
@@ -8,6 +8,97 @@
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Stars</h1>
+    <div class="container">
+      <div class="star starfirst">
+        <img class="image" src="./images/star.svg">
```
vkryvytskyy, line 13: add these images as background to div elements

src/index.html
```diff
@@ -8,6 +8,97 @@
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Stars</h1>
+    <div class="container">
+      <div class="star starfirst">
```
vkryvytskyy, line 12:
```suggestion
      <div class="star">
```
These star elements should have only one class

vkryvytskyy (changes requested):
use css to set star images and nth-child selectors to decide which one should be active

src/index.html
```diff
@@ -8,6 +8,187 @@
...
   </head>
   <body>
-    <h1>Stars</h1>
+    <div class="container">
+      <div class="star">
+        <img class="image"
+          src="./images/star.svg"
+          alt="star1"
+        >
+      </div>
```
aliceblunt8, line 17: all stars should be just divs, not images, add image with css

```suggestion
      <div class="star"></div>
```

aliceblunt8: also add corresponding classes to these elements (read task carefully)

```
Implement 6 blocks with stars class
The first should have stars--0 class
The next 5 blocks should have an extra class stars--1, stars--2 ... stars--5 (one extra class per block)
```

aliceblunt8 (changes requested):

aliceblunt8 (approved):

PR #991, opened 2022-01-18
stars added

barannikovArtem:
- [DEMO LINK](https://barannikovArtem.github.io/layout_stars/)
- [TEST REPORT LINK](https://barannikovArtem.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,24 @@
+* {
+  margin: 0;
+}
+
+.stars {
+  display: flex;
+}
+
+.stars__star {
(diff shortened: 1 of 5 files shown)


src/style.css
```diff
@@ -0,0 +1,24 @@
+* {
+  margin: 0;
+}
```
solaryasha, line 3: Avoid using `*` selector with margins

solaryasha (approved):
Good job !

PR #999, opened 2022-01-19
fly to the STARS:)

max-shlikhta:
- [DEMO](https://max-shlikhta.github.io/layout_stars/)
- [TEST](https://max-shlikhta.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,24 @@
+* {
+  margin: 0;
+}
+
+.stars {
+  display: flex;
+}
+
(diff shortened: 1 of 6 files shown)


src/style.css
```diff
@@ -0,0 +1,24 @@
+* {
+  margin: 0;
+}
```
solaryasha, line 3: Avoid using `*` selector with margins

solaryasha (approved):

PR #1012, opened 2022-01-20
complete task

irreligious86:
- [DEMO LINK](https://irreligious86.github.io/layout_stars/)
- [TEST REPORT LINK](https://irreligious86.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,37 @@
+* {
+  margin: 0;
+  padding: 0;
+  box-sizing: border-box;
+}
+
+.wrapper {
+  margin: 0 auto;
+  display: flex;
+  flex-direction: column;
+  width: 98px;
+  height: 98px;
+}
+
+.stars {
+  display: flex;
+  width: 96px;
+  justify-content: space-between;
+  gap: 2px;
+}
+
+.stars__star {
+  width: 12px;
+  height: 14px;
+  margin: 1px 2px;
+  background-image: url("./images/star.svg");
+  background-position: center;
+  background-repeat: no-repeat;
+}
+
+.stars--1 :nth-child(-n+1),
+.stars--2 :nth-child(-n+2),
+.stars--3 :nth-child(-n+3),
+.stars--4 :nth-child(-n+4),
+.stars--5 :nth-child(-n+5) {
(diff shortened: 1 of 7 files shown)


src/style.css
```diff
@@ -0,0 +1,37 @@
...
+  box-sizing: border-box;
+}
+
+.wrapper {
+  margin: 0 auto;
+  display: flex;
+  flex-direction: column;
+  width: 98px;
+  height: 98px;
+}
```
solaryasha, line 13: You don't need this wrapper

src/style.css
```diff
@@ -0,0 +1,37 @@
...
+
+.stars {
+  display: flex;
+  width: 96px;
```
solaryasha, line 17: You don't need to hardcode width size

src/style.css
```diff
@@ -0,0 +1,37 @@
...
+  justify-content: space-between;
```
solaryasha, line 18: We don't need this rule

src/style.css
```diff
@@ -0,0 +1,37 @@
...
+  gap: 2px;
```
solaryasha, line 19: Test program don't support gap.But, it's cool that you using it

src/style.css
```diff
@@ -0,0 +1,37 @@
...
+}
+
+.stars__star {
+  width: 12px;
+  height: 14px;
+  margin: 1px 2px;
```
solaryasha, line 25: We don't need this margins

src/style.css
```diff
@@ -0,0 +1,37 @@
...
+.stars {
+  display: flex;
+  width: 96px;
+  justify-content: space-between;
+  gap: 2px;
+}
+
+.stars__star {
+  width: 12px;
+  height: 14px;
```
solaryasha, line 24: It should be 16px

solaryasha (changes requested):

src/style.css
```diff
@@ -0,0 +1,26 @@
+* {
+  margin: 0;
+  padding: 0;
```
solaryasha, line 3: Avoid zeroing margins using `*` selector

solaryasha (approved):

PR #1018, opened 2022-01-21
stars are too thin:(

Marta1993-droid:
- [DEMO LINK](https://Marta1993-droid.github.io/layout_search-bar-airbnb/)
- [TEST REPORT LINK](https://Marta1993-droid.github.io/layout_search-bar-airbnb/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,32 @@
+* {
+  margin-top: 0;
+}
+
+.stars {
+  display: flex;
+}
+
+.stars--star {
+  height: 16px;
+  width: 16px;
+  background-image: url(./images/star.svg);
+  background-repeat: no-repeat;
+  background-position: center;
+  padding-right: 4px;
+  margin-right: 4px;
+}
+
+/* .stars--star:last-child {
+  margin-right: 0;
+} */
+ 
+.stars--1 .stars--star:first-child,
+.stars--2 .stars--star:nth-child(-n+2),
+.stars--3 .stars--star:nth-child(-n+3),
+.stars--4 .stars--star:nth-child(-n+4),
+.stars--5 .stars--star:nth-child(-n+5) {
+  background-image: url(./images/star-active.svg);
+  background-repeat: no-repeat;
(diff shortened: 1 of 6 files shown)


src/style.css
```diff
@@ -0,0 +1,32 @@
+* {
+  margin-top: 0;
+}
```
solaryasha, line 3: You should remove all margins in `body` tag

src/style.css
```diff
@@ -0,0 +1,32 @@
...
+  background-image: url(./images/star.svg);
+  background-repeat: no-repeat;
+  background-position: center;
+  padding-right: 4px;
+  margin-right: 4px;
+}
+
+/* .stars--star:last-child {
+  margin-right: 0;
+} */
```
solaryasha, line 21: Don't leave comments

src/style.css
```diff
@@ -0,0 +1,32 @@
...
+ 
+.stars--1 .stars--star:first-child,
+.stars--2 .stars--star:nth-child(-n+2),
+.stars--3 .stars--star:nth-child(-n+3),
+.stars--4 .stars--star:nth-child(-n+4),
+.stars--5 .stars--star:nth-child(-n+5) {
+  background-image: url(./images/star-active.svg);
+  background-repeat: no-repeat;
+  height: 16px;
+  width: 16px;
```
solaryasha, line 31: You don't need overwrite these rules

solaryasha (changes requested):

src/style.css
Marta1993-droid: got it

src/style.css
Marta1993-droid: got it

src/style.css
Marta1993-droid: got it

src/style.css
Marta1993-droid: that was the point! thank you!

Ihor-Karpyn (approved):

PR #1035, opened 2022-02-05
solution

anna-tovkailo:


src/index.html
```diff
@@ -7,7 +7,72 @@
...
+      <div class="stars--star"></div>
+
+      <div class="stars--star"></div>
+      
+      <div class="stars--star"></div>
+      
+      <div class="stars--star"></div>
+      
+      <div class="stars--star"></div>
+    </div>
```
aliceblunt8, line 21: stars is block, star is element, stars--0 is modifier (BEM)

```suggestion
    <div class="stars stars--0">
      <div class="stars__star"></div>
      <div class="stars__star"></div>
      <div class="stars__star"></div>
      <div class="stars__star"></div>
      <div class="stars__star"></div>
    </div>
```

aliceblunt8: also remove empty lines between empty divs

src/style.css
```diff
@@ -0,0 +1,37 @@
...
+.stars--3 .stars--star:nth-child(3),
+.stars--4 .stars--star:nth-child(1),
+.stars--4 .stars--star:nth-child(2),
+.stars--4 .stars--star:nth-child(3),
+.stars--4 .stars--star:nth-child(4),
+.stars--5 .stars--star:nth-child(1),
+.stars--5 .stars--star:nth-child(2),
+.stars--5 .stars--star:nth-child(3),
+.stars--5 .stars--star:nth-child(4),
+.stars--5 .stars--star:nth-child(5) {
```
aliceblunt8, line 35: there should be one selector for each modifier
for example all stars inside stars--5 should be yellow, we can achieve that with this selector `.stars--5 .stars__stars` (= selecting all stars__stars divs inside stars--5)
and for stars--3 only first 3 stars should be yellow, it's `.stars--3 .stars__item:nth-child(-n+3)` (= selecting first 3 stars__stars divs inside stars--3)

I suggest you read more about nth-child pseudoclass to understand the meaning of n argument inside it 🙂 
https://css-tricks.com/almanac/selectors/n/nth-child/
http://nthmaster.com/

aliceblunt8 (changes requested):

danylo-onishchuk (approved):

PR #1042, opened 2022-02-11
styles have been removed

divineretribution1:
- [DEMO LINK](https://divineretribution1.github.io/layout_stars/)
- [TEST REPORT LINK](https://divineretribution1.github.io/layout_stars/report/html_report/)


aliceblunt8 (changes requested):
remove star-is-active class, use nth-child to select needed amount of stars

PR #1047, opened 2022-02-15
Stars added

VETALEK:
https://vetalek.github.io/layout_stars/
https://vetalek.github.io/layout_stars/report/html_report/

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -8,6 +8,52 @@
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Stars</h1>
+    <div class="stars stars--0">
+      <div class="empty"></div>
+      <div class="empty"></div>
+      <div class="empty"></div>
+      <div class="empty"></div>
+      <div class="empty"></div>
+    </div>
+
+    <div class="stars stars--1">
+      <div class="empty"></div>
+      <div class="empty"></div>
+      <div class="empty"></div>
+      <div class="empty"></div>
+      <div class="empty"></div>
+    </div>
+
+    <div class="stars stars--2">
+      <div class="empty"></div>
+      <div class="empty"></div>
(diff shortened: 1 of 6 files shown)


src/index.html
```diff
@@ -8,6 +8,52 @@
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Stars</h1>
+    <div class="stars stars--0">
+      <div class="empty"></div>
```
illiarerun, line 12: Each star in a block should have a class stars__star and no extra classes or inline styles

src/style.css
```diff
@@ -0,0 +1,28 @@
+body {
+  margin: 0;
+}
+
+.stars {
+  display: flex;
+  flex-direction: row-reverse;
+  justify-content: space-around;
```
illiarerun, line 8: you don't really need it, in fact it can break the tests
```suggestion
```

src/style.css
```diff
@@ -0,0 +1,28 @@
...
+  align-items: center;
+  width: 100px;
```
illiarerun, line 10: don't set width for container explicitly

src/style.css
```diff
@@ -0,0 +1,28 @@
...
+  height: 16px;
```
illiarerun, line 11: same with height, let the content inside dictate it

src/style.css
```diff
@@ -0,0 +1,28 @@
...
+}
+
+.empty {
+  width: 13px;
+  height: 12px;
```
illiarerun, line 16: both should be 16

illiarerun (changes requested):
Nice start, but your test link doesn't work also you need to apply some fixes:

illiarerun (approved):

PR #1050, opened 2022-02-15
Create stars

ivan-kadykalo:
- [DEMO LINK](https://ivan-kadykalo.github.io/layout_stars/)
- [TEST REPORT LINK](https://ivan-kadykalo.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -7,50 +7,48 @@
     <title>Stars</title>
     <link rel="stylesheet" href="./style.css">
   </head>
-  <body>
-    <main class="container">
-      <div class="stars stars--0">
-        <div class="stars__star"></div>
-        <div class="stars__star"></div>
-        <div class="stars__star"></div>
-        <div class="stars__star"></div>
-        <div class="stars__star"></div>
-      </div>
-      <div class="stars stars--1">
-        <div class="stars__star"></div>
-        <div class="stars__star"></div>
-        <div class="stars__star"></div>
-        <div class="stars__star"></div>
-        <div class="stars__star"></div>
-      </div>
-      <div class="stars stars--2">
-        <div class="stars__star"></div>
-        <div class="stars__star"></div>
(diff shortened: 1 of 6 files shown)


ivan-kadykalo:
I have tried to use gap: 4px in flex but it doesn't past a tests, and it's looks strange because in a viewport it looks the same as with margin-right. 
I'm doing something wrong or it is problem with a tests? 
The part of code with gap i left in the code, look on it please.

src/index.html
```diff
@@ -8,6 +8,49 @@
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Stars</h1>
+    <main class="container">
```
illiarerun, line 11: you don't really need this container, remove it

src/style.css
```diff
@@ -0,0 +1,30 @@
+* {
+  margin: 0;
+}
```
illiarerun, line 3: don't reset your margins with * selector, it's bad practice, you only need to reset margin for body in this task

src/style.css
```diff
@@ -0,0 +1,30 @@
...
+
+.stars__star {
+  height: 16px;
+  width: 16px;
+  box-sizing: border-box;
```
illiarerun, line 8: box-sizing should be the same for all elements, that's the property you can change with * selector

src/style.css
```diff
@@ -0,0 +1,30 @@
...
+  background-image: url(images/star.svg);
+  background-repeat: no-repeat;
+  background-position: center center;
+}
+
+.stars {
+  display: flex;
+  /* Don`t past a tests but looks correct */
+  /* gap: 4px; */
```
illiarerun, line 17: The thing with flex-gap is that not all of the browsers support it yet, that's why tests doesn't take it to account

illiarerun (approved):

PR #1052, opened 2022-02-15
some message

bohdan-daniliuk:


src/style.css
```diff
@@ -0,0 +1,30 @@
+
+* {
+  margin: 0;
+}
```
illiarerun, line 4: don't reset your margins with * selector, it's a bad practice, you only need to reset body margin in this task

src/style.css
```diff
@@ -0,0 +1,30 @@
...
+
+.stars {
+  display: flex;
+
+}
```
illiarerun, line 9:
```suggestion
.stars {
  display: flex;
}
```

src/style.css
```diff
@@ -0,0 +1,30 @@
...
+
+.stars__star {
+  height: 16px;
+  width: 16px;
+  background-image: url(images/star.svg);
+  background-repeat: no-repeat;
+  background-position: center center;
+
+}
```
illiarerun, line 22:
```suggestion
  background-position: center center;
}
```

illiarerun (changes requested):
Good job 🔥  Little fixes left to do:

src/style.css
```diff
@@ -0,0 +1,28 @@
+
+body {
+  margin: 0;
+}
+
+.stars {
+  display: flex;
+}
+
+.stars > :not(:last-child) {
```
illiarerun, line 10: don't increase specificity if it's unnecessary
```suggestion
stars__star:not(:last-child) {
```

illiarerun (approved):
Good job 🔥  One comment left:

PR #1056, opened 2022-02-16
implemented stars

ElizaChernysh:
https://elizachernysh.github.io/layout_stars/
https://elizachernysh.github.io/layout_stars/report/html_report/

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,29 @@
+* {
+  margin: 0;
+  padding: 0;
+}
+
+.stars__star {
+  height: 16px;
+  width: 16px;
+ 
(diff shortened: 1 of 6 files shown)


src/style.css
```diff
@@ -0,0 +1,29 @@
+* {
+  margin: 0;
+  padding: 0;
+}
```
illiarerun, line 4: don't reset your margins and paddings with * selector, it's a bad practice, you should only reset margin for body in this task

illiarerun (approved):
Good job 🔥  Left one comment, make sure to fix it:

PR #1095, opened 2022-02-18
active stars

MalukhaD:
https://MalukhaD.github.io/layout_stars/
https://MalukhaD.github.io/layout_stars/report/html_report/

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,24 @@
+html {
+  margin: 0;
+}
+
+.stars .star {
+  margin-right: 4px;
+  height: 16px;
+  width: 16px;
+  background-image: url(images/star.svg);
+  background-position: center;
+  background-repeat: no-repeat;
+}
+
+.stars {
+  display: flex;
+}
+
+.stars--1 :nth-child(1),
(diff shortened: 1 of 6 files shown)


src/style.css
```diff
@@ -0,0 +1,24 @@
+html {
+  margin: 0;
+}
```
illiarerun, line 3: html doesn't have margins, you may want to reset margins for body

src/style.css
```diff
@@ -0,0 +1,24 @@
...
+
+.stars .star {
```
illiarerun, line 5: don't increase specificity
```suggestion
.star {
```

src/style.css
```diff
@@ -0,0 +1,24 @@
...
+  margin-right: 4px;
```
illiarerun, line 6: last one shouldn't have this margin

illiarerun (changes requested):
Good job, few fixes left:

MaksimusCRS (changes requested):
Ссылки на демо и тесты не работают

MaksimusCRS (approved):

PR #1096, opened 2022-02-18
looks ok)

yanaost:
https://yanaost.github.io/layout_stars/
https://yanaost.github.io/layout_stars/report/html_report/


danylo-onishchuk (approved):
good job!

PR #1099, opened 2022-02-18
Develop

Yurii837:


src/style.css
```diff
@@ -0,0 +1,41 @@
+body {
+  margin: 0;
+}
+
+.starCard {
+  display: flex;
+  flex-direction: column;
+}
```
illiarerun, line 8: that's redundant styles, div are positioned in column by default
```suggestion
```

src/style.css
```diff
@@ -0,0 +1,41 @@
...
+
+.rows {
+  display: flex;
+}
+
+.stars_star {
+  width: 16px;
+  height: 16px;
+  margin-right: 4px;
```
illiarerun, line 17: last one shouldn't have margin

src/style.css
```diff
@@ -0,0 +1,41 @@
...
+  background-image: url(./images/star-active.svg);
+}
+
+.stars--4 > .stars_star:nth-child(-n+4) {
+  background-image: url(./images/star-active.svg);
+}
+
+.stars--5 > .stars_star:nth-child(-n+5) {
+  background-image: url(./images/star-active.svg);
+}
```
illiarerun, line 41: you can combine these selectors

illiarerun (changes requested):
Good job, but let's work on fixes now:

src/style.css
Yurii837: Thanks

Yurii837: It seems you are wrong, try it

src/style.css
```diff
@@ -0,0 +1,45 @@
...
```
MaksimusCRS, line 45: Найди способ записать это сокращенно, что бы не повторять одно и тоже много раз

MaksimusCRS (changes requested):

MaksimusCRS (approved):

PR #1123, opened 2022-03-16
index html created, styles added to classes. 

efdalyalcin:
The test works except there is no references for 3 tests

- [DEMO LINK](https://efdalyalcin.github.io/layout_stars/)
- [TEST REPORT LINK](https://efdalyalcin.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,45 @@
+body {
+  margin: 0;
+  box-sizing: border-box;
+  font-size: 0;
+}
+
+div {
+  margin: 0;
+  padding: 0;
+}
+
+.container {
+  display: flex;
+  flex-flow: column wrap;
+  width: 100px;
+}
+
+.stars__star {
+  display: inline-block;
+  background-image: url(images/star.svg);
+  background-size: cover;
+  height: 12px;
+  width: 12px;
+  margin: 2px 6px 2px 2px;
+}
+
+.stars--1 :nth-child(1) {
+  background-image: url(images/star-active.svg);
+}
+
+.stars--2 :nth-child(-n+2) {
+  background-image: url(images/star-active.svg);
+}
+
+.stars--3 :nth-child(-n+3) {
+  background-image: url(images/star-active.svg);
+}
+
+.stars--4 :nth-child(-n+4) {
+  background-image: url(images/star-active.svg);
+}
+
+.stars--5 :nth-child(n) {
+  background-image: url(images/star-active.svg);
+}
--- a/readme.md
+++ b/readme.md
@@ -1,7 +1,7 @@
 # Stars block
 Replace `<your_account>` with your Github username and copy the links to Pull Request description:
-- [DEMO LINK](https://<your_account>.github.io/layout_stars/)
-- [TEST REPORT LINK](https://<your_account>.github.io/layout_stars/report/html_report/)
+- [DEMO LINK](https://efdalyalcin.github.io/layout_stars/)
+- [TEST REPORT LINK](https://efdalyalcin.github.io/layout_stars/report/html_report/)
 
 > Follow [this instructions](https://mate-academy.github.io/layout_task-guideline)
 ___
--- a/src/index.html
+++ b/src/index.html
@@ -8,6 +8,49 @@
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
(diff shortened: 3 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,45 @@
...
+  background-image: url(images/star-active.svg);
+}
+
+.stars--4 :nth-child(-n+4) {
+  background-image: url(images/star-active.svg);
+}
+
+.stars--5 :nth-child(n) {
+  background-image: url(images/star-active.svg);
+}
```
aliceblunt8, line 45: combine these selectors
like these

```
.stars--1 :nth-child(1),
.stars--2 :nth-child(-n+2) {
}
```

aliceblunt8 (changes requested):

src/style.css
```diff
@@ -0,0 +1,33 @@
+body {
+  margin: 0;
+  box-sizing: border-box;
```
ihor-jpeg, line 3: this will be only applied to the body. i'm not sure why you'd need that

src/style.css
```diff
@@ -0,0 +1,33 @@
...
+  font-size: 0;
```
ihor-jpeg, line 4: but why?

src/style.css
```diff
@@ -0,0 +1,33 @@
...
+}
+
+div {
+  margin: 0;
+  padding: 0;
+}
```
ihor-jpeg, line 10: let's not style tags. also divs do not have any default margins or paddings

src/style.css
```diff
@@ -0,0 +1,33 @@
...
+
+.container {
+  display: flex;
+  flex-flow: column wrap;
+  width: 100px;
+}
```
ihor-jpeg, line 16: if you remove these styles, everything will still work fine

the idea here was to not only set the modifier, like `stars--3`, but also give that div the `stars` class. and set `display: flex` there. but you solved this problem using `display: inline-block;` for `stars__star`

ihor-jpeg (changes requested):
Well done! Let's clean up the code!

src/style.css
```diff
@@ -0,0 +1,28 @@
...
+
+/* comment .stars and uncomment the others, font-size metters there which i am curios about, why?*/
+.stars {
+  display: flex;
+  flex-flow: row wrap;
+  width: 100px;
+}
+
+.stars__star {
+  /* display: inline-block; */
```
ihor-jpeg, line 14: also, if you set `display: flex` for the parent, the `display` property for direct children is ignored. that's why you do not need it here, since `.stars__star` are direct children of `.stars`

src/style.css
```diff
@@ -0,0 +1,28 @@
+body {
+  margin: 0;
+  /* font-size: 0; */
+}
+
+/* comment .stars and uncomment the others, font-size metters there which i am curios about, why?*/
+.stars {
+  display: flex;
+  flex-flow: row wrap;
+  width: 100px;
```
ihor-jpeg, line 10: do not set the width explicitly, it is not needed here.

also, `flex-flow` is a shorthand property for
- flex-direction
- flex-wrap

`flex-direction` has the default value of `row` and `flex-wrap` is not needed here. so you can just remove it

src/style.css
```diff
@@ -0,0 +1,28 @@
...
+  margin: 2px 6px 2px 2px;
+}
+
+.stars--1 :nth-child(1),
+.stars--2 :nth-child(-n+2),
+.stars--3 :nth-child(-n+3),
+.stars--4 :nth-child(-n+4),
+.stars--5 :nth-child(n) {
+  background-image: url(images/star-active.svg);
+}
```
ihor-jpeg, line 28: what will happen if I decide to change the number of stars in one row to, let's say, 7?

ihor-jpeg (changes requested):
I have left a few comments. Hope it makes things clearer

src/style.css
```diff
@@ -0,0 +1,28 @@
+body {
+  margin: 0;
+  /* font-size: 0; */
+}
+
+/* comment .stars and uncomment the others, font-size metters there which i am curios about, why?*/
```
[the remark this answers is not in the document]
efdalyalcin: Thanks, it is much clear now but i will study the theory of flexbox in the weekend to make sure...

PR #1124, opened 2022-03-16
completed layout_stars task

59MrRobot:
- [DEMO LINK](https://59MrRobot.github.io/layout_stars/)
- [TEST REPORT LINK](https://59MrRobot.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -7,7 +7,55 @@
     <title>Stars</title>
     <link rel="stylesheet" href="./style.css">
   </head>
-  <body>
-    <h1>Stars</h1>
+  <body class="body">
+    <div class="container">
+      <div class="stars--0">
+        <div class="star__star"></div>
+        <div class="star__star"></div>
+        <div class="star__star"></div>
+        <div class="star__star"></div>
+        <div class="star__star"></div>
+      </div>
+
+      <div class="stars--1">
+        <div class="star__star"></div>
+        <div class="star__star"></div>
+        <div class="star__star"></div>
+        <div class="star__star"></div>
+        <div class="star__star"></div>
+      </div>
+
+      <div class="stars--2">
+        <div class="star__star"></div>
+        <div class="star__star"></div>
+        <div class="star__star"></div>
+        <div class="star__star"></div>
+        <div class="star__star"></div>
+      </div>
+
+      <div class="stars--3">
+        <div class="star__star"></div>
+        <div class="star__star"></div>
+        <div class="star__star"></div>
+        <div class="star__star"></div>
+        <div class="star__star"></div>
+      </div>
+
+      <div class="stars--4">
+        <div class="star__star"></div>
+        <div class="star__star"></div>
+        <div class="star__star"></div>
+        <div class="star__star"></div>
+        <div class="star__star"></div>
+      </div>
+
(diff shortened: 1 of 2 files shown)


src/style.css
```diff
@@ -0,0 +1,40 @@
...
+
+.stars--3 :nth-child(-n+3) {
+  background-image: url(./images/star-active.svg);
+}
+
+.stars--4 :nth-child(-n+4) {
+  background-image: url(./images/star-active.svg);
+}
+
+.stars--5 :nth-child(n) {
```
ihor-jpeg, line 38: if i add more stars to each row, let's say 7, this modifier will not make 5 stars colored yellow, it will make all of them yellow, which might be confusing

src/style.css
```diff
@@ -0,0 +1,40 @@
...
+  background-image: url(./images/star-active.svg);
+}
+
+.stars--3 :nth-child(-n+3) {
+  background-image: url(./images/star-active.svg);
+}
+
+.stars--4 :nth-child(-n+4) {
+  background-image: url(./images/star-active.svg);
+}
```
ihor-jpeg, line 36: you are using the same styles for these classes, better group them and make the code cleaner

```suggestion
.stars--1 :nth-child(1),
.stars--2 :nth-child(-n+2),
.stars--2 :nth-child(-n+3),
.stars--2 :nth-child(-n+4),
.stars--2 :nth-child(-n+5) {
  background-image: url(./images/star-active.svg);
}
```

src/index.html
```diff
@@ -7,7 +7,55 @@
     <title>Stars</title>
     <link rel="stylesheet" href="./style.css">
   </head>
-  <body>
-    <h1>Stars</h1>
+  <body class="body">
+    <div class="container">
+      <div class="stars--0">
```
ihor-jpeg, line 12: i don't think you are using the `stars--0` modifier

ihor-jpeg (changes requested):
Well done! Let's fix a few things. I have left a few comments below

src/index.html
59MrRobot: I've removed it

src/style.css
59MrRobot: I've grouped them together

src/style.css
59MrRobot: Noted, I have changed my logic.

src/style.css
```diff
@@ -0,0 +1,28 @@
+.body {
+  margin: 0;
+  box-sizing: border-box;
+  font-size: 0;
+}
+
+.container {
+  display: flex;
+  flex-flow: column wrap;
```
ihor-jpeg, line 9: sorry, missed this one.
there is no property `column wrap` for flex-flow. the reason it works is that it ignores the `wrap` part 😆 
so `flex-flow: column` is the way to go.

another way of putting blocks with stars in rows would be setting `flex-direction: column`
JFYI

ihor-jpeg (approved):
Well done. 🚀  Have a look at my comment, please.

src/style.css
59MrRobot: Alright thank you, I will remember this for the future.

PR #1125, opened 2022-03-16
created html layout for stars, added css and deployed to gh-pages

LukaszGryziewicz:
- [DEMO LINK](https://LukaszGryziewicz.github.io/layout_stars/)
- [TEST REPORT LINK](https://LukaszGryziewicz.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -8,6 +8,52 @@
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Stars</h1>
+    <div class="stars stars--0">
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+    </div>
+
+    <div class="stars stars--1">
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+    </div>
+
+    <div class="stars stars--2">
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+    </div>
+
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,30 @@
...
+.stars--4 .stars__star:nth-child(1),
+.stars--4 .stars__star:nth-child(2),
+.stars--4 .stars__star:nth-child(3),
+.stars--4 .stars__star:nth-child(4),
+.stars--3 .stars__star:nth-child(1),
+.stars--3 .stars__star:nth-child(2),
+.stars--3 .stars__star:nth-child(3),
+.stars--2 .stars__star:nth-child(1),
+.stars--2 .stars__star:nth-child(2),
+.stars--1 .stars__star:nth-child(1) {
```
ihor-jpeg, line 28: you can use a formula like :nth-child(-n + #stars-number) to acheive the same in a cleaner way

src/index.html
```diff
@@ -8,6 +8,52 @@
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Stars</h1>
+    <div class="stars stars--0">
```
ihor-jpeg, line 11: i don't think `stars--0` is necessary

src/style.css
```diff
@@ -0,0 +1,30 @@
+body {
+  margin: 0;
+}
+
+.stars {
+  display: flex;
+  width: 100%;
```
ihor-jpeg, line 7: your stars container is a div, is display is block, so the width is 100% by default

ihor-jpeg (changes requested):
Well done! I have left a few comments below. Let's fix them!

src/index.html
LukaszGryziewicz: Tests fail without this class :/

src/style.css
ihor-jpeg: yikes tests 🤢

ihor-jpeg (approved):
Well done! 🚀 
Very clean CSS

PR #1126, opened 2022-03-16
layout stars task done

amitpatiljc:
- [DEMO LINK](https://amitpatiljc.github.io/layout_stars/)
- [TEST REPORT LINK](https://amitpatiljc.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,39 @@
+.body {
+  margin: 0;
+  box-sizing: border-box;
+}
+
+.stars {
+  display: flex;
+  flex-direction: row;
+  flex-wrap: wrap;
+}
+
+.stars__star {
+  display: inline-block;
+  margin: 2px 6px 2px 2px;
+  height: 12px;
+  width: 12px;
+  background-image: url(./images/star.svg);
+  background-size: cover;
+}
+
+.stars--1 :nth-child(1) {
+  background-image: url(./images/star-active.svg);
+}
+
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,39 @@
+.body {
+  margin: 0;
+  box-sizing: border-box;
+}
+
+.stars {
+  display: flex;
+  flex-direction: row;
+  flex-wrap: wrap;
```
ihor-jpeg, line 9: `flex-direction: row` - this is the default. no need to set it explicitly
`flex-wrap` is not needed here

ihor-jpeg (approved):
Well done!

src/style.css
```diff
@@ -0,0 +1,39 @@
...
+  background-image: url(./images/star-active.svg);
+}
+
+.stars--4 :nth-child(-n+4) {
+  background-image: url(./images/star-active.svg);
+}
+
+.stars--5 :nth-child(-n+5) {
+  background-image: url(./images/star-active.svg);
+}
```
ihor-jpeg, line 39: You can group these.

```suggestion
.stars--1 :nth-child(1),
.stars--2 :nth-child(-n+2),
.stars--2 :nth-child(-n+3),
.stars--2 :nth-child(-n+4),
.stars--2 :nth-child(-n+5), {
  background-image: url(./images/star-active.svg);
}
```

PR #1127, opened 2022-03-16
Stars

huberciak:


src/index.html
```diff
@@ -8,6 +8,48 @@
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Stars</h1>
+    <div class="star-container stars--0">
```
ihor-jpeg, line 11: you are not using `stars--0`

src/style.css
```diff
@@ -0,0 +1,40 @@
+body {
+  margin: 0;
+}
+
+.star-container {
+  display: flex;
+  width: 100%;
```
ihor-jpeg, line 7: your container is a `div`, its `display` prop is block, so the width is 100% by default

ihor-jpeg (changes requested):
Please add the DEMO link and the TEST REPORT LINK to the PR description

src/style.css
```diff
@@ -0,0 +1,39 @@
...
+  background-image: url(images/star-active.svg);
+}
+
+.stars--3 :nth-child(-n+3) {
+  background-image: url(images/star-active.svg);
+}
+
+.stars--4 :nth-child(-n+4) {
+  background-image: url(images/star-active.svg);
+}
```
ihor-jpeg, line 35: Let's group these bad boys. The styles applied are the same.

```suggestion
.stars--1 :nth-child(1),
.stars--2 :nth-child(-n+2),
.stars--2 :nth-child(-n+3),
.stars--2 :nth-child(-n+4),
.stars--2 :nth-child(-n+5) {
  background-image: url(images/star-active.svg);
}
```

src/style.css
```diff
@@ -0,0 +1,39 @@
...
+
+.stars--5 :nth-child(n) {
```
ihor-jpeg, line 37: the modifier says there should be 5 active stars. now imagine I change my design and there are 7 stars in each row. the design will be broken

ihor-jpeg (changes requested):
Well done! There a few small bits to fix.

src/style.css
```diff
@@ -0,0 +1,27 @@
...
+
+.stars__star:first-child {
+  margin-left: 2px;
+}
+
+.stars--1 :nth-child(1),
+.stars--2 :nth-child(-n+2),
+.stars--3 :nth-child(-n+3),
+.stars--4 :nth-child(-n+4),
+.stars--5 :nth-child(n) {
```
ihor-jpeg, line 25: again, if I have 7 stars per row, I use this modifier (stars--5), then I expect only 5 stars to have a different background. in your case all 7 stars will have different background, not 5 as this modifier says

ihor-jpeg: again, if I have 7 stars per row, I use this modifier (stars--5), then I expect only 5 stars to have a different background. in your case all 7 stars will have different background, not 5 as this modifier says

ihor-jpeg (changes requested):

ihor-jpeg (approved):
Well done! 🚀

PR #1128, opened 2022-03-16
Initial commit

AndreiSuc:
- [DEMO LINK](https://andreisuc.github.io/layout_stars/)
- [TEST REPORT LINK](https://andreisuc.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,40 @@
+body {
+  margin: 0;
+}
+
+.stars {
+  display: flex;
+}
+
+.stars__star {
+  display: inline-block;
+  height: 12px;
+  width: 12px;
+  margin: 2px 6px 2px 2px;
+  background-image: url("/images/star.svg");
+  background-size: cover;
+}
+
+.stars__star:nth-child(n) {
+  background-image: url("/images/star-active.svg");
+}
+
+.stars--0 .stars__star {
+  background-image: url("/images/star.svg");
+}
+
+.stars--1 .stars__star:not(:first-child) {
+  background-image: url("/images/star.svg");
+}
+
+.stars--4 .stars__star:last-child {
+  background-image: url("/images/star.svg");
+}
+
+.stars--2 .stars__star:not(:first-child) + .stars__star {
+  background-image: url("/images/star.svg");
+}
+
+.stars--3 .stars__star:not(:first-child) + .stars__star + .stars__star {
+  background-image: url("/images/star.svg");
+}
--- a/readme.md
+++ b/readme.md
@@ -1,7 +1,7 @@
 # Stars block
 Replace `<your_account>` with your Github username and copy the links to Pull Request description:
-- [DEMO LINK](https://<your_account>.github.io/layout_stars/)
-- [TEST REPORT LINK](https://<your_account>.github.io/layout_stars/report/html_report/)
+- [DEMO LINK](https://andreisuc.github.io/layout_stars/)
+- [TEST REPORT LINK](https://andreisuc.github.io/layout_stars/report/html_report/)
 
 > Follow [this instructions](https://mate-academy.github.io/layout_task-guideline)
 ___
--- a/src/index.html
+++ b/src/index.html
@@ -8,6 +8,47 @@
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Stars</h1>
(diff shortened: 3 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,40 @@
...
+  height: 12px;
+  width: 12px;
+  margin: 2px 6px 2px 2px;
+  background-image: url("/images/star.svg");
+  background-size: cover;
+}
+
+.stars__star:nth-child(n) {
+  background-image: url("/images/star-active.svg");
+}
```
ihor-jpeg, line 20: it seems like you are just resetting the background image here.

ihor-jpeg (changes requested):
Well done! But I think this task can be done in an easier way.
If you set the default bg to be a gray start and then using :nth-child(-n + <row number>) set yellow start bg, you won't need the starts--0 modifier and the code will be simplier to understand

src/style.css
AndreiSuc: Yes, I have a talent for doing things backward. WIll try to fix this now.

src/style.css
```diff
@@ -0,0 +1,44 @@
+body {
+  margin: 0;
+}
+
+.stars {
+  display: flex;
+}
+
+.stars__star {
+  display: inline-block;
```
ihor-jpeg, line 10: after you give the parent `display: flex`, you cannot set `display` for direct children. i mean, technically you can, but it won't do anything
JFYI

src/style.css
```diff
@@ -0,0 +1,44 @@
...
+  width: 12px;
+  margin: 2px 6px 2px 2px;
+  background-image: url("/images/star.svg");
+  background-size: cover;
+}
+
+.stars__star:nth-child(n) {
+  background-image: url("/images/star.svg");
+}
+
```
ihor-jpeg, line 21: let's think, here you set the bg for every `.stars__star`, which you have already done in the selector right above

src/style.css
```diff
@@ -0,0 +1,44 @@
...
+.stars--0 .stars__star:nth-child(-n+0) {
+  background-image: url("/images/star-active.svg");
+}
```
ihor-jpeg, line 24: this one is not needed, because this selector basically says "do not set `background-image: url("/images/star-active.svg")` to any `.stars__star`"

this is because `n` will start from 1,  but you have `-n`, so it will be `:nth-child(-1)`, and then `:nth-child(-2)` and so on. and since there can be no negative children, the styles inside this selector will never apply

src/style.css
```diff
@@ -0,0 +1,44 @@
...
+
+.stars--4 .stars__star:nth-child(-n+4) {
+  background-image: url("/images/star-active.svg");
+}
+
+.stars--5 .stars__star:nth-child(-n+5) {
+  background-image: url("/images/star-active.svg");
+}
```
ihor-jpeg, line 44: let's group these, because the styles applied are the same

```suggestion
.stars--1 .stars__star:nth-child(-n+1),
.stars--2 .stars__star:nth-child(-n+2),
.stars--2 .stars__star:nth-child(-n+3),
.stars--2 .stars__star:nth-child(-n+4),
.stars--2 .stars__star:nth-child(-n+5), {
  background-image: url("/images/star-active.svg");
}
```

ihor-jpeg (changes requested):
Almost there, man! I can almost visualize the green "approved" check mark ✅

src/style.css
AndreiSuc: Makes sense.

src/style.css
AndreiSuc: Done.

AndreiSuc: I basically changed my old code per your suggestion without thinking much of what I'm writing. Need to pay more attention in the future.

ihor-jpeg (approved):
Well done! Many things regarding simplification and writing excess code can be avoided if you go into developer tools and start turning off some styles. if you see that you turned something off and everything still works, there is a good chance you don't need that line of code. It's okay that you are changing things based on what mentors say, but it's important to understand why you change that. If you do, then everything is fine, if not, ask questions, tag us in slack and annoy us until you get your answer. I'm not joking. 
Screenshot 2022-03-18 at 16 35 09

AndreiSuc:
image

PR #1129, opened 2022-03-16
Develop

leny-jeanbaptiste:


src/index.html
```diff
@@ -8,6 +8,54 @@
...
+        <div class="stars__star stars__active"></div>
+        <div class="stars__star"></div>
+      </div>
+
+      <div class="stars stars--5">
+        <div class="stars__star stars__active"></div>
+        <div class="stars__star stars__active"></div>
+        <div class="stars__star stars__active"></div>
+        <div class="stars__star stars__active"></div>
+        <div class="stars__star stars__active"></div>
```
ihor-jpeg, line 57: there should be only one class here `stars__start`. the bg should be controlled with the modifier of the parent (`stars--#number`)

ihor-jpeg (changes requested):
Good work, but let's implement it differently, I have left a comment. Also, please do not push the config files. Only readme, and files in the `src` folder

src/style.css
```diff
@@ -0,0 +1,40 @@
...
+  background-image: url(images/star-active.svg);
+}
+
+.stars--2 :nth-child(-n+2) {
+  background-image: url(images/star-active.svg);
+}
+
+.stars--3 :nth-child(-n+3) {
+  background-image: url(images/star-active.svg);
+}
```
ihor-jpeg, line 32: you can group the selectors

```suggestion
.stars--1 :nth-child(1),
.stars--2 :nth-child(-n+2),
.stars--2 :nth-child(-n+2) {
  background-image: url(images/star-active.svg);
}
```

src/style.css
```diff
@@ -0,0 +1,40 @@
...
+
+.stars--4 :nth-child(-n+4) {
+  background-image: url(images/star-active.svg);
+}
+
+.stars--5 :nth-child(n) {
+  background-image: url(images/star-active.svg);
+}
```
ihor-jpeg, line 40: based on this piece of code what will happen if I have 7 stars in one row?

ihor-jpeg (changes requested):
Almost there!

src/style.css
leny-jeanbaptiste: There will be 7 active stars

src/style.css
leny-jeanbaptiste: Aight got it

src/style.css
ihor-jpeg: yep! but the modifier says there should be only 5

src/style.css
```diff
@@ -0,0 +1,28 @@
...
+  display: flex;
+}
+
+.stars__star {
+  margin: 2px 6px 2px 2px;
+  width: 12px;
+  height: 12px;
+  background-image: url(images/star.svg);
+  background-position: cover;
+  display: inline-block;
```
ihor-jpeg, line 19: if the parent has `display: flex` then display for direct children will not be applied, so this is redundant
JFYI

ihor-jpeg (approved):
Well done! 🚀 
Have a look at the comment i have left below

PR #1132, opened 2022-03-17
Develop

MabuzaM:
- [DEMO LINK](https://MabuzaM.github.io/layout_stars/)
- [TEST REPORT LINK](https://MabuzaM.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -5,9 +5,14 @@
     <meta name="viewport"
           content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0">
     <title>Stars</title>
-    <link rel="stylesheet" href="./style.css">
+    <link rel="stylesheet" href="style.css">
   </head>
   <body>
-    <h1>Stars</h1>
+    <div class="stars stars--0"></div>
+    <div class="stars stars--1"></div>
+    <div class="stars stars--2"></div>
+    <div class="stars stars--3"></div>
+    <div class="stars stars--4"></div>
+    <div class="stars stars--5"></div>
   </body>
 </html>
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,58 @@
+html {
+  margin: 0;
+}
+
+.stars {
+  display: flex;
+}
+
+.stars--0 {
+  background-image: url("./images/star.svg"),
+    url("./images/star.svg"),
+    url("./images/star.svg"),
+    url("./images/star.svg"),
+    url("./images/star.svg");
+  background-size: auto;
+  background-repeat: no-repeat;
+  background-attachment: fixed;
+}
+
+body > .stars--1 {
+  background-image: url("./images/star-active.svg"),
+    url("./images/star.svg"),
+    url("./images/star.svg"),
+    url("./images/star.svg"),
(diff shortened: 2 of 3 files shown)


src/index.html
```diff
@@ -5,9 +5,55 @@
     <meta name="viewport"
           content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0">
     <title>Stars</title>
-    <link rel="stylesheet" href="./style.css">
+    <link rel="stylesheet" href="style.css">
   </head>
   <body>
-    <h1>Stars</h1>
+    <div class="stars stars--0">
```
ihor-jpeg, line 11: i don't think `stars--0` modifier is needed

src/style.css
```diff
@@ -0,0 +1,34 @@
+* {
+  margin: 0;
+}
```
ihor-jpeg, line 3: do not reset margins like this. only do it for the tags that you actually need to reset margins for

src/style.css
```diff
@@ -0,0 +1,34 @@
...
+  background-image: url("./images/star-active.svg");
+}
+
+.stars--4 .stars__star:nth-child(-n+4) {
+  background-image: url("./images/star-active.svg");
+}
+
+.stars--5 .stars__star:nth-child(-n+5) {
+  background-image: url("./images/star-active.svg");
+}
```
ihor-jpeg, line 34:
```suggestion
.stars--1 .stars__star:nth-child(-n+1),
.stars--2 .stars__star:nth-child(-n+2),
.stars--2 .stars__star:nth-child(-n+3),
.stars--2 .stars__star:nth-child(-n+4),
.stars--2 .stars__star:nth-child(-n+5) {
  background-image: url("./images/star-active.svg");
}
```

ihor-jpeg (changes requested):
Nicely done! Let's fix a few things. I have left comments below

src/index.html
MabuzaM: okay noted.... let me remove it

src/style.css
MabuzaM: Owkay noted.... How do I reset the browser's default margin? Do I use the html or body selector? and do I set it to 0 or 8px?

src/style.css
MabuzaM: Okay done... Wow! it's quite shorter

src/style.css
MabuzaM: Okay thanks had to set the body margin to 0 for the test to pass

src/index.html
MabuzaM: It seems the test requires it to search for the first block.

ihor-jpeg: yeah, damn tests

src/style.css
ihor-jpeg: generally you want to reset margins for tag that you know you need margin 0.
for example: 

`body,
  p,
  ul {
    margin: 0;
  }`

something like that

ihor-jpeg (approved):
Nicely done! 🚀

src/style.css
ihor-jpeg: but yeah, in this case you only needed to reset it for the body

PR #1133, opened 2022-03-18
WIP

proznerka:
- [DEMO LINK](https://proznerka.github.io/layout_stars/)
- [TEST REPORT LINK](https://proznerka.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -8,6 +8,12 @@
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Stars</h1>
+    <div class="container">
+      <div class="star--1"></div>
+      <div class="star--2"></div>
+      <div class="star--3"></div>
+      <div class="star--4"></div>
+      <div class="star--5"></div>
+    </div>
   </body>
 </html>
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -19,5 +19,11 @@ jobs:
       uses: actions/setup-node@v1
       with:
         node-version: ${{ matrix.node-version }}
-    - run: npm ci
-    - run: npm start & sleep 5 && npm test
+    - run: npm install
+    - run: npm test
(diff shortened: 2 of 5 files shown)


src/index.html
```diff
@@ -8,6 +8,12 @@
...
   </head>
   <body>
-    <h1>Stars</h1>
+    <div class="container">
+      <div class="star--1"></div>
+      <div class="star--2"></div>
+      <div class="star--3"></div>
+      <div class="star--4"></div>
+      <div class="star--5"></div>
+    </div>
```
ihor-jpeg, line 17: now, these divs should also have the `stars` class. also, put five stars with class `stars__star` inside each div

src/index.html
```diff
@@ -8,6 +8,12 @@
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Stars</h1>
+    <div class="container">
+      <div class="star--1"></div>
```
ihor-jpeg, line 12: at the moment you have five containers, each will have colored stars and the number of colored stars should be equal to the number on the class modifier. the first row of stars should have 0 stars colored. so before `<div class="star--1"></div>` you need to add another div. something like <div class="stars"></div>

src/index.html
```diff
@@ -8,6 +8,12 @@
...
+      <div class="star--2"></div>
+      <div class="star--3"></div>
```
ihor-jpeg, line 14: each of these should also have `stars` class

```suggestion
      <div class="stars stars--2"></div>
      <div class="stars stars--3"></div>
```

ihor-jpeg (changes requested):
Nice start! Keep it going! 🚀

ihor-jpeg (changes requested):
I have left a few comments above

ihor-jpeg (approved):
Nicely done! 🔥 🚒

PR #1134, opened 2022-03-18
stars

elkhayate:
- [DEMO LINK](https://elkhayate.github.io/layout_stars/)
- [TEST REPORT LINK](https://elkhayate.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,30 @@
+* {
+  margin: 0;
+  box-sizing: border-box;
+}
+
+.stars {
+  display: flex;
+  height: 16px;
+  width: auto;
+}
+
+.stars__star {
+  background-image: url(images/star.svg);
+  height: 12px;
+  width: 13px;
+  margin-top: 2px;
+  margin-bottom: 2px;
+  margin-left: 2px;
+  margin-right: 5px;
+  background-size: auto;
+  background-position: center;
+}
+
+.stars--1 .stars__star:nth-child(1),
+.stars--2 .stars__star:nth-child(-n+2),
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,30 @@
+* {
+  margin: 0;
```
ihor-jpeg, line 2: do not reset margins with the `*` selector. do it only for the tags that you need to reset margins for

src/style.css
```diff
@@ -0,0 +1,30 @@
...
+  box-sizing: border-box;
+}
+
+.stars {
+  display: flex;
+  height: 16px;
+  width: auto;
```
ihor-jpeg, line 9: if you change the size of an individual star you would also need to change the size here. if you hardcode some values, do it only in one place. for example, if you have a parent and a child elements, then only set the size for the child. no need to do it for both

src/style.css
```diff
@@ -0,0 +1,30 @@
...
+}
+
+.stars__star {
+  background-image: url(images/star.svg);
+  height: 12px;
+  width: 13px;
+  margin-top: 2px;
+  margin-bottom: 2px;
+  margin-left: 2px;
+  margin-right: 5px;
```
ihor-jpeg, line 19: use `margin` shorthand property

ihor-jpeg (changes requested):
Well done! But let's fix a few small issues. I have left comments below

ihor-jpeg (changes requested):
Almost there!

PR #1135, opened 2022-03-18
Stars, made to pass the test

WojciechCzarnowski:
- [DEMO LINK](https://WojciechCzarnowski.github.io/layout_stars/)
- [TEST REPORT LINK](https://WojciechCzarnowski.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,29 @@
+.stars {
+  display: flex;
+}
+
+.stars__star {
+  /* display: inline-block; */
+  height: 16px;
+  width: 16px;
+  margin-right: 6px;
+  background-image: url(images/star.svg);
+  /* background-size:; */
+  background-repeat: no-repeat;
+  background-position: center;
+}
+
+.stars__star:nth-child(n) {
+  background-image: url(images/star-active.svg);
+}
+
+.stars--0 .stars__star {
+  background-image: url(images/star.svg);
+}
+
+.stars--1 .stars__star:nth-child(n + 2),
+.stars--2 .stars__star:nth-child(n + 3),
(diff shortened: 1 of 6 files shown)


src/style.css
```diff
@@ -0,0 +1,31 @@
...
+  background-position: center;
+}
+
+.stars__star:nth-child(n) {
+  background-image: url(images/star-active.svg);
+}
+
+.stars--0 .stars__star {
+  background-image: url(images/star.svg);
+}
```
ihor-jpeg, line 24: this one seems to be redundant because all `stars__star` have bg of `images/star.svg`. the only reason you might need it is the previous selector.

src/style.css
```diff
@@ -0,0 +1,31 @@
...
+  width: 16px;
+  margin-right: 4px;
+  background-image: url(images/star.svg);
+  background-repeat: no-repeat;
+  background-position: center;
+}
+
+.stars__star:nth-child(n) {
+  background-image: url(images/star-active.svg);
+}
```
ihor-jpeg, line 20: oooooh, I see now. you won't need this if you use a slightly different formula for the :nth-child()

ihor-jpeg (changes requested):
Well done! The logic behind bg for stars is a bit confusing. You have set bg for every star, then you change it to star-active, and then in the :nth-child() selector you change it back to normal. I think it would make more sense to set the default bg of star or star-active and change it to the opposite one in :nth-child()

ihor-jpeg (commented):
Well done! 🚀

ihor-jpeg (approved):

PR #1136, opened 2022-03-18
Stars layout

Sardorbek1122:
- [DEMO LINK](https://Sardorbek1122.github.io/layout_stars/)
- [TEST REPORT LINK](https://Sardorbek1122.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,29 @@
+* {
+  margin: 0;
+  padding: 0;
+  box-sizing: border-box;
+}
+
+.stars {
+  display: flex;
+}
+
+.stars_star {
+  display: inline-block;
+  margin: 2px 6px 2px 2px;
+  background-image: url(./images/star.svg);
+  background-repeat: no-repeat;
+  background-size: cover;
+  border-radius: 0.5%;
+  width: 12px;
+  height: 12px;
+}
+
+.stars--1 :nth-child(1),
+.stars--2 :nth-child(-n+2),
+.stars--3 :nth-child(-n+3),
+.stars--4 :nth-child(-n+4),
+.stars--5 :nth-child(-n+5),
+.stars--6 :nth-child(-n+6) {
+  background-image: url(images/star-active.svg);
+}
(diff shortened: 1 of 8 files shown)


src/style.css
```diff
@@ -0,0 +1,29 @@
+* {
+  margin: 0;
+  padding: 0;
```
ihor-jpeg, line 3: it is not a good idea to reset margins and paddings here

src/style.css
```diff
@@ -0,0 +1,29 @@
...
+  box-sizing: border-box;
+}
+
+.stars {
+  display: flex;
+}
+
+.stars_star {
+  display: inline-block;
```
ihor-jpeg, line 12: display property does not work for elements inside a `flex` container

src/style.css
```diff
@@ -0,0 +1,29 @@
...
+  margin: 2px 6px 2px 2px;
+  background-image: url(./images/star.svg);
+  background-repeat: no-repeat;
+  background-size: cover;
+  border-radius: 0.5%;
```
ihor-jpeg, line 17: it is not very clear why you have used percents here. use pixels instead

src/style.css
```diff
@@ -0,0 +1,29 @@
...
+  width: 12px;
+  height: 12px;
+}
+
+.stars--1 :nth-child(1),
+.stars--2 :nth-child(-n+2),
+.stars--3 :nth-child(-n+3),
+.stars--4 :nth-child(-n+4),
+.stars--5 :nth-child(-n+5),
+.stars--6 :nth-child(-n+6) {
```
ihor-jpeg, line 27: no need for the `.stars--6` modifier

ihor-jpeg (changes requested):
Good work. I have left a few comments for you to fix

src/style.css
Sardorbek1122: I did not attend when I was checking figma,

src/style.css
```diff
@@ -0,0 +1,27 @@
+body {
+  box-sizing: border-box;
+  margin: 0;
+  padding: 0;
```
ihor-jpeg, line 4: body does not have any default paddings

ihor-jpeg (approved):
Good work! But check my comment below

PR #1137, opened 2022-03-18
Solution

EnjSadman:
[DEMO LINK](https://EnjSadman.github.io/layout_stars/)
[TEST REPORT LINK](https://EnjSadman.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,45 @@
+* {
+  margin: 0;
+  padding: 0;
+}
+
+.stars_row {
+  display: flex;
+  height: 100%;
+  width: 100%;
+}
+
+.stars {
+  height: 16px;
+  width: 16px;
+  display: flex;
+  background: url(./images/star-active.svg) no-repeat center;
+}
+
+.stars:not(:last-child) {
+  margin-right: 4px;
+}
+
+.stars--0 > .stars {
+  background: url(./images/star.svg) no-repeat center;
+}
+
+.stars--1 > .stars:nth-child(n+2) {
+  background: url(./images/star.svg) no-repeat center;
+}
+
+.stars--2 > .stars:nth-child(n+3) {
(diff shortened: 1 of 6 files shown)


src/style.css
```diff
@@ -0,0 +1,45 @@
+* {
+  margin: 0;
+  padding: 0;
+}
```
Denchepornyuk, line 4: don't reset styles in `*` selector
not every tag has margin or padding

src/style.css
```diff
@@ -0,0 +1,45 @@
...
+
+.stars_row {
+  display: flex;
+  height: 100%;
+  width: 100%;
```
Denchepornyuk, line 9: height and width are reduntant for html block element
height and width by default are 100%

Screenshot 2022-03-21 at 17 09 30

src/style.css
```diff
@@ -0,0 +1,45 @@
...
+
+.stars--3 > .stars:nth-child(n+4) {
+  background: url(./images/star.svg) no-repeat center;
+}
+
+.stars--4 > .stars:nth-child(n+5) {
+  background: url(./images/star.svg) no-repeat center;
+}
+
+.stars--5 > .stars:nth-child(n+6) {
```
Denchepornyuk, line 43: you can combine all these selectors into one selector

Denchepornyuk (changes requested):

src/style.css
```diff
@@ -0,0 +1,28 @@
+body {
+  margin: 0;
+  padding: 0;
```
Denchepornyuk, line 3: body doesn't have default padding, remove `padding: 0;` from body

Denchepornyuk (changes requested):

Denchepornyuk (approved):

PR #1142, opened 2022-03-21
Test failed, what should I do?

yelnikov-andrii:


src/style.css
```diff
@@ -0,0 +1,62 @@
+* {
+  all: initial;
+  display: revert;
```
Denchepornyuk, line 3: don't add these styles to `*` you don't need them

src/style.css
```diff
@@ -0,0 +1,62 @@
...
+  background-repeat: no-repeat;
+  background-position: center;
+
+}
+
+.stars--1 .stars__star:first-child {
+  background-image: url(./images/star-active.svg);
+  background-size: auto;
+  background-repeat: no-repeat;
+  background-position: center;
```
Denchepornyuk, line 33: you have already added these styles to `stars__star`
you can remove them here

src/style.css
```diff
@@ -0,0 +1,62 @@
...
+}
+
+.stars--3 .stars__star:nth-child(-n+3) {
+  background-image: url(./images/star-active.svg);
+  background-size: auto;
+  background-repeat: no-repeat;
+  background-position: center;
+}
+
+.stars--4 .stars__star:nth-child(-n+4) {
```
Denchepornyuk, line 50: combine all these selectors into one selector

Denchepornyuk (changes requested):

PR #1166, opened 2022-03-30
add code for stars task

v-vytrykush:
- [DEMO LINK](https://v-vytrykush.github.io/layout_stars/)
- [TEST REPORT LINK](https://v-vytrykush.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,26 @@
+* {
+  margin: 0;
+}
+
+.star {
+  width: 16px;
+  height: 16px;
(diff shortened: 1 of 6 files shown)


src/style.css
```diff
@@ -0,0 +1,26 @@
+* {
+  margin: 0;
+}
```
Denchepornyuk, line 3: don't reset margins in `*` selector
add it to body

Denchepornyuk (approved):

PR #1177, opened 2022-03-31
Task 'Stars" is done

enrikadej:
- [DEMO LINK](https://enrikadej.github.io/layout_stars/)
- [TEST REPORT LINK](https://enrikadej.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,32 @@
+body {
+  margin: 0;
+}
+
+.container {
+  display: flex;
+  flex-direction: column;
+}
+
+.stars {
+  display: flex;
+  flex-direction: row;
+}
+
+.stars__star {
+
+  background: url(images/star.svg);
+  background-repeat: no-repeat;
+
+  width: 16px;
+  height: 16px;
+  margin: 0 4px;
+}
+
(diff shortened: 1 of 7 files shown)


src/style.css
```diff
@@ -0,0 +1,32 @@
...
+  margin: 0 4px;
+}
+
+.stars--1 > :nth-child(-n + 1),
+.stars--2 > :nth-child(-n + 2),
+.stars--3 > :nth-child(-n + 3),
+.stars--4 > :nth-child(-n + 4),
+.stars--5 > :nth-child(-n + 5) {
+  background: url(images/star-active.svg);
+  background-repeat: no-repeat;
```
Denchepornyuk, line 31: remove background-repeat, you have already added this rule for `.stars__star`

Denchepornyuk (changes requested):

src/style.css
enrikadej: .stars__star has `background-repeat: no-repeat` for gray stars
and another one for yellow stars
If I remove it, my layout will be incorrect

Denchepornyuk (approved):

PR #1180, opened 2022-04-02
Stars

Alyona-Pisotska:
https://Alyona-Pisotska.github.io/layout_stars/
https://Alyona-Pisotska.github.io/layout_stars/report/html_report/

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,52 @@
+* {
+  margin: 0;
+  padding: 0;
+}
+
+.stars {
+  display: flex;
+}
+
+.stars .stars__star {
+  height: 16px;
+  width: 16px;
+  margin-right: 4px;
+}
+
+.stars--0 .stars__star {
+  background: url(./images/star.svg) no-repeat;
+  background-position: center;
+}
+
+.stars--1 .stars__star,
+.stars--2 .stars__star,
+.stars--3 .stars__star,
+.stars--4 .stars__star,
+.stars--5 .stars__star {
+  background: url(./images/star-active.svg) no-repeat;
+  background-position: center;
+}
+
+.stars--1 .stars__star:nth-child(n + 2) {
+  background: url(./images/star.svg) no-repeat;
+  background-position: center;
+}
+
+.stars--2 .stars__star:nth-child(n + 3) {
+  background: url(./images/star.svg) no-repeat;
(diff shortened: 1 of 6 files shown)


src/style.css
```diff
@@ -0,0 +1,52 @@
...
+
+.stars .stars__star {
+  height: 16px;
+  width: 16px;
+  margin-right: 4px;
+}
+
+.stars--0 .stars__star {
+  background: url(./images/star.svg) no-repeat;
+  background-position: center;
```
aliceblunt8, line 18: add default background-position, background-repeat and background-image (not active star) to each star (.stars .stars__star) and then just rewrite background-image with active star for each modifier

src/style.css
```diff
@@ -0,0 +1,52 @@
...
+}
+
+.stars--1 .stars__star,
+.stars--2 .stars__star,
+.stars--3 .stars__star,
+.stars--4 .stars__star,
+.stars--5 .stars__star {
+  background: url(./images/star-active.svg) no-repeat;
+  background-position: center;
+}
```
aliceblunt8, line 28: remove these it's useless, each star should have a simple star as default background, then use nth-child to add active star

aliceblunt8 (changes requested):

src/style.css
Alyona-Pisotska: @aliceblunt8 Done

src/style.css
```diff
@@ -0,0 +1,41 @@
...
+  background-image: url(./images/star-active.svg);
+}
+
+.stars--3 .stars__star:nth-child(-n + 3) {
+  background-image: url(./images/star-active.svg);
+}
+
+.stars--4 .stars__star:nth-child(-n + 4) {
+  background-image: url(./images/star-active.svg);
+}
```
ArtemTopolskyi, line 37: you add the same css rule for all these 5 selectors, you can merge them:
```
.stars-5,
.stars-4,
.stars-3 {
  my-background-rule: #something;
}
```

ArtemTopolskyi (changes requested):

src/style.css
Alyona-Pisotska: @ArtemTopolskyi Done

ArtemTopolskyi (approved):

PR #1182, opened 2022-04-02
stars

ISRLv2021:
- [DEMO LINK](https://ISRLv2021.github.io/layout_stars/)
- [TEST REPORT LINK](https://ISRLv2021.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,30 @@
+body {
+  margin: 0;
+}
+
+.stars {
+  display: flex;
+  margin: 0;
+}
+
+.star {
+  width: 16px;
+  height: 16px;
+  margin: 0 2px;
+  
+  background: url(./images/star.svg) no-repeat;
+  background-position: center;
+}
+
+.stars > :first-child {
+  margin-left: 0;
+}
+
+.stars--1 .star:nth-child(1),
+.stars--2 .star:nth-child(-n+2),
+.stars--3 .star:nth-child(-n+3),
+.stars--4 .star:nth-child(-n+4),
+.stars--5 .star:nth-child(-n+5) {
+  background: url(./images/star-active.svg) no-repeat;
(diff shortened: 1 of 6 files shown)


src/style.css
```diff
@@ -0,0 +1,30 @@
+body {
+  margin: 0;
+}
+
+.stars {
+  display: flex;
+  margin: 0;
```
Anna-Tkachenko, line 7: This block has margin: 0 by default, remove this styles

src/style.css
```diff
@@ -0,0 +1,30 @@
...
+  margin-left: 0;
+}
+
+.stars--1 .star:nth-child(1),
+.stars--2 .star:nth-child(-n+2),
+.stars--3 .star:nth-child(-n+3),
+.stars--4 .star:nth-child(-n+4),
+.stars--5 .star:nth-child(-n+5) {
+  background: url(./images/star-active.svg) no-repeat;
+  background-position: center;
```
Anna-Tkachenko, line 29: Move background-position and background-repeat properties to the main star class
You should change only background-image for active stars

Anna-Tkachenko (changes requested):

src/style.css
ISRLv2021: Right! Thank you

Anna-Tkachenko (approved):
Good job!

src/style.css
```diff
@@ -0,0 +1,24 @@
...
+}
+
+.stars {
+  display: flex;
+}
+
+.star {
+  width: 16px;
+  height: 16px;
+  margin-right: 4px;
```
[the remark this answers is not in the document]
ISRLv2021: .star:not(nth:last-child) { 
  margin-right: 4px;
}
would this solution work?

PR #1184, opened 2022-04-04
First_commit_Stars

igoree88:
- [DEMO LINK](https://igoree88.github.io/layout_stars/)
- [TEST REPORT LINK](https://igoree88.github.io/layout_stars/report/html_report/)

I passed the first test perfectly, but the other tests are slightly different from the reference sample.

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,45 @@
+* {
+  margin: 0;
+}
+
+.container {
+  display: flex;
+  flex-direction: column;
+  width: 96px;
+  height: 96px;
+  justify-content: space-between;
+}
+
+.stars {
+  display: flex;
+  justify-content: space-between;
+
+}
+
+.stars .stars__star {
+  background-image: url(./images/star.svg);
+  background-repeat: no-repeat;
+  background-position: center;
+  height: 16px;
+  width: 16px;
+}
+
+.stars--1 .stars__star:first-child {
+  background-image: url(./images/star-active.svg);
+}
+
+.stars--2 .stars__star:nth-child(-n+2) {
+  background-image: url(./images/star-active.svg);
+}
+
+.stars--3 .stars__star:nth-child(-n+3) {
+  background-image: url(./images/star-active.svg);
+}
+
+.stars--4 .stars__star:nth-child(-n+4) {
+  background-image: url(./images/star-active.svg);
+}
+
+.stars--5 .stars__star:nth-child(n) {
+  background-image: url(./images/star-active.svg);
+}
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -19,5 +19,11 @@ jobs:
       uses: actions/setup-node@v1
       with:
         node-version: ${{ matrix.node-version }}
-    - run: npm ci
-    - run: npm start & sleep 5 && npm test
+    - run: npm install
+    - run: npm test
+    - name: Upload HTML report(backstop data)
+      if: ${{ always() }}
+      uses: actions/upload-artifact@v2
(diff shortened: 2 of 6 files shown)


src/style.css
```diff
@@ -0,0 +1,45 @@
+* {
+  margin: 0;
+}
```
Anna-Tkachenko, line 3: Reset margin only for body

src/style.css
```diff
@@ -0,0 +1,45 @@
...
+
+.container {
+  display: flex;
+  flex-direction: column;
+  width: 96px;
+  height: 96px;
```
Anna-Tkachenko, line 9: Don't hardcode width and height for main container

src/style.css
```diff
@@ -0,0 +1,45 @@
...
+  justify-content: space-between;
+}
+
+.stars {
+  display: flex;
+  justify-content: space-between;
```
Anna-Tkachenko, line 15: Delete this style, you should only add correct margin to each star

src/style.css
```diff
@@ -0,0 +1,45 @@
...
+
+}
+
+.stars .stars__star {
```
Anna-Tkachenko, line 19: Don't increase specificity using 2 classes in one selector
Use only .stars__star class

src/style.css
```diff
@@ -0,0 +1,45 @@
...
+  background-image: url(./images/star.svg);
+  background-repeat: no-repeat;
+  background-position: center;
```
Anna-Tkachenko, line 22: Set correct margin to each star (4 px)

src/style.css
```diff
@@ -0,0 +1,45 @@
...
+  background-image: url(./images/star-active.svg);
+}
+
+.stars--4 .stars__star:nth-child(-n+4) {
+  background-image: url(./images/star-active.svg);
+}
+
+.stars--5 .stars__star:nth-child(n) {
+  background-image: url(./images/star-active.svg);
+}
```
Anna-Tkachenko, line 45: Combine this selectors using coma

Anna-Tkachenko (changes requested):

src/style.css
igoree88: Fixed

src/style.css
igoree88: Fixed

src/style.css
igoree88: Fixed

src/style.css
igoree88: fixed

src/style.css
igoree88: fixed

src/style.css
igoree88: Fixed

src/style.css
```diff
@@ -0,0 +1,30 @@
...
+.stars {
+  display: flex;
+}
+
+.stars__star {
+  background-image: url(./images/star.svg);
+  background-repeat: no-repeat;
+  height: 16px;
+  width: 16px;
+  margin: 0 2px;
```
Anna-Tkachenko, line 20: Set only margin-right: 4px for all stars

src/style.css
```diff
@@ -0,0 +1,30 @@
...
+  justify-content: space-between;
+}
+
+.stars {
+  display: flex;
+}
+
+.stars__star {
+  background-image: url(./images/star.svg);
+  background-repeat: no-repeat;
```
Anna-Tkachenko, line 17: Add background-position: center

Anna-Tkachenko (approved):

PR #1186, opened 2022-04-06
stars 1

YarinKiri2l:
- [DEMO LINK](https://yarinkiri2l.github.io/layout_stars/)
- [TEST REPORT LINK](https://yarinkiri2l.github.io/layout_stars/report/html_report/)


src/style.css
```diff
@@ -0,0 +1,29 @@
...
+  justify-content: space-between;
+}
+
+.stars {
+  display: flex;
+}
+
+.stars__star {
+  background-image: url(/src/images/star.svg);
+  background-repeat: no-repeat;
```
Anna-Tkachenko, line 17: Don't forget to add background-position: center

src/style.css
```diff
@@ -0,0 +1,29 @@
+body {
+  margin: 0;
+}
+
+.container {
+  display: flex;
+  flex-direction: column;
+  justify-content: space-between;
```
Anna-Tkachenko, line 8: flex-direction is redundant
all blocks are started from the new line by default
same with justify-content: delete it

Anna-Tkachenko (approved):
Good job!
Just fix demo link and test report link

PR #1189, opened 2022-04-07
First Commit

DimaK123:
- [DEMO LINK](https://Dimak123.github.io/layout_stars/)
- [TEST REPORT LINK](https://Dimak123.github.io/layout_stars/report/html_report/)


src/style.css
```diff
@@ -0,0 +1,58 @@
...
+  background-repeat: no-repeat;
+  background-position: center;
+}
+
+.stars__star:not(:last-child) {
+  margin-right: 4px;
+}
+
+.stars--0 > .stars__star {
+  background-image: url(./images/star.svg);
```
aliceblunt8, line 21: add background-image with inactive stars by default to each star

aliceblunt8: .stars__star

src/style.css
```diff
@@ -0,0 +1,58 @@
...
+  background-image: url(./images/star-active.svg);
+}
+
+.stars--2 > .stars__star:nth-child(n+3) {
+  background-image: url(./images/star.svg);
+}
+
+.stars--3 > .stars__star:nth-child(-n + 3) {
+  background-image: url(./images/star-active.svg);
+}
```
aliceblunt8, line 46: combine all selectors with background-image with active stars

aliceblunt8 (changes requested):

src/style.css
DimaK123: I've fixed problem, can you check?

PR #1194, opened 2022-04-11
Develop

frontend112:
https://frontend112.github.io/layout_stars/
https://frontend112.github.io/layout_stars/report/html_report/


src/style.css
```diff
@@ -0,0 +1,46 @@
+* {
+  margin: 0;
+}
+
+/* .container {
+  width: 96px;
+} */
+
```
meforoma, line 8: don't keep commented out code

src/style.css
```diff
@@ -0,0 +1,46 @@
...
+.stars--4 :nth-last-child(n + 2) {
+  background: url(images/star-active.svg) no-repeat;
+  background-position: left 25% center;
+}
+
+.stars--5 > * {
+  background: url(images/star-active.svg) no-repeat;
+  background-position: left 25% center;
+}
```
meforoma, line 46: you can stack similar styling
and no need to duplicate `background-position: left 25% center;` - you defined it already
```suggestion
.stars--1 :nth-child(1),
.stars--2 :nth-last-child(n + 4),
.stars--3 :nth-last-child(n + 3),
.stars--4 :nth-last-child(n + 2),
.stars--5 > * {
  background: url(images/star-active.svg) no-repeat;
}
```

meforoma (changes requested):
better avoid `git commit -am` - since it bulks up `package-lock.json` file that shouldn't be pushed
better `git add src/.` and then `git commit ...`

src/style.css
frontend112: I added this with shorthand, like this:
background: left 25% center url(images/star-active.svg) no-repeat;
because after apply this suggestion yellow stars moved to top left, 
I think that because of default background-position is top 0 and left 0

meforoma (approved):
👍

PR #1242, opened 2022-05-05
i was create block-stars

sergey-vorobei:
- [DEMO LINK](https://sergey-vorobei.github.io/layout_stars/)
- [TEST REPORT LINK](https://sergey-vorobei.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -8,6 +8,150 @@
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Stars</h1>
+    <div class="stars-block">
+      <div class="stars--0">
+        <ul class="stars-item">
+          <li class="stars__star">
+            <a class="link-star" href="#"></a>
+          </li>
+  
+          <li class="stars__star">
+            <a href="#"></a>
+          </li>
+  
+          <li class="stars__star">
+            <a href="#"></a>
+          </li>
+  
+          <li class="stars__star">
+            <a href="#"></a>
+          </li>
+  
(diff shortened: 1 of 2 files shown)


aliceblunt8 (changes requested):
remove is-active class from active stars and do this according to the task

Screenshot 2022-05-06 at 16 45 24

sergey-vorobei:
Исправил ошибку.

src/style.css
```diff
@@ -0,0 +1,31 @@
+body {
+  margin: 0;
+}
+
+.stars-item {
```
Bohdan-Maksymiuk, line 5:
```suggestion
.stars {
```

src/index.html
```diff
@@ -8,6 +8,150 @@
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Stars</h1>
+    <div class="stars-block">
+      <div class="stars stars--0">
+        <ul class="stars-item">
```
Bohdan-Maksymiuk, line 13: no need in this additional class
and you can use only <div> instead of unnecessary ul and li

Bohdan-Maksymiuk (changes requested):
why do you need `<a />`?

sergey-vorobei:
I thinking stars must be links

sergey-vorobei:
Сделал исправления по вашим замечаниям

aliceblunt8 (approved):

PR #1247, opened 2022-05-06
Stars layout

dimabelotskyi:
- [DEMO LINK](https://dimabelotskyi.github.io/layout_stars/)
- [TEST REPORT LINK](https://dimabelotskyi.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -8,16 +8,7 @@
   flex-direction: row;
 }
 
-.stars__star:not(:first-child) {
-  margin-left: 4px;
-}
-
-.stars--0 > .stars__star,
-.stars--1 > .stars__star,
-.stars--2 > .stars__star,
-.stars--3 > .stars__star,
-.stars--4 > .stars__star,
-.stars--5 > .stars__star {
+.stars__star {
   height: 16px;
   width: 16px;
   background-image: url(./images/star.svg);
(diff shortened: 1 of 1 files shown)


src/style.css
```diff
@@ -0,0 +1,124 @@
+/* #region General styles + Stars margin */
```
Olexandr6, line 1: Delete this regions, these are redundant in the production code

src/style.css
```diff
@@ -0,0 +1,124 @@
...
+}
+
+.stars--5 > .stars__star:nth-child(-n+5) {
+  height: 16px;
+  width: 16px;
+  background-image: url(./images/star-active.svg);
+  background-position: center;
+  /* background-size: contain; */
+  background-repeat: no-repeat;
+}
```
Olexandr6, line 122: How do you think it is okay to duplicate many such styles?  [Checklist](https://github.com/mate-academy/layout_stars/blob/master/checklist.md) 2nd point

Olexandr6 (changes requested):

src/style.css
```diff
@@ -0,0 +1,38 @@
...
+.stars--2 > .stars__star:nth-child(-n+2),
+.stars--3 > .stars__star:nth-child(-n+3),
+.stars--4 > .stars__star:nth-child(-n+4),
+.stars--5 > .stars__star:nth-child(-n+5) {
+  height: 16px;
+  width: 16px;
+  background-image: url(./images/star-active.svg);
+  background-position: center;
+  background-repeat: no-repeat;
+}
```
illiarerun, line 38: one of these, should be a default style for your `stars__star` class

src/style.css
```diff
@@ -0,0 +1,38 @@
...
+
+.stars--1 > .stars__star:nth-child(-n+1),
+.stars--2 > .stars__star:nth-child(-n+2),
+.stars--3 > .stars__star:nth-child(-n+3),
+.stars--4 > .stars__star:nth-child(-n+4),
+.stars--5 > .stars__star:nth-child(-n+5) {
+  height: 16px;
+  width: 16px;
+  background-image: url(./images/star-active.svg);
```
illiarerun, line 35: and all you will need to change in this huge rule is only background image

illiarerun (changes requested):

src/style.css
```diff
@@ -0,0 +1,29 @@
+* {
+  box-sizing: border-box;
+  margin: 0;
```
illiarerun, line 3: don't reset your margins with * selector, it's a bad practice
```suggestion
```

PR #1255, opened 2022-05-06
five stars

Grodh06:
- [DEMO LINK](https://Grodh06.github.io/layout_stars/)
- [TEST REPORT LINK](https://Grodh06.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,38 @@
+body {
+  margin: 0;
+}
+
+.stars {
+  display: flex;
+}
+
+.stars--star {
+  display: block;
+  width: 16px;
+  height: 16px;
+  margin: 0 2px;
+  background-image: url(images/star-active.svg);
+  background-repeat: no-repeat;
+  background-position: center;
+
+}
+
+.stars--0 .stars--star:nth-child(n) {
+  background-image: url(images/star.svg);
+}
+
+.stars--1 .stars--star:nth-child(n + 2) {
+  background-image: url(images/star.svg);
+}
+
+.stars--2 .stars--star:nth-child(n + 3) {
+  background-image: url(images/star.svg);
+}
+
+.stars--3 .stars--star:nth-child(n + 4) {
+  background-image: url(images/star.svg);
+}
+
+.stars--4 .stars--star:nth-child(n + 5) {
+  background-image: url(images/star.svg);
+}
--- a/src/style.css
+++ b/src/style.css
@@ -2,6 +2,10 @@ body {
   margin: 0;
 }
 
+h1 {
+  display: none;
+}
+
 .stars {
   display: flex;
 }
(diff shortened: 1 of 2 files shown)


src/style.css
```diff
@@ -0,0 +1,50 @@
...
+h1 {
+  display: none;
+}
+
+.stars {
+  display: flex;
+}
+
+.stars--star {
+  display: block;
```
jstmpelowycz, line 14: You don't need `display: block;` here, `div` is a block element by default

src/style.css
```diff
@@ -0,0 +1,50 @@
+body {
+  margin: 0;
+}
+
+h1 {
+  display: none;
+}
```
jstmpelowycz, line 7: That's a bad idea to remove all `h1` from the page, just remove the correspondent `h1` in your `index.html`

src/style.css
```diff
@@ -0,0 +1,50 @@
...
+
+.stars--star {
+  display: block;
+  width: 16px;
+  height: 16px;
+  margin: 0 2px;
+  background-image: url(images/star-active.svg);
+  background-repeat: no-repeat;
+  background-position: center;
+
```
jstmpelowycz, line 21: Don't forget to remove extra spaces from your code

src/style.css
```diff
@@ -0,0 +1,50 @@
...
+}
+
+.stars .stars--star:first-child {
+  margin-left: 0;
+}
+
+.stars .stars--star:last-child {
+  margin-right: 0;
+}
```
jstmpelowycz, line 30: Don't repeat yourself, combine it

src/style.css
```diff
@@ -0,0 +1,50 @@
...
+  background-image: url(images/star.svg);
+}
+
+.stars--3 .stars--star:nth-child(n + 4) {
+  background-image: url(images/star.svg);
+}
+
+.stars--4 .stars--star:nth-child(n + 5) {
+  background-image: url(images/star.svg);
+}
```
jstmpelowycz, line 50: The same issue with duplicating styles here

jstmpelowycz (changes requested):
A few fixes to apply, and you're done!

src/style.css
```diff
@@ -0,0 +1,28 @@
...
+  height: 16px;
+  margin-left: 4px;
+  background-image: url(images/star.svg);
+  background-repeat: no-repeat;
+  background-position: center;
+}
+
+.stars .stars--star:first-child {
+  margin-left: 0;
+}
```
jstmpelowycz, line 20: There is no need to increase the selector specificity

jstmpelowycz (changes requested):
One thing to fix

jstmpelowycz (approved):

PR #1271, opened 2022-05-07
blocks of stars and styles were added

LAndrikevych:
- [DEMO LINK](https://LAndrikevych.github.io/layout_stars/)
- [TEST REPORT LINK](https://LAndrikevych.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,34 @@
+body {
+  margin: 0;
+  box-sizing: border-box;
+}
+
+.stars {
+  display: flex;
+}
+
+.stars__star {
+  height: 16px;
+  width: 16px;
+  margin: 0 2px;
+  background-image: url(./images/star.svg);
+  background-position: center;
+  background-repeat: no-repeat;
+  background-clip: content-box;
+}
+
+.stars__star:first-child {
+  margin-left: 0;
+}
+
+.stars__star:last-child {
+  margin-right: 0;
+}
+
+.stars--1 .stars__star:nth-child(1),
+.stars--2 .stars__star:nth-child(-n+2),
+.stars--3 .stars__star:nth-child(-n+3),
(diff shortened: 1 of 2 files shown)


src/style.css
```diff
@@ -0,0 +1,34 @@
...
+}
+
+.stars__star {
+  height: 16px;
+  width: 16px;
+  margin: 0 2px;
+  background-image: url(./images/star.svg);
+  background-position: center;
+  background-repeat: no-repeat;
+  background-clip: content-box;
```
Olexandr6, line 17: Do you really need this` background-clip` here?

src/style.css
```diff
@@ -0,0 +1,34 @@
...
+}
+
+.stars {
+  display: flex;
+}
+
+.stars__star {
+  height: 16px;
+  width: 16px;
+  margin: 0 2px;
```
Olexandr6, line 13: Let's set margins only from one side, not from both

src/style.css
```diff
@@ -0,0 +1,34 @@
+body {
+  margin: 0;
+  box-sizing: border-box;
```
Olexandr6, line 3: Set border-box in `* {...}` selector

src/style.css
```diff
@@ -0,0 +1,34 @@
...
+  background-clip: content-box;
+}
+
+.stars__star:first-child {
+  margin-left: 0;
+}
+
+.stars__star:last-child {
+  margin-right: 0;
+}
```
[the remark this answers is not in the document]
LAndrikevych: using margin only on one side lets me do it with one selector, right?)

src/style.css
illiarerun: that's redudnant here
```suggestion
```

illiarerun (approved):

PR #1279, opened 2022-05-07
stars block was added with style css

LesyaSvk:
- [DEMO LINK](https://LesyaSvk.github.io/layout_stars/)
- [TEST REPORT LINK](https://LesyaSvk.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,25 @@
+body {
+  margin: 0;
+}
+
+.stars {
+  display: flex;
+}
+
+.stars__star {
+  background-image: url("images/star-active.svg");
+  background-repeat: no-repeat;
+  background-position: center;
+  width: 16px;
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,25 @@
...
+  display: flex;
+}
+
+.stars__star {
+  background-image: url("images/star-active.svg");
+  background-repeat: no-repeat;
+  background-position: center;
+  width: 16px;
+  height: 16px;
+  border-radius: 0.5px;
```
mykola-koval, line 15: do we really need this style?

mykola-koval (approved):

src/style.css
LesyaSvk: @mykola-koval do you mean only border-radius?

PR #1286, opened 2022-05-09
Develop

SaperZP:
[DEMO LINK](https://saperzp.github.io/layout_stars/)
[TEST REPORT LINK](https://saperzp.github.io/layout_stars/report/html_report/)


src/style.css
```diff
@@ -0,0 +1,25 @@
+body {
+  margin: 0;
+}
+
+.stars {
+  display: flex;
+  justify-content: space-between;
+  width: 96px;
```
illiarerun, line 8: [checklist](https://github.com/mate-academy/layout_stars/blob/master/checklist.md) 3rd item. don't hardcode width, let the content inside dictate it
```suggestion
```

src/style.css
```diff
@@ -0,0 +1,25 @@
...
+}
+
+.stars__star {
```
illiarerun, line 11: so just use margins to split your stars

illiarerun (changes requested):
GJ, now let's work on some fixes:

src/style.css
SaperZP: done

src/style.css
SaperZP: done

mykola-koval (approved):

PR #1291, opened 2022-05-10
star created

Ihor-Novakovskyi:
- [DEMO LINK](https://Ihor-Novakovskyi.github.io/layout_stars/)
- [TEST REPORT LINK](https://Ihor-Novakovskyi.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -16,7 +16,6 @@
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-
     <div class="stars--0 star">
       <div class="stars__star"></div>
       <div class="stars__star"></div>
@@ -64,6 +63,5 @@
       <div class="stars__star"></div>
       <div class="stars__star"></div>
     </div>
-
   </body>
 </html>


src/index.html
```diff
@@ -2,12 +2,68 @@
...
+        maximum-scale=1.0,
+        minimum-scale=1.0
+        "
+    >
     <title>Stars</title>
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Stars</h1>
+
```
jstmpelowycz, line 19: Don't leave any extra whitespaces if not needed

src/index.html
```diff
@@ -2,12 +2,68 @@
...
+    <div class="stars--0">
```
jstmpelowycz, line 20: It should look like
```suggestion
    <div class="stars stars--0">
```

src/style.css
```diff
@@ -0,0 +1,45 @@
...
+
+.stars--0,
+.stars--1,
+.stars--2,
+.stars--3,
+.stars--4,
+.stars--5 {
+  margin: 0;
+  display: flex;
+}
```
jstmpelowycz, line 45: There are a few issues to fix:
- you didn't use class `stars`, that's better to add a flexbox there but not to customize `start--...`
- duplicating styles when setting `background-image`, try to combine it as you did down below

As a reference, you can review solutions of your peers

jstmpelowycz (changes requested):
Almost done, just a few things to fix

jstmpelowycz (approved):
LGTM

PR #1295, opened 2022-05-10
stars block

natalia-maryshko:
https://natalia-maryshko.github.io/layout_stars/
https://natalia-maryshko.github.io/layout_stars/report/html_report/


vkryvytskyy (approved):
well done

PR #1344, opened 2022-06-05
Develop - Added Stars block

LordAsmodey:
- [DEMO LINK](https://LordAsmodey.github.io/layout_stars/)
- [TEST REPORT LINK](https://LordAsmodey.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -15,22 +15,10 @@
   background: url("./images/star.svg") center no-repeat;
 }
 
-.stars--1 .stars__star:nth-child(1) {
-  background: url("./images/star-active.svg") center no-repeat;
-}
-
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,24 @@
...
+  padding: 0;
+  box-sizing: border-box;
+}
+
+.stars {
+  display: flex;
+}
+
+.stars__star {
+  margin-right: 4px;
```
VETALEK, line 12: the rightmost star also has margin, but it is not supposed to 🥲

VETALEK (changes requested):
Good 🌟
Just a few fixes 🙃

VETALEK (approved):
Nice 👍

PR #1359, opened 2022-06-07
Created star block

denys-danyliuk:
- [DEMO LINK](https://denys-danyliuk.github.io/layout_stars/)
- [TEST REPORT LINK](https://denys-danyliuk.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,22 @@
+body {
+  margin: 0;
+}
+
+.stars {
+  display: flex;
+}
+
+.stars__star {
+  background: url(images/star.svg) center no-repeat;
+  margin-right: 4px;
+  width: 16px;
+  height: 16px;
+}
+
+.stars--1 :first-child,
+.stars--2 :nth-child(-n+2),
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,22 @@
+body {
+  margin: 0;
+}
```
VETALEK, line 3: better use a class, attached to `body` 😉

src/style.css
```diff
@@ -0,0 +1,22 @@
...
+
+.stars {
+  display: flex;
+}
+
+.stars__star {
+  background: url(images/star.svg) center no-repeat;
+  margin-right: 4px;
```
VETALEK, line 11: but what about the rightmost stars? 🤔😉

VETALEK (changes requested):
Cool 😎 
commented 🔥

src/style.css
denys-danyliuk: fixed!

solaryasha (approved):

PR #1360, opened 2022-06-07
30 stars

Kras4iy:
- [DEMO LINK](https://Kras4iy.github.io/layout_stars/)
- [TEST REPORT LINK](https://Kras4iy.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,4 +1,4 @@
-body {
+.main_body {
   margin: 0;
 }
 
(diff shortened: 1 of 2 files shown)


src/style.css
```diff
@@ -0,0 +1,24 @@
+body {
```
VETALEK, line 1: better use some class, attached to `body` 😉

VETALEK (changes requested):
Nice 👌 
commented 🙃

VETALEK (approved):
Like 🔥

PR #1361, opened 2022-06-07
Develop

OlgaStakhova:
- [DEMO LINK](https://OlgaStakhova.github.io/layout_stars/)
- [TEST REPORT LINK](https://OlgaStakhova.github.io/layout_stars/report/html_report/)
Test failed because stars have size bigger to 1px. I dont know how fixed it.

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,46 @@
+body {
+  margin: 0;
+  padding: 0;
+}
+
+.wraper {
+  width: 96px;
+  margin: 0 auto;
+}
+
+.star {
+  display: flex;
+  justify-content: space-between;
+  align-items: center;
+}
+
+.stars__star {
+  background-image: url("./images/star.svg");
+  border-radius: 0.5px;
+  width: 16px;
+  height: 16px;
+  display: block;
+  background-position: center;
+  background-repeat: no-repeat;
+  background-size: 13px 12px;
+}
+
+.stars--1 .stars__star:first-child {
+  background-image: url("./images/star-active.svg");
+}
+
+.stars--2 .stars__star:nth-child(-n+2) {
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,46 @@
...
+  margin: 0;
+  padding: 0;
+}
+
+.wraper {
+  width: 96px;
+  margin: 0 auto;
+}
+
+.star {
```
VETALEK, line 11:
```suggestion
.stars {
```

src/style.css
```diff
@@ -0,0 +1,46 @@
...
+  background-image: url("./images/star-active.svg");
+}
+
+.stars--4 .stars__star:nth-child(-n+4) {
+  background-image: url("./images/star-active.svg");
+}
+
+.stars--5 .stars__star:nth-child(-n+5) {
+  background-image: url("./images/star-active.svg");
+}
```
VETALEK, line 46: you may group these rules 😉

```
selector1,
selector2,
...
selectorN {
  rules;
}
```

VETALEK (changes requested):
Good 👍 
commented 😁

src/style.css
```diff
@@ -3,15 +3,12 @@ body {
...
-  width: 96px;
-  margin: 0 auto;
-}
-
-.star {
+.stars {
   display: flex;
   justify-content: space-between;
   align-items: center;
+  width: 96px;
```
VETALEK, line 10: don't hardcode width, cause you may need only 4 or 6 start 😉

VETALEK (changes requested):
Almost 💪

VETALEK (approved):
Good 👍🏽
try using `width: min-content` on `.stars`, so it won't be the whole screen wide 😉

PR #1363, opened 2022-06-07
stars solution

Yura-Lysenko:


src/style.css
```diff
@@ -0,0 +1,24 @@
...
+.stars {
+  display: flex;
+}
+
+.stars__star {
+  background-image: url(images/star.svg);
+  background-repeat: no-repeat;
+  height: 16px;
+  width: 16px;
+  margin-right: 4px;
```
VETALEK, line 14: the rightmost start has margin too, but it shouldn't 🥲

VETALEK (changes requested):
Well done 🤩 
commented 🙃

solaryasha (approved):
GJ!

PR #1368, opened 2022-06-08
Develop

LeonidPustovit:
- [DEMO LINK](https://LeonidPustovit.github.io/layout_stars/)
- [TEST REPORT LINK](https://LeonidPustovit.github.io/layout_stars/report/html_report/)
- Якісь криві тести, але наче говорили що така штука буває, через піксель перфектність.


src/style.css
```diff
@@ -0,0 +1,31 @@
+* {
+  box-sizing: border-box;
+}
+
+body {
+  margin: 0;
+}
+
+.stars {
+  width: 96px;
```
VETALEK, line 10: don't hardcode the width, what if you want only 4 or 6 stars? this will break the gap between them 🥲

src/style.css
```diff
@@ -0,0 +1,31 @@
...
+
+.stars__star {
+  background-image: url(./images/star-active.svg);
+  background-repeat: no-repeat;
+  background-position: center;
+  height: 16px;
+  width: 16px;
+}
+
+.stars--0 > .stars__star:nth-child(n+1),
```
VETALEK, line 25:
```suggestion
.stars--0 :nth-child(n+1),
```
you don't need so much specificity 😉

VETALEK (changes requested):
GJ 👍 
commented 😎

src/style.css
LeonidPustovit: But I don't have external container, and they will take all free space(screen width) if no width added.

src/style.css
```diff
@@ -0,0 +1,32 @@
+* {
+  box-sizing: border-box;
+}
+
+body {
+  margin: 0;
+}
+
+.stars {
+  width: 96px;
```
VETALEK, line 10: check out about [`min-content`](https://developer.mozilla.org/en-US/docs/Web/CSS/min-content)

VETALEK (changes requested):
Almost 💪

src/style.css
LeonidPustovit: Overall, I can add margins to create gaps between stars, but this looks like "bicycle", as soon as we can't use "gap" attribute in flex container (task restriction).

VETALEK: old engine on tests 🥲🥲🥲

VETALEK (approved):
Nice 🔥

PR #1377, opened 2022-06-08
Stars were created

yuliia-dronova:
- [DEMO LINK](https://yuliia-dronova.github.io/layout_stars/)
- [TEST REPORT LINK](https://yuliia-dronova.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -8,7 +8,6 @@
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1 hidden>Stars</h1>
 
     <div class="stars stars--0">
       <div class="stars_star"></div>
--- a/src/index.html
+++ b/src/index.html
@@ -8,6 +8,54 @@
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Stars</h1>
+    <h1 hidden>Stars</h1>
+
+    <div class="stars stars--0">
+      <div class="stars_star"></div>
+      <div class="stars_star"></div>
+      <div class="stars_star"></div>
+      <div class="stars_star"></div>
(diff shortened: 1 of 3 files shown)


src/index.html
```diff
@@ -8,6 +8,54 @@
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Stars</h1>
+    <h1 hidden>Stars</h1>
```
VETALEK, line 11:
```suggestion
```

you don't need it (you even hid it) 😉

src/style.css
```diff
@@ -0,0 +1,25 @@
...
+  display: flex;
+}
+
+.stars_star {
+  width: 16px;
+  height: 16px;
+  background-image: url(images/star.svg);
+  background-repeat: no-repeat;
+  background-position: center;
+  margin-right: 4px;
```
VETALEK, line 16: the rightmost star don't need a `marign` 🥲

image

VETALEK (changes requested):
Good start 👍

src/index.html
yuliia-dronova: Ok:)
But earlier Misha (as far as I remember) said that it is essential to keep structure as it is has to be. And unrelevant elements can be just hidden.. Is it true?)

VETALEK: you may hide element while developing, but you don't need element that user can't even see in production 😉

VETALEK: it is just useless 🙃

VETALEK (approved):
Fire 🔥

src/index.html
yuliia-dronova: Clear, thank you ^^)

PR #1382, opened 2022-06-08
Stars were created

Ihor-Myslovskyi:
- [DEMO LINK](https://ihor-myslovskyi.github.io/layout_stars/)
- [TEST REPORT LINK](https://ihor-myslovskyi.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,25 @@
+body {
+  margin: 0;
+}
+
+.stars {
+  width: 96px;
+  display: flex;
+  justify-content: space-between;
+}
+
+.stars__star {
+  width: 16px;
+  height: 16px;
+  background-image: url(./images/star.svg);
+  background-repeat: no-repeat;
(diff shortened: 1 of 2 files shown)


src/style.css
```diff
@@ -0,0 +1,25 @@
+body {
+  margin: 0;
+}
+
+.stars {
+  width: 96px;
```
VETALEK, line 6: don't hardcode the `width`, what if you want only 4 or 6 stars 😉
batter use `margin`/`gap` to set space between starts

VETALEK (changes requested):
👍

src/style.css
Ihor-Myslovskyi: I got the point! It even helped to pass all tests! Thanks :)

VETALEK (approved):
Fire 🔥

PR #1396, opened 2022-06-09
Added stars

LovelyDrug:
[DEMO LINK](https://LovelyDrug.github.io/layout_stars/)
[TEST REPORT LINK](https://LovelyDrug.github.io/layout_stars/report/html_report/)


src/style.css
```diff
@@ -0,0 +1,29 @@
+body {
+  margin: 0;
+  box-sizing: border-box;
```
VETALEK, line 3: you don't use neither `padding` no `border`, so you don't need `border-box` 😉

src/style.css
```diff
@@ -0,0 +1,29 @@
...
+}
+
+.stars {
+  display: flex;
+}
+
+.stars__star {
+  margin-right: 4px;
```
VETALEK, line 11: don't forget about the last star in each row 😉

image

VETALEK (changes requested):
Good start 👍

VETALEK (approved):
🔥

PR #1431, opened 2022-06-26
Star box was added

Sergey-Mironenko:


src/style.css
```diff
@@ -0,0 +1,84 @@
...
+.stars--1 .stars__star:nth-child(n + 2) {
+  background-image: url(images/star.svg);
+  background-repeat: no-repeat;
+  background-position: center;
+}
+    
+.stars--2 .stars__star:nth-child(-n + 2) {
+  background-image: url(images/star-active.svg);
+  background-repeat: no-repeat;
+  background-position: center;
```
Bohdan-Maksymiuk, line 53: so much repeated code

Bohdan-Maksymiuk: you could set default star for bg-image
and then change (+ combine selectors)
```
}
```

Bohdan-Maksymiuk (changes requested):
Screenshot 2022-06-29 at 14 46 54

next time it should be links here

src/style.css
```diff
@@ -0,0 +1,35 @@
+body {
+  margin: 0;
+  display: flex;
+}
+   
+.stars_block {
+  display: flex;
+  flex-direction: column;
+  height: 96px;
+  width: 96px;
```
Bohdan-Maksymiuk, line 10: do not hardcode - it is unnecessary here
just set `margin-right: 4px;` for `.stars__star`

src/style.css
```diff
@@ -0,0 +1,35 @@
...
+  display: flex;
+  flex-direction: column;
+  height: 96px;
+  width: 96px;
+  min-width: 96px;
+}
+   
+.stars {
+  display: flex;
+  flex-direction: row;
```
Bohdan-Maksymiuk, line 16: default - remove

src/style.css
```diff
@@ -0,0 +1,35 @@
...
+  justify-content: space-between;
```
Bohdan-Maksymiuk, line 17: good, but you don't need responsiveness here, margin-right will be enough

Bohdan-Maksymiuk (changes requested):

src/style.css
```diff
@@ -0,0 +1,32 @@
+body {
+  margin: 0;
+  display: flex;
+}
+   
+.stars_block {
+  display: flex;
+  flex-direction: column;
+  min-width: 96px;
```
alexandra-protyanova, line 9: remove and nothing will change

alexandra-protyanova (approved):

PR #1446, opened 2022-07-04
commit

OleksChernikov:
https://<olekschernikov>.github.io/layout_stars/

https://<olekschernikov>.github.io/layout_stars/report/html_report/


src/style.css
```diff
@@ -0,0 +1,30 @@
+body {
+  margin: 0;
+}
+
+.stars--0 {
+  display: flex;
+  flex-direction: row;
```
vkryvytskyy, line 7: it's default `flex-direction` value, could be omitted

src/style.css
```diff
@@ -0,0 +1,30 @@
...
+}
+
+.stars__star {
+  background-image: url(./images/star.svg);
+  background-repeat: no-repeat;
+  width: 16px;
+  height: 16px;
+  display: flex;
+  flex-direction: row;
```
vkryvytskyy, line 16: remove this

vkryvytskyy (approved):

ArturErmolenko (approved):

PR #1456, opened 2022-07-09
Develop

devWizHub:
- [DEMO LINK](https://devWizHub.github.io/layout_stars/)
- [TEST REPORT LINK](https://devWizHub.github.io/layout_stars/report/html_report/)


src/style.css
```diff
@@ -0,0 +1,24 @@
...
+  margin: 0;
+  padding: 0;
+}
+
+.box {
+  display: flex;
+}
+
+.stars {
+  background-image: url(images/star.svg);
```
meforoma, line 11: don't forget to define 
```
background-repeat: ...;
background-position: ...;
```

src/style.css
```diff
@@ -0,0 +1,24 @@
...
+  color: transparent;
+  height: 12px;
+  width: 14px;
+  margin: 0.2em;
```
meforoma, line 15: better define a one-direction margin;
and i guess the width and height could be 16px each
```suggestion
  height: 16px;
  width: 16px;
  margin-right: 4px;
```

and remove the margin-right with the `last-child` selector - to not have that last margin stick-out

devWizHub: understood. thanks

src/style.css
devWizHub: understood. thanks

amitpatiljc (approved):
Good Job!!

meforoma (approved):

PR #1458, opened 2022-07-10
done

savchuksavchuk:
[DEMO LINK](https://savchuksavchuk.github.io/layout_stars/)
[TEST REPORT LINK](https://savchuksavchuk.github.io/layout_stars/report/html_report/)


src/style.css
```diff
@@ -0,0 +1,51 @@
...
+  background: url(images/star-active.svg);
+  background-repeat: no-repeat;
+  background-position: center;
+}
+
+.stars--5 .stars__star:nth-child(-n+5) {
+  background: url(images/star-active.svg);
+  background-repeat: no-repeat;
+  background-position: center;
+}
```
alexandra-protyanova, line 51: your code is repeating in these lines, think how u can combine them together

alexandra-protyanova (changes requested):

src/style.css
savchuksavchuk: Hi, 
I've added the variables for bg.

PR #1470, opened 2022-07-12
created blocks with stars, added style, wrote down user name

oksana-logos-frontend:
https://oksana-logos-frontend.github.io/layout_stars/
https://oksana-logos-frontend.github.io/layout_stars/report/html_report/


danylo-onishchuk (approved):
good job!

PR #1540, opened 2022-07-17
added

SerhiiZiatiuk:
- [DEMO LINK](https://SerhiiZiatiuk.github.io/layout_stars/)
- [TEST REPORT LINK](https://SerhiiZiatiuk.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,26 @@
+* {
+  margin: 0;
+  box-sizing: border-box;
+}
+
+.stars {
+  display: flex;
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,26 @@
+* {
+  margin: 0;
```
AlinaMatuschak, line 2: Don't use * selector for zeroing out your margins or paddings

AlinaMatuschak (approved):

PR #1549, opened 2022-07-20
first commit

olena-hapon:
https://olena-hapon.github.io/layout_stars/
https://olena-hapon.github.io/layout_stars/report/html_report/


ArturErmolenko (approved):
Well done 👍

PR #1578, opened 2022-07-29
stars added

bradleymatjie:


src/style.css
```diff
@@ -0,0 +1,61 @@
...
+  height: 16px;
+  margin-right: 4px;
+}
+
+.stars--1 :nth-child(-n + 1) {
+  background-image: url("./images/star-active.svg");
+  background-repeat: no-repeat;
+  background-position: center;
+  width: 16px;
+  height: 16px;
```
VinnieJ-2k20, line 28: These rules don't change, no need to overwrite them.

src/style.css
```diff
@@ -0,0 +1,61 @@
...
+}
+
+.stars--3 :nth-child(-n + 3) {
+  background-image: url("./images/star-active.svg");
+  background-repeat: no-repeat;
+  background-position: center;
+  width: 16px;
+  height: 16px;
+}
```
VinnieJ-2k20, line 45: Don't repeat yourself. Combine selectors for same styles.

VinnieJ-2k20 (changes requested):
Solution works, but code can be improved.

VinnieJ-2k20 (approved):
Good job 👍

PR #1585, opened 2022-07-31
added stars block

Sasha-Green:
https://Sasha-Green.github.io/layout_stars/
https://Sasha-Green.github.io/layout_stars/report/html_report/


ArturErmolenko (approved):
well done 👍

PR #1586, opened 2022-07-31
stars where added

AlexKinda:
https://AlexKinda.github.io/layout_stars/
https://AlexKinda.github.io/layout_stars/report/html_report/


ArturErmolenko (approved):
Well done 👍

PR #1597, opened 2022-08-05
First try

Proph7000:
https://proph7000.github.io/layout_stars/
https://proph7000.github.io/layout_stars/report/html_report/


alexandra-protyanova (approved):
well done!

PR #1608, opened 2022-08-09
Stars

MaxManis:
- [DEMO LINK](https://MaxManis.github.io/layout_stars/)
- [TEST REPORT LINK](https://MaxManis.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -8,6 +8,47 @@
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Stars</h1>
+    <ul class="stars--0">
+      <li class="stars__star"></li>
+      <li class="stars__star"></li>
+      <li class="stars__star"></li>
+      <li class="stars__star"></li>
+      <li class="stars__star"></li>
+    </ul>
+    <ul class="stars--1">
+      <li class="stars__star"></li>
+      <li class="stars__star"></li>
+      <li class="stars__star"></li>
+      <li class="stars__star"></li>
+      <li class="stars__star"></li>
+    </ul>
+    <ul class="stars--2">
+      <li class="stars__star"></li>
+      <li class="stars__star"></li>
+      <li class="stars__star"></li>
+      <li class="stars__star"></li>
+      <li class="stars__star"></li>
+    </ul>
+    <ul class="stars--3">
+      <li class="stars__star"></li>
+      <li class="stars__star"></li>
+      <li class="stars__star"></li>
+      <li class="stars__star"></li>
+      <li class="stars__star"></li>
+    </ul>
+    <ul class="stars--4">
+      <li class="stars__star"></li>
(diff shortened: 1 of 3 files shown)


src/index.html
```diff
@@ -8,6 +8,47 @@
...
   </head>
   <body>
-    <h1>Stars</h1>
+    <ul class="stars--0">
+      <li class="stars__star"></li>
+      <li class="stars__star"></li>
+      <li class="stars__star"></li>
+      <li class="stars__star"></li>
+      <li class="stars__star"></li>
+    </ul>
```
OleksiiBlezniuk, line 17: Well, div is enough for this purpose 🙂

src/style.css
```diff
@@ -0,0 +1,47 @@
+body {
+  margin: 0;
+  padding: 0;
+}
+
+ul {
+  list-style: none;
+  padding-left: 0;
+  margin: 0;
+}
```
OleksiiBlezniuk, line 10: That's what i meant, you could avoid this using div

src/style.css
```diff
@@ -0,0 +1,47 @@
...
+  background: url("images/star-active.svg") no-repeat center;
+}
+
+.stars--4 > .stars__star:not(:nth-child(1n+5)) {
+  background: url("images/star-active.svg") no-repeat center;
+}
+
+.stars--5 > .stars__star:not(:nth-child(1n+6)) {
+  background: url("images/star-active.svg") no-repeat center;
+}
```
OleksiiBlezniuk, line 47: This styles are same, combine them to one rule

OleksiiBlezniuk (changes requested):

src/style.css
```diff
@@ -0,0 +1,29 @@
...
+
+.stars--0,
+.stars--1,
+.stars--2,
+.stars--3,
+.stars--4,
+.stars--5 {
+  display: flex;
+}
```
OleksiiBlezniuk, line 21: Strange solution
This `--n` part - is a modifier that define how many stars should be active in this stars list

Instead of applying one rule for 5 classes better to create one class for 5 elements 🙂

src/style.css
```diff
@@ -0,0 +1,29 @@
...
+
+.stars--1 > .stars__star:not(:nth-child(1n+2)),
```
OleksiiBlezniuk, line 23:
```suggestion
.stars--1 > .stars__star:nth-child(-n + 1),
```

OleksiiBlezniuk: Don't make it more complicated that it should be 🙂

OleksiiBlezniuk (approved):

OleksiiBlezniuk (changes requested):
sorry, missclick, please fix comments 🙂

OleksiiBlezniuk (approved):
👍

PR #1622, opened 2022-08-12
start created

dmitriyn3679:
- [DEMO LINK](https://dmitriyn3679.github.io/layout_stars/)
- [TEST REPORT LINK](https://dmitriyn3679.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,40 @@
+body {
+  margin: 0;
+}
+
+.stars {
+  display: flex;
+}
+
+.stars__star {
+  width: 16px;
+  height: 16px;
+  background-image: url("images/star.svg");
+  background-repeat: no-repeat;
+  background-position: center;
+  margin-right: 4px;
+}
+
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,40 @@
...
+  background-image: url("images/star-active.svg");
+}
+
+.stars--4 > .stars__star:nth-of-type(-n+4) {
+  background-image: url("images/star-active.svg");
+}
+
+.stars--5 > .stars__star:nth-of-type(n) {
+  background-image: url("images/star-active.svg");
+}
```
alexandra-protyanova, line 40: let us combine all these selectors

alexandra-protyanova (changes requested):

src/style.css
dmitriyn3679: > let us combine all these selectors

combined

valerii-derkach (approved):

PR #1624, opened 2022-08-12
solution

Quelaan1:


src/index.html
```diff
@@ -8,6 +8,49 @@
...
-    <h1>Stars</h1>
+    <div class="star">
+      <div class="stars stars--0">
+        <div class="stars__star"></div>
+        <div class="stars__star"></div>
+        <div class="stars__star"></div>
+        <div class="stars__star"></div>
+        <div class="stars__star"></div>
+      </div>
+      <div class="stars stars--1">
```
VinnieJ-2k20, line 19:
```suggestion
      </div>
      
      <div class="stars stars--1">
```

src/style.css
```diff
@@ -0,0 +1,53 @@
+* {
+  margin: 0;
+}
```
VinnieJ-2k20, line 3: Don't add `margin` rule under `*` selector

src/style.css
```diff
@@ -0,0 +1,53 @@
...
+
+.star {
+  width: 100px;
+  height: 96px;
+}
```
VinnieJ-2k20, line 8: Are you sure you need total container size hardcoded since you provide size of individual element later?

src/index.html
```diff
@@ -8,6 +8,49 @@
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Stars</h1>
+    <div class="star">
```
VinnieJ-2k20, line 11: Not the best class name. It's not a star.

src/style.css
```diff
@@ -0,0 +1,53 @@
...
+  display: flex;
+  background-repeat: no-repeat;
+  background-image: url("images/star-active.svg");
+}
+
+.stars--2 > :nth-child(n+0):nth-last-child(n+4) {
+  display: flex;
+  background-repeat: no-repeat;
+  background-image: url("images/star-active.svg");
+}
```
VinnieJ-2k20, line 35: 1. Don't repeat yourself, combine selectors for same rules.
2. Are you sure you need display flex here?
3. Background-repeat rule is the same in stars__star. You don't need to overwrite it.

VinnieJ-2k20 (changes requested):
Solution works, but code can be improved.

VinnieJ-2k20 (approved):
Good job 👍

PR #1626, opened 2022-08-12
Solution

nikitaovch99:
https://nikitaovch99.github.io/layout_stars/
https://nikitaovch99.github.io/layout_stars/report/html_report/


src/style.css
```diff
@@ -0,0 +1,32 @@
...
+}
+
+.stars--0,
+.stars--1,
+.stars--2,
+.stars--3,
+.stars--4,
+.stars--5 {
+  display: flex;
+}
```
Denchepornyuk, line 12: add one class for all these divs and use this class instead of this big selector

Denchepornyuk (changes requested):

Denchepornyuk (approved):

PR #1633, opened 2022-08-15
The stars are ready

denisdudin50:
https://denisdudin50.github.io/layout_stars/
https://denisdudin50.github.io/layout_stars/report/html_report/

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,96 @@
+* {
+  margin: 0;
+  padding: 0;
+}
+
+.stars {
+  display: flex;
+}
+
+.stars__star {
+  margin-right: 4px;
+}
+
+.stars__star:last-child {
+  margin-right: 0;
+}
+
+.stars--0 .stars__star:nth-child(n) {
+  width: 16px;
+  height: 16px;
+  background-image: url("images/star.svg");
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,36 @@
+* {
+  margin: 0;
+  padding: 0;
+}
```
OleksiiBlezniuk, line 4: Don't reset styles using * selector

src/style.css
```diff
@@ -0,0 +1,36 @@
...
+  background-image: url("images/star.svg");
+}
+
+.stars--1 .stars__star:first-child,
+.stars--2 .stars__star:nth-child(-n + 2),
+.stars--3 .stars__star:nth-child(-n + 3),
+.stars--4 .stars__star:nth-child(-n + 4),
+.stars--5 .stars__star:nth-child(n) {
+  background-image: url("images/star-active.svg");
+}
```
OleksiiBlezniuk, line 36: You can just set star image to `stars__star` class, star-active will override it

OleksiiBlezniuk (changes requested):

OleksiiBlezniuk (approved):
Good job!

PR #1634, opened 2022-08-15
stars block was created

artem-nasadchuk:
- [DEMO LINK](https://artem-nasadchuk.github.io/layout_stars/)
- [TEST REPORT LINK](https://artem-nasadchuk.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,26 @@
+body {
+  margin: 0;
+  padding: 0;
+}
+
+.stars {
+  display: flex;
+}
+
+.stars__star {
+  width: 16px;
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,26 @@
+body {
+  margin: 0;
+  padding: 0;
+}
```
Denchepornyuk, line 4: body doesn't have default padding, remove it

Denchepornyuk (changes requested):

Denchepornyuk (approved):

PR #1646, opened 2022-08-16
Created stars_block

Vladosik1337:
- [DEMO LINK](https://Vladosik1337.github.io/layout_stars/)
- [TEST REPORT LINK](https://Vladosik1337.github.io/layout_stars/report/html_report/)


src/style.css
```diff
@@ -0,0 +1,28 @@
+* {
+  margin: 0;
+}
```
Denchepornyuk, line 3: don't reset styles in * tag

Denchepornyuk (changes requested):

src/style.css
Denchepornyuk, line 3: remove this css rule

Denchepornyuk (approved):

PR #1654, opened 2022-08-16
solved task

myroslav-diiak:
- [DEMO LINK](https://myroslav-diiak.github.io/layout_stars/)
- [TEST REPORT LINK](https://myroslav-diiak.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -7,7 +7,58 @@
     <title>Stars</title>
     <link rel="stylesheet" href="./style.css">
   </head>
-  <body>
-    <h1>Stars</h1>
+  <body class="body">
+
+    <div class="main-container">
+      
+      <div class="container stars--0">
+        <div class="star"></div>
+        <div class="star"></div>
+        <div class="star"></div>
+        <div class="star"></div>
+        <div class="star"></div>
+      </div>
+
+      <div class="container stars--1">
+        <div class="star"></div>
+        <div class="star"></div>
+        <div class="star"></div>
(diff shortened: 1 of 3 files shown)


src/index.html
```diff
@@ -7,7 +7,58 @@
...
-  <body>
-    <h1>Stars</h1>
+  <body class="body">
+
+    <div class="main-container">
+      
+      <div class="container stars--0">
+        <div class="star"></div>
+        <div class="star"></div>
+        <div class="star"></div>
```
OleksiiBlezniuk, line 17: It's a wrong class names due to BEM methodology
There is a block (`stars`) and it has elements (`stars__star`)
It's okay now, but when you will reuse this block, you'll have to rename classes

src/style.css
```diff
@@ -0,0 +1,25 @@
...
+
+.container {
+  display: flex;
+}
+
+.star {
+  height: 16px;
+  width: 16px;
+  background-image: url(images/star.svg);
+  background-size: 13px;
```
OleksiiBlezniuk, line 13: This rule changes nothing here, I believe it can be removed

src/style.css
```diff
@@ -0,0 +1,25 @@
...
+  margin-right: 4px;
```
OleksiiBlezniuk, line 14: Would be great to disable margins for last element

OleksiiBlezniuk (approved):

PR #1658, opened 2022-08-16
stars was created

iPanda177:
- [DEMO LINK](https://iPanda177.github.io/layout_stars/)
- [TEST REPORT LINK](https://iPanda177.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,24 @@
+body {
+  margin: 0;
+}
+
+.stars {
+  display: flex;
+}
+
+.stars__star {
+  background-image: url(./images/star.svg);
+  background-repeat: no-repeat;
+  background-position: center;
+  width: 16px;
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,24 @@
...
+  display: flex;
+}
+
+.stars__star {
+  background-image: url(./images/star.svg);
+  background-repeat: no-repeat;
+  background-position: center;
+  width: 16px;
+  height: 16px;
+  margin-right: 4px;
```
OleksiiBlezniuk, line 15: Would be nice to disable margins for last element

OleksiiBlezniuk (approved):

src/style.css
iPanda177: ty so much)

PR #1660, opened 2022-08-16
layout_stars task

Pallaci0:
DEMO LINK  https://Pallaci0.github.io/layout_stars/
TEST REPORT LINK  https://Pallaci0.github.io/layout_stars/report/html_report/

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,36 @@
+
+.stars {
+  display: flex;
+}
+
+.stars__star {
+  background-image: url(images/star-active.svg);
+  background-repeat: no-repeat;
+  height: 16px;
+  width: 16px;
+  margin-right: 4px;
+}
+
+.stars--0 .stars__star:nth-child(n+1) {
+  background-image: url(images/star.svg);
+}
+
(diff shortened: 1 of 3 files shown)


ArturErmolenko (changes requested):
as you could see test didn't pass, it should not be like that, the tests should be passed

src/style.css
```diff
@@ -0,0 +1,44 @@
...
+.stars--3 .stars__star:nth-child(n+4) {
+  background-image: url(images/star.svg);
+}
+
+.stars--4 .stars__star:nth-child(n+5) {
+  background-image: url(images/star.svg);
+}
+
+.stars--5.stars__star:nth-child(n+6) {
+  background-image: url(images/star.svg);
```
alexandra-protyanova, line 43: combine all these selectors

alexandra-protyanova (changes requested):

src/style.css
Pallaci0: Done

alexandra-protyanova (approved):
good job!

PR #1691, opened 2022-08-17
Stars block was created

SanyaKovalenko:
[DEMO LINK](https://SanyaKovalenko.github.io/layout_stars/)
[TEST REPORT LINK](https://SanyaKovalenko.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,47 @@
+.body {
+  margin: 0;
+  padding: 0;
+}
+
+.stars {
+  margin: 0;
+  padding: 0;
+
+  list-style: none;
+
+  display: flex;
+
+}
+
+.stars__star {
+  width: 16px;
+  height: 16px;
+
+  background-image: url(./images/star.svg);
+  background-repeat: no-repeat;
+  background-position: center;
+}
+
+.stars__star:not(:last-child) {
+  margin-right: 4px;
+}
+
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,47 @@
+.body {
+  margin: 0;
+  padding: 0;
+}
```
Denchepornyuk, line 4: body doesn't have default padding, you can remove it

src/style.css
```diff
@@ -0,0 +1,47 @@
...
+
+.stars {
+  margin: 0;
+  padding: 0;
```
Denchepornyuk, line 8: div also doesn't have default padding and margin you can remove it

src/style.css
```diff
@@ -0,0 +1,47 @@
...
+  background-image: url(./images/star-active.svg);
+}
+
+.stars--4 .stars__star:nth-child(-n+4) {
+  background-image: url(./images/star-active.svg);
+}
+
+.stars--5 .stars__star:nth-child(-n+5) {
+  background-image: url(./images/star-active.svg);
+}
```
Denchepornyuk, line 47: you can combine all these selectors into one

Denchepornyuk (changes requested):

Denchepornyuk (approved):

PR #1697, opened 2022-08-17
added stars

VikaMartin22:
- [DEMO LINK](https://VikaMartin22.github.io/layout_stars/)
- [TEST REPORT LINK](https://VikaMartin22.github.io/layout_stars/report/html_report/)


src/style.css
```diff
@@ -0,0 +1,24 @@
...
+  display: flex;
+
+}
+
+.stars__star {
+  width: 16px;
+  height: 16px;
+  background-repeat: no-repeat;
+  background-image: url(./images/star.svg);
+  margin-right: 4px;
```
Denchepornyuk, line 15: let's add only the right margin and remove margin from first start in row then tests will pass

Denchepornyuk (changes requested):

src/style.css
```diff
@@ -0,0 +1,23 @@
+* {
+  margin: 0;
+}
```
Denchepornyuk, line 3: don't reset styles in `*` tag

Denchepornyuk (approved):

PR #1711, opened 2022-08-18
Develop

igor-1987:
- [DEMO LINK](https://igor-1987.github.io/layout_stars/)
- [TEST REPORT LINK](https://igor-1987.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -2,7 +2,7 @@
   --bg-active: url("./images/star-active.svg") center no-repeat;
   --bg-inactive: url("./images/star.svg") center no-repeat;
   --size: 16px;
-  --star-margin: 4px;
+  --star-distance: 4px;
 }
 
 .body {
@@ -17,7 +17,7 @@
   height: var(--size);
   width: var(--size);
   background: var(--bg-inactive);
-  margin-right: var(--star-margin);
+  margin-right: var(--star-distance);
 }
 
 .inactive {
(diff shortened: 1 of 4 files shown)


src/style.css
```diff
@@ -0,0 +1,37 @@
...
+  margin-right: var(--star-distance);
+}
+
+.inactive {
+  background: var(--bg-inactive);
+}
+
+.active {
+  background: var(--bg-active);
+}
```
OleksiiBlezniuk, line 29: Remove this classes, you didn't use them

src/style.css
```diff
@@ -0,0 +1,37 @@
...
+
+.stars--1 > .stars__star:not(:nth-child(n + 2)),
+.stars--2 > .stars__star:not(:nth-child(n + 3)),
```
OleksiiBlezniuk, line 32:
```suggestion
.stars--2 > .stars__star:nth-child(-n + 2),
```

OleksiiBlezniuk: You can make it simpler 🙂

OleksiiBlezniuk (changes requested):
Don't forget to deploy your homework

src/style.css
igor-1987: Thank you for review! I've made it according to the second remark/ and then I've tried to replace initial active and inactive state of stars. Solution is pretty much the same, but with all positive digits.

OleksiiBlezniuk (approved):

src/style.css
OleksiiBlezniuk: Well, solution is same, but now it has opposite logic 🙂 But ok

PR #1712, opened 2022-08-18
Develop

bokorzita:
https://bokorzita.github.io/layout_stars/
https://bokorzita.github.io/layout_stars/report/html_report/


VinnieJ-2k20 (changes requested):
Code looks good, but demo link is not working.

PR #1721, opened 2022-08-18
stars have been created

hajnaloltyan:
- [DEMO LINK](https://hajnaloltyan.github.io/layout_stars/)
- [TEST REPORT LINK](https://hajnaloltyan.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,27 @@
+body {
+  margin: 0;
+}
+
+.stars {
+  display: flex;
+}
+
+.stars__star {
+  background-image: url(/images/star.svg);
+  height: 16px;
+  width: 16px;
+
+  background-repeat: no-repeat;
+  margin-right: 4px;
+  background-position: 50%;
+}
+
+.stars--1 .stars__star:nth-child(-n+1),
+.stars--2 .stars__star:nth-child(-n+2),
+.stars--3 .stars__star:nth-child(-n+3),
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,27 @@
...
+}
+
+.stars__star {
+  background-image: url(/images/star.svg);
+  height: 16px;
+  width: 16px;
+
+  background-repeat: no-repeat;
+  margin-right: 4px;
+  background-position: 50%;
```
VinnieJ-2k20, line 16: Better to just use `center` here.

src/style.css
```diff
@@ -0,0 +1,27 @@
...
+}
+
+.stars--1 .stars__star:nth-child(-n+1),
+.stars--2 .stars__star:nth-child(-n+2),
+.stars--3 .stars__star:nth-child(-n+3),
+.stars--4 .stars__star:nth-child(-n+4),
+.stars--5 .stars__star:nth-child(-n+5) {
+  background-image: url(/images/star-active.svg);
+  background-repeat: no-repeat;
+  background-position: 50%;
```
VinnieJ-2k20, line 26: These rules don't change, so there's no need to override them.

VinnieJ-2k20 (changes requested):
Looks good, just couple of improvements left.

PR #1723, opened 2022-08-19
Develop2

La-La-Op:


src/style.css
```diff
@@ -0,0 +1,37 @@
...
+  background-image: url(images/star-active.svg);
+}
+
+.stars--4 :nth-last-child(n + 2) {
+  background-image: url(images/star-active.svg);
+}
+
+.stars--5 :nth-last-child(n + 1) {
+  background-image: url(images/star-active.svg);
+}
```
OleksiiBlezniuk, line 37: This properties are duplicated, combine them into one rule

OleksiiBlezniuk (changes requested):

src/style.css
La-La-Op: done

OleksiiBlezniuk (approved):

PR #1727, opened 2022-08-19
Initial commit

Bibicutu:


src/style.css
```diff
@@ -0,0 +1,33 @@
...
+  display: flex;
+
+  background-image: url("./images/star-active.svg");
+  background-size: 13px 13px;
+
+  height: 13px;
+  width: 13px;
+
+  margin: 3px 7px 3px 3px;
+
```
VinnieJ-2k20, line 23: Use correct star container size from design (16x16) and set correct background parameters.

src/style.css
```diff
@@ -0,0 +1,33 @@
...
+  border-radius: 0.5px;
```
VinnieJ-2k20, line 24: Not sure we need border-radius in this task.

VinnieJ-2k20 (changes requested):
Couple of improvements needed before approve.
Also, add links to pull request description.

ihor-jpeg (approved):

PR #1733, opened 2022-08-20
six stars blocks was added, but not passed autotest

ctacrh:
- [DEMO LINK](https://ctacrh.github.io/layout_stars/)
- [TEST REPORT LINK](https://ctacrh.github.io/layout_stars/report/html_report/)


src/style.css
```diff
@@ -0,0 +1,27 @@
...
+  margin-right: 4px;
+}
+
+.stars__star {
+  width: 16px;
+  height: 16px;
+  background: url(/images/star.svg) no-repeat center;
+}
+
+.stars--1 > :nth-child(-n+1),
```
liliia-livandovska, line 21: You can also use this option **.stars--1 .stars__star:nth-child(-n+1)**

src/style.css
```diff
@@ -0,0 +1,27 @@
+body {
+  margin: 0;
+}
+
+.stars {
+  display: flex;
+  justify-content: space-between;
+  width: 100px;
```
liliia-livandovska, line 8: let's move it from there to  stars__star. You would need to add a width and height of 16px; and also a margin according to Figma
Знімок екрана 2022-08-22 о 16 25 35

liliia-livandovskaya (changes requested):
GJ! Pay attention to comments

src/style.css
ctacrh: > let's move it from there to stars__star. You would need to add a width and height of 16px; and also a margin according to Figma Знімок екрана 2022-08-22 о 16 25 35

I can't understand why to use flexbox without taking advantage to it? Why use margin for this? And i was added a width and height of 16px

src/style.css
ctacrh: > You can also use this option **.stars--1 .stars__star:nth-child(-n+1)**

ok, i will use it

stanislav-topikha (approved):

PR #1735, opened 2022-08-21
star block is designed and all styles are identical to mock

Soroushsp:
https://soroushsp.github.io/layout_stars/
https://soroushsp.github.io/layout_stars/report/html_report/


ihor-jpeg (changes requested):
let's make the tests pass
pay attention to the reference

Screenshot 2022-08-23 at 17 11 19
Screenshot 2022-08-23 at 17 11 09

src/style.css
```diff
@@ -0,0 +1,71 @@
...
+  background-image: url("./images/star-active.svg");
+}
+
+.stars--0 {
+  margin-top: 32px;
+}
+
+.stars--5 {
+  margin-bottom: 32px;
+}
```
ihor-jpeg, line 71: group these

src/style.css
```diff
@@ -0,0 +1,71 @@
...
+  background-repeat: no-repeat;
+}
+
+.stars__star:last-child {
+  margin-right: 0;
+}
+
+.stars--0 > :nth-child(n) {
+  background-image: url("./images/star.svg");
+}
```
ihor-jpeg, line 27: you have multiple selectors that use the same styles. group them
```suggestion
.stars--0 > :nth-child(n),
.stars--1 > :nth-child(n+2),
.stars--2 > :nth-child(n+3) {
  background-image: url("./images/star.svg");
}
```

src/style.css
```diff
@@ -0,0 +1,71 @@
...
+
+.stars--1 > :first-child {
+  background-image: url("./images/star-active.svg");
+}
```
ihor-jpeg, line 31: group

ihor-jpeg (changes requested):

src/style.css
Soroushsp: sir I tried that but it didn't work 
because they are not the same styling.

src/style.css
Soroushsp: nice ... thank you for this advice. 
applied !!!

src/style.css
Soroushsp: nice ... thank you for this advice. 
applied !!!

Soroushsp:
dear Ihor, now you see all selectors which creates colored stars are grouped. 
the same for the selectors which create stars without color, they are also grouped. 
one pseudo selector had tag inside it and as styling with tags are not accepted I modified that. 

everything's working perfectly now.

src/style.css
```diff
@@ -0,0 +1,47 @@
+.container {
+  height: 160px;
+  width: 140px;
+  margin: 0;
+}
+
+.stars {
+  display: flex;
+  margin-left: 22px;
+  margin-right: 22px;
```
ihor-jpeg, line 10: margin: 0 22px;

ihor-jpeg (approved):

PR #1736, opened 2022-08-21
created a stars block

MeshackDev:
- [DEMO LINK](https://MeshackDev.github.io/layout_stars/)
- [TEST REPORT LINK](https://MeshackDev.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,198 @@
+body {
+  margin: 0;
+}
+
+.container {
+  width: 96px;
+  height: 96px;
+}
+
+.stars {
+  display: flex;
+  justify-content: space-between;
+}
+
+.stars__star {
+  height: 16px;
+  width: 16px;
+}
+
+.star--0 :nth-child(1) {
+  background-image: url(images/star.svg);
+  background-repeat: no-repeat;
+  background-position: center;
+}
+
+.star--0 :nth-child(2) {
+  background-image: url(images/star.svg);
+  background-repeat: no-repeat;
+  background-position: center;
+}
+
+.star--0 :nth-child(3) {
+  background-image: url(images/star.svg);
+  background-repeat: no-repeat;
+  background-position: center;
+}
+
+.star--0 :nth-child(4) {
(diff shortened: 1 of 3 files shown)


ihor-jpeg (changes requested):
please, make sure that the test results are available

redeploy, please

src/style.css
```diff
@@ -0,0 +1,197 @@
...
+  background-image: url(images/star-active.svg);
+  background-repeat: no-repeat;
+  background-position: center;
+}
+
+.star--5 :nth-child(4) {
+  background-image: url(images/star-active.svg);
+  background-repeat: no-repeat;
+  background-position: center;
+}
```
ihor-jpeg, line 191: maybe there is a shorter way

the styles are identical. use commas

ihor-jpeg (changes requested):

MeshackDev:
I've deployed it again, not sure why it does not appear but the DEMO LINK appears

src/style.css
MeshackDev: CSS changes made to make it shorter

src/style.css
```diff
@@ -0,0 +1,80 @@
...
+.star--4 :nth-child(5) {
+  background-image: url(images/star.svg);
+}
+
+.star--5 :nth-child(1),
+.star--5 :nth-child(2),
+.star--5 :nth-child(3),
+.star--5 :nth-child(4),
+.star--5 :nth-child(5) {
+  background-image: url(images/star-active.svg);
```
ihor-jpeg, line 79: come on, these are the same too

ihor-jpeg (changes requested):
user `nth-child(n)`

src/style.css
MeshackDev: Shortened css even more with nth-child(n)

MeshackDev: Tests were failing because my class name was star--# instead of stars--# 
Now solved.

ihor-jpeg (approved):

PR #1738, opened 2022-08-22
added  header

Inn0o:
https://Inn0o.github.io/layout_stars/
https://Inn0o.github.io/layout_stars/report/html_report/


src/style.css
```diff
@@ -0,0 +1,29 @@
+* {
+  margin: 0;
+  padding: 0;
+}
```
liliia-livandovska, line 4: its forbidden to use margin: 0; padding: 0; inside *

src/style.css
```diff
@@ -0,0 +1,29 @@
...
+.stars {
+  display: flex;
+}
+
+.stars__star {
+  background-image: url("./images/star.svg");
+  height: 16px;
+  width: 16px;
+  background-position: center;
+  background-size: cover;
```
liliia-livandovska, line 15: remove background-size and add background-repeat: no-repeat;
  background-position: center;

liliia-livandovskaya (changes requested):
GJ! But pay attention to comments

PR #1741, opened 2022-08-22
stars added

evgeny-bobrov:
[DEMO LINK](https://evgeny-bobrov.github.io/layout_stars/)
 [TEST REPORT LINK](https://evgeny-bobrov.github.io/layout_stars/report/html_report/)


src/style.css
```diff
@@ -0,0 +1,34 @@
...
+  background-position: center;
+  margin-right: 4px;
+}
+
+.stars--1 .stars__star:nth-child(-n+1),
+.stars--2 .stars__star:nth-child(-n+2),
+.stars--3 .stars__star:nth-child(-n+3),
+.stars--4 .stars__star:nth-child(-n+4),
+.stars--5 .stars__star:nth-child(-n+5) {
+  height: 16px;
```
alexandra-protyanova, line 29: delete all repetitive styles, they will inherit by default

src/style.css
```diff
@@ -0,0 +1,34 @@
+* {
+  margin: 0;
+  box-sizing: border-box;
+}
+
+.stars {
+  display: flex;
+}
+
+.stars--0 .stars__star:nth-child(-n+5),
```
alexandra-protyanova, line 10: u can set this image for all stars, and add yellow only for needed, it will work the same way

alexandra-protyanova (changes requested):

stanislav-topikha (approved):
GJ 🔥

PR #1744, opened 2022-08-23
Stars block was created.

ShkarupaDenys:
https://ShkarupaDenys.github.io/layout_stars/
https://ShkarupaDenys.github.io/layout_stars/report/html_report/


src/style.css
```diff
@@ -0,0 +1,17 @@
...
+  display: flex;
+}
+
+.stars__star {
+  width: 16px;
+  height: 16px;
+  background: url(./images/star.svg) no-repeat center;
+}
+
+.stars__star.is-active {
```
alexandra-protyanova, line 11: don't use extra class for stars, use `nth-child` instead

alexandra-protyanova (changes requested):
G

src/style.css
ShkarupaDenys: Extra classes for start were deleted.

src/style.css
```diff
@@ -8,10 +8,30 @@
...
+  background: url(./images/star-active.svg) no-repeat center;
+}
+
+.stars--5 .stars__star:nth-child(-n + 5) {
+  background: url(./images/star-active.svg) no-repeat center;
+}
+
+.stars--6 .stars__star:nth-child(-n + 6) {
+  background: url(./images/star-active.svg) no-repeat center;
 }
```
alexandra-protyanova, line 37: combine all these selectors to prevent code duplicating

alexandra-protyanova (changes requested):

src/style.css
ShkarupaDenys: Thanks for advice!

ArturErmolenko (approved):

PR #1753, opened 2022-08-27
stars first commit

metinbicaksiz:
- [DEMO LINK](https://metinbicaksiz.github.io/layout_stars/)
- [TEST REPORT LINK](https://metinbicaksiz.github.io/layout_stars/report/html_report/)


src/style.css
```diff
@@ -0,0 +1,38 @@
...
+  background-image: url(./images/star.svg);
+  background-repeat: no-repeat;
+  background-position: center;
+}
+
+.stars-star:last-child {
+  margin-right: 0;
+}
+
+.stars--1 .stars__star:nth-child(1),
```
amitpatiljc, line 22: you can use a formula like :nth-child(-n + #stars-number) to achieve the same in a cleaner way

amitpatiljc (changes requested):
Well done! I have left a few comments below. Let's fix them!

amitpatiljc (approved):
Again Great Work! But plaese avoid pushing unwanted files
Just push files which are in src folders

PR #1775, opened 2022-09-04
Rating stars

kyrylo-schegoliutin:
[DEMO LINK](https://kyrylo-schegoliutin.github.io/layout_stars/)
[TEST REPORT LINK](https://kyrylo-schegoliutin.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,124 @@
+html,
+body,
+div,
+span,
+applet,
+object,
+iframe,
+h1,
+h2,
+h3,
+h4,
+h5,
+h6,
+p,
+blockquote,
+pre,
+a,
+abbr,
+acronym,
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,109 @@
...
+var,
+b,
+u,
+i,
+center,
+dl,
+dt,
+dd,
+ol,
+ul,
```
Bohdan-Maksymiuk, line 48: do you use all this tags here?
delete for unused

Bohdan-Maksymiuk: and if you use reset - pass it into separate file

Bohdan-Maksymiuk (approved):
Good job 👍

PR #1786, opened 2022-09-09
practice stars block

iamgranovskaya:
[DEMO](https://iamgranovskaya.github.io/layout_stars/)
[TEST](https://iamgranovskaya.github.io/layout_stars/report/html_report/)


src/style.css
```diff
@@ -0,0 +1,25 @@
+* {
+  margin: 0;
+}
```
SanyaBratashchuk, line 3: Do not apply any styles to the `*` selector except `box-sizing`

SanyaBratashchuk (changes requested):
GJ! Check comment)

SanyaBratashchuk (approved):

PR #1821, opened 2022-09-19
layout_stars

07Spirit:
- [DEMO LINK](https://07Spirit.github.io/layout_stars/src/)
- [TEST REPORT LINK](https://07Spirit.github.io/layout_stars/backstop_data/html_report/index.html)

image


src/style.css
```diff
@@ -0,0 +1,53 @@
...
+.stars--2 .stars__star:nth-of-type(5),
+
+.stars--3 .stars__star:nth-of-type(4),
+.stars--3 .stars__star:nth-of-type(5),
+
+.stars--4 .stars__star:nth-of-type(5) {
+  --size: 16px;
+
+  background: url(./images/star.svg) left no-repeat;
+  width: var(--size);
```
SanyaBratashchuk, line 30: too much duplications, set `width` and `height` to the `stars__star` class, and change `background-image` like below 
`.stars--1 > :nth-child(1),
.stars--2 > :nth-child(-n + 2),`

SanyaBratashchuk (changes requested):

src/style.css
```diff
@@ -0,0 +1,34 @@
...
+}
+
+.stars {
+  display: flex;
+  justify-content: space-around;
+  width: 100px;
+}
+
+.stars__star {
+  --size: 16px;
```
SanyaBratashchuk, line 14: move it to the `:root`

src/style.css
```diff
@@ -0,0 +1,34 @@
...
+  height: var(--size);
+}
+
+.stars--0 > *,
+.stars--1 > :nth-child(n + 2),
+.stars--2 > :nth-child(n + 3),
+.stars--3 > :nth-child(n + 4),
+.stars--4 > :nth-child(5) {
+  background: url(./images/star.svg) left no-repeat;
+}
```
SanyaBratashchuk, line 26: set it as default to the stars__star

SanyaBratashchuk (changes requested):

src/style.css
07Spirit: Okey. But redesign one place is good?
I mean that in .stars__star I create all star noactive, and then part of then, again, recreate.

07Spirit:
New links:

- [DEMO LINK](https://07Spirit.github.io/layout_stars/)
- [TEST REPORT LINK](https://07Spirit.github.io/layout_stars/report/html_report/)

SanyaBratashchuk (approved):

PR #1847, opened 2022-09-27
layout_start first version

Vladimir-Zadorozhnyi:
https://Vladimir-Zadorozhnyi.github.io/layout_stars/
https://Vladimir-Zadorozhnyi.github.io/layout_stars/report/html_report/

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,40 @@
+* {
+  margin: 0;
+}
+
+.stars {
+  display: flex;
+  flex-direction: row;
+  justify-content: space-between;
+  align-items: center;
+  height: 16px;
+  width: 96px;
+}
+
+.stars__star {
+  height: 16px;
+  width: 16px;
+  background-image: url(images/star.svg);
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,40 @@
...
+  margin: 0;
+}
+
+.stars {
+  display: flex;
+  flex-direction: row;
+  justify-content: space-between;
+  align-items: center;
+  height: 16px;
+  width: 96px;
```
ArturErmolenko, line 11: this styles are useless you could delete it, display flex is enough
```suggestion
```

ArturErmolenko: don't hard code container width and height

src/style.css
```diff
@@ -0,0 +1,24 @@
+* {
+  margin: 0;
+}
```
SanyaBratashchuk, line 3: do not apply any styles to the `*` selector except `box-sizing`

SanyaBratashchuk (approved):

PR #1856, opened 2022-09-28
stars added

romana-mn:
[DEMO LINK](https://romana-mn.github.io/layout_stars/)
[TEST REPORT LINK](https://romana-mn.github.io/layout_stars/report/html_report/)


SanyaBratashchuk (changes requested):
Add all files

PR #1899, opened 2022-10-01
layout stars is done

vlad-shkv:


src/style.css
```diff
@@ -0,0 +1,57 @@
+body {
+  margin: 0;
+}
+
+.stars {
+  display: flex;
+  flex-direction: row-reverse;
```
mykola-koval, line 7: why reverse?

src/style.css
```diff
@@ -0,0 +1,57 @@
...
+}
+
+.stars__star {
+  position: relative !important;
```
mykola-koval, line 11: do not ever use important 
why is it even needed here?

src/style.css
```diff
@@ -0,0 +1,57 @@
...
+  margin: 0 4px;
+}
+
+.stars-5 .stars__star:nth-child(n+1) {
+  background-image: url(images/star-active.svg);
+  background-repeat: no-repeat;
+  height: 16px;
+  width: 16px;
+  margin: 0 4px;
+}
```
mykola-koval, line 57: you can combine all these rules

mykola-koval (changes requested):
you should pass the tests and fix all the other comments

mykola-koval (changes requested):
seems like you passed the tests and deployed
but the code on github is old
have you pushed the code to github?
follow the instructions

PR #1918, opened 2022-10-04
1st commit

ihor-halkin:
- [DEMO LINK](https://ihor-halkin.github.io/layout_stars/)
- [TEST REPORT LINK](https://ihor-halkin.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,32 @@
+body {
+  margin: 0;
+}
+
+.stars {
+  display: flex;
+  justify-content: flex-start;
+  flex-direction: row;
+}
+
+.stars__star {
+  display: flex;
+  margin-right: 4px;
+  width: 16px;
+  height: 16px;
+  border-radius: 0.5px;
+  background-image: url(/images/star.svg);
+  background-repeat: no-repeat;
+  background-position: center;
+}
+
+.stars__star:last-child {
+  margin: 0;
+}
+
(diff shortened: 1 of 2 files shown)


src/style.css
```diff
@@ -0,0 +1,32 @@
...
+  justify-content: flex-start;
+  flex-direction: row;
+}
+
+.stars__star {
+  display: flex;
+  margin-right: 4px;
+  width: 16px;
+  height: 16px;
+  border-radius: 0.5px;
```
Bohdan-Maksymiuk, line 16: why do you need this border-radius?

src/style.css
```diff
@@ -0,0 +1,32 @@
+body {
+  margin: 0;
+}
+
+.stars {
+  display: flex;
+  justify-content: flex-start;
+  flex-direction: row;
```
Bohdan-Maksymiuk, line 8: useless - default values

Bohdan-Maksymiuk (changes requested):

src/style.css
ihor-halkin: Copy/Paste from Figma)
Del

src/style.css
ihor-halkin: thanks
del

Bohdan-Maksymiuk (approved):

PR #1931, opened 2022-10-08
initial commit

44shaggy:
- [DEMO LINK](https://44shaggy.github.io/layout_stars/)
- [TEST REPORT LINK](https://44shaggy.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -8,6 +8,52 @@
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Stars</h1>
+    <div class="stars stars--0">
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+    </div>
+
+    <div class="stars stars--1">
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+    </div>
+
+    <div class="stars stars--2">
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,28 @@
...
+  width: 16px;
+  margin-right: 4px;
+  background: url(./images/star.svg) no-repeat center;
+}
+
+.stars__star:last-child {
+  margin-right: 0;
+}
+
+.stars--5 .stars__star,
```
amitpatiljc, line 22: do it same way as all other
```suggestion
.stars--5 .stars__star:nth-child(-n + 5),
```

src/index.html
```diff
@@ -8,6 +8,52 @@
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Stars</h1>
+    <div class="stars stars--0">
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
```
amitpatiljc, line 16: you want them in one line so can use inline element instead of block like `span`
```suggestion
      <span class="stars__star"></span>
```

src/style.css
```diff
@@ -0,0 +1,28 @@
+body {
+  margin: 0;
+  display: inline-flex;
+  flex-flow: column;
```
amitpatiljc, line 4: if you use inline tag like `span` in HTML instead of `div` then you do not require line 3 and 4

amitpatiljc (changes requested):

jstmpelowycz (approved):
GJ 🔥

PR #1954, opened 2022-10-20
Stars were added

shyptia:
https://shyptia.github.io/layout_stars/
https://shyptia.github.io/layout_stars/report/html_report/


Ihor-Karpyn (changes requested):
Let's remove margin in the last stars

Ihor-Karpyn (approved):

PR #1968, opened 2022-10-22
stars created

yevhen1408:
https://yevhen1408.github.io/layout_stars/
https://yevhen1408.github.io/layout_stars/report/html_report/

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,26 @@
+* {
+  margin: 0;
+}
+
+.stars {
+  display: flex;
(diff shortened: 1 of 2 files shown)


src/style.css
```diff
@@ -0,0 +1,26 @@
+* {
+  margin: 0;
+}
```
SanyaBratashchuk, line 3: do not reset margins using `*`

SanyaBratashchuk (approved):

PR #1978, opened 2022-10-24
Advance HTML and CSS Lesson 2 - Stars

ysrckr:
- [DEMO LINK](https://ysrckr.github.io/layout_stars/)
- [TEST REPORT LINK](https://ysrckr.github.io/layout_stars/report/html_report/)


src/index.html
```diff
@@ -2,12 +2,59 @@
...
-          content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0">
+    <meta name="viewport" content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0">
     <title>Stars</title>
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Stars</h1>
+    <div class="container">
+      <div class="stars stars--0">
+        <span class="stars--star"></span>
```
mgrinko, line 12: Typically we use -- for a state of an element and __ for a part of an element

src/style.css
```diff
@@ -0,0 +1,58 @@
+html,
+body {
```
mgrinko, line 2: don't group selectors together
Later you can add some other grouping
and finally will need to check many places to get styles for 1 selector

At some point you will diside to change some rules for one selector but not the other one
It is better to write the same rule twice for each selector individually

src/style.css
```diff
@@ -0,0 +1,58 @@
...
+  padding: 0;
+}
+
+.stars {
+  display: flex;
+  align-items: center;
+}
+
+.stars:not(:first-child) {
+  margin-top: -6px;
```
mgrinko, line 13: It is strange for me that you needed this to pass the tests

src/style.css
```diff
@@ -0,0 +1,58 @@
...
+}
+
+.stars--star {
+  width: 20px;
+  height: 20px;
```
mgrinko, line 18: The star is actually 16x16 
please check the design
https://www.figma.com/file/EIBkG1dy1jnK88YPO34Qir/Moyo-%2F-Catalog-(updated)?node-id=11325%3A2961

src/style.css
```diff
@@ -0,0 +1,58 @@
...
+  background-image: url("images/star.svg");
+  background-repeat: no-repeat;
+  margin-left: 2px;
+  margin-top: 2px;
```
mgrinko, line 22: I think if you set size of 16 you can remove this

src/style.css
```diff
@@ -0,0 +1,58 @@
...
+}
+
+.stars--1 > .stars--star:first-child {
+  width: 20px;
+  height: 20px;
+  background-image: url("images/star-active.svg");
+  background-repeat: no-repeat;
+}
```
mgrinko, line 30: I think only `background-image` changes
so no need to duplicate all the other rules

src/style.css
```diff
@@ -0,0 +1,58 @@
...
+
+.stars--4 > .stars--star:nth-child(-n + 4) {
+  width: 20px;
+  height: 20px;
+  background-image: url("images/star-active.svg");
+  background-repeat: no-repeat;
+}
+
+.stars--5 > .stars--star:nth-child(-n + 5) {
```
mgrinko, line 53: It is a good example when you actually can group selector because they MUST have the same styling and one can't change without the others

mgrinko (changes requested):

mgrinko (approved):

PR #1993, opened 2022-10-25
Made a layout with stars

ia-lerimund:
- [DEMO LINK](https://ia-lerimund.github.io/layout_stars/)
- [TEST REPORT LINK](https://ia-lerimund.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,38 @@
+* {
+  margin: 0;
+  box-sizing: border-box;
+}
+
+.stars {
+  display: flex;
+}
+
+.stars__star {
+  height: 16px;
+  width: 16px;
+  background-repeat: no-repeat;
+  background-position: center;
+  margin: 0 2px;
+  padding: 0.5px;
+}
+
+.stars--0 .stars__star,
+.stars--1 .stars__star:nth-child(1n),
+.stars--2 .stars__star:nth-child(1n+3),
+.stars--3 .stars__star:nth-child(1n+4),
+.stars--4 .stars__star:nth-child(1n+5),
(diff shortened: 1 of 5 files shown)


src/style.css
```diff
@@ -0,0 +1,38 @@
...
+  padding: 0.5px;
+}
+
+.stars--0 .stars__star,
+.stars--1 .stars__star:nth-child(1n),
+.stars--2 .stars__star:nth-child(1n+3),
+.stars--3 .stars__star:nth-child(1n+4),
+.stars--4 .stars__star:nth-child(1n+5),
+.stars--5 .stars__star:nth-child(1n+6) {
+  background-image: url(./images/star.svg);
```
DarkMistyRoom, line 25: you can do it in the main selector

src/style.css
```diff
@@ -0,0 +1,38 @@
...
+.stars {
+  display: flex;
+}
+
+.stars__star {
+  height: 16px;
+  width: 16px;
+  background-repeat: no-repeat;
+  background-position: center;
+  margin: 0 2px;
```
DarkMistyRoom, line 15: image
and better use one-direction margins. for example margin-right: 4px.

DarkMistyRoom (changes requested):

src/style.css
ia-lerimund: Thank you! Changed styles but tests failed :( What can be the reason?

ivanchuk-yurii (approved):
I think you haven't deployed your last changes, but this error in tests is minor. however remember that you should do `npm run deploy` after every changes in pr

PR #2011, opened 2022-10-30
Develop

Vaskovics:
https://vaskovics.github.io/layout_stars/
https://vaskovics.github.io/layout_stars/report/html_report/


src/style.css
```diff
@@ -0,0 +1,24 @@
...
+  display: flex;
+}
+
+.stars__star {
+  width: 16px;
+  height: 16px;
+  margin-right: 4px;
+  background-image: url("./images/star.svg");
+  background-repeat: no-repeat;
+  background-position: center;
```
ivanchuk-yurii, line 15: it's better to move everything inside `background` property

ivanchuk-yurii (approved):

PR #2019, opened 2022-10-31
star-box was created

oleksandrnikora:
- [DEMO LINK](https://oleksandrnikora.github.io/layout_stars/)
- [TEST REPORT LINK](https://oleksandrnikora.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,7 +1,9 @@
 * {
-  margin: 0;
-  padding: 0;
   box-sizing: border-box;
+}
+
+body {
+  margin: 0;
 }
 
 .stars {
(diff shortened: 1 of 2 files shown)


src/style.css
```diff
@@ -0,0 +1,29 @@
+* {
+  margin: 0;
+  padding: 0;
+  box-sizing: border-box;
+}
```
Ihor-Karpyn, line 5: Do not use the selector `*` to reset something

Ihor-Karpyn (changes requested):

Ihor-Karpyn (approved):

PR #2034, opened 2022-11-02
layout_stars ready

UmizDemud:
- [DEMO LINK](https://UmizDemud.github.io/layout_stars/)
- [TEST REPORT LINK](https://UmizDemud.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -8,6 +8,74 @@
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Stars</h1>
+    <main class="main">
+      <div class="stars stars--0">
+        <div class="stars__star"></div>
+
+        <div class="stars__star"></div>
+
+        <div class="stars__star"></div>
+
+        <div class="stars__star"></div>
+
+        <div class="stars__star"></div>
+      </div>
+      <div class="stars stars--1">
+        <div class="stars__star"></div>
+
+        <div class="stars__star"></div>
+
+        <div class="stars__star"></div>
+
+        <div class="stars__star"></div>
+
+        <div class="stars__star"></div>
+      </div>
+      <div class="stars stars--2">
+        <div class="stars__star"></div>
+
+        <div class="stars__star"></div>
+
+        <div class="stars__star"></div>
+
+        <div class="stars__star"></div>
+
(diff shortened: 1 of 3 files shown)


src/index.html
```diff
@@ -8,6 +8,74 @@
...
+
+        <div class="stars__star"></div>
+
+        <div class="stars__star"></div>
+
+        <div class="stars__star"></div>
+
+        <div class="stars__star"></div>
+      </div>
+      <div class="stars stars--1">
```
mgrinko, line 23: Don't add empty line if siblings are single line
Just is each sibling element has several lines

```suggestion
      <div class="stars stars--0">
        <div class="stars__star"></div>
        <div class="stars__star"></div>
        <div class="stars__star"></div>
      </div>
      
      <div class="stars stars--1">
        <div class="stars__star"></div>
        <div class="stars__star"></div>
```

src/style.css
```diff
@@ -0,0 +1,49 @@
...
+}
+
+body {
+  margin: 0;
+}
+
+.main {
+  display: flex;
+  flex-direction: column;
+}
```
mgrinko, line 12: I don't think it is required

src/style.css
```diff
@@ -0,0 +1,49 @@
...
+
+.stars {
+  display: flex;
+}
+
+.stars > .stars__star {
+  width: 16px;
+  height: 16px;
+  margin-right: 4px;
```
mgrinko, line 21: what about the last star?

src/style.css
```diff
@@ -0,0 +1,49 @@
...
+  background-image: url("./images/star.svg");
+}
+
+.stars--4 > .stars__star:nth-child(n+5) {
+  background-image: url("./images/star.svg");
+}
+
+.stars--5 > .stars__star:nth-child(n+6) {
+  background-image: url("./images/star.svg");
+}
```
mgrinko, line 49: It is a good example when you actually can group selector because they MUST have the same styling and one can't change without the others

mgrinko (approved):

src/style.css
UmizDemud: makes sense

PR #2050, opened 2022-11-03
passed task

Alireza-Mohammadhossein:


src/index.html
```diff
@@ -1,13 +1,66 @@
...
+        <span class="stars__star"></span>
+        <span class="stars__star"></span>
+        <span class="stars__star"></span>
+        <span class="stars__star"></span>
+        <span class="stars__star"></span>
+      </div>
+
+    </main>
+
   </body>
```
mgrinko, line 65: No need to put a blank line between a parent and its child
```suggestion
        <span class="stars__star"></span>
      </div>
    </main>
  </body>
```

src/style.css
```diff
@@ -0,0 +1,27 @@
+* {
+  margin: 0;
+  padding: 0;
+  box-sizing: border-box;
+}
```
mgrinko, line 5: It is better to set it where it is required but not for all the elements

mgrinko (approved):

PR #2057, opened 2022-11-03
initial commit

CharAnagn:
- [DEMO LINK](https://CharAnagn.github.io/layout_stars/)
- [TEST REPORT LINK](https://CharAnagn.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,23 @@
+body {
+  margin: 0;
+}
+
+.stars {
+  display: flex;
+}
+
+.stars__star {
+  padding: 8px;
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,23 @@
+body {
+  margin: 0;
+}
+
+.stars {
+  display: flex;
+}
+
+.stars__star {
+  padding: 8px;
```
mgrinko, line 10: I think it should be height and width of 16px but not padding

mgrinko (approved):

PR #2076, opened 2022-11-04
task finished

RedkoSerhii:
- [DEMO LINK](https://RedkoSerhii.github.io/layout_stars/)
- [TEST REPORT LINK](https://RedkoSerhii.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -8,6 +8,54 @@
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Stars</h1>
+
+    <ul class="block stars-1">
+      <li class="stars_star"></li>
+      <li class="stars_star"></li>
+      <li class="stars_star"></li>
+      <li class="stars_star"></li>
+      <li class="stars_star"></li>
+    </ul>
+
+    <ul class="block stars-2">
+      <li class="stars_star"></li>
+      <li class="stars_star"></li>
+      <li class="stars_star"></li>
+      <li class="stars_star"></li>
+      <li class="stars_star"></li>
+    </ul>
+
+    <ul class="block stars-3">
+      <li class="stars_star"></li>
+      <li class="stars_star"></li>
+      <li class="stars_star"></li>
+      <li class="stars_star"></li>
+      <li class="stars_star"></li>
+    </ul>
+
+    <ul class="block stars-4">
+      <li class="stars_star"></li>
+      <li class="stars_star"></li>
+      <li class="stars_star"></li>
+      <li class="stars_star"></li>
+      <li class="stars_star"></li>
+    </ul>
+
+    <ul class="block stars-5">
+      <li class="stars_star"></li>
+      <li class="stars_star"></li>
+      <li class="stars_star"></li>
+      <li class="stars_star"></li>
+      <li class="stars_star"></li>
+    </ul>
+
+    <ul class="block stars-6">
+      <li class="stars_star"></li>
+      <li class="stars_star"></li>
+      <li class="stars_star"></li>
+      <li class="stars_star"></li>
+      <li class="stars_star"></li>
+    </ul>
+
   </body>
 </html>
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,33 @@
...
+  list-style: none;
+}
+
+.stars_star {
+  height: 16px;
+  width: 16px;
+  margin-right: 4px;
+
+  background-image: url(images/star.svg);
+  background-repeat: no-repeat;
```
Ihor-Karpyn, line 19: https://developer.mozilla.org/en-US/docs/Web/CSS/background-position

Ihor-Karpyn (changes requested):

src/style.css
RedkoSerhii: fixed it, good advice

src/style.css
```diff
@@ -0,0 +1,32 @@
...
+
+  list-style: none;
+}
+
+.stars_star {
+  height: 16px;
+  width: 16px;
+  margin-right: 4px;
+
+  background: url(images/star.svg) no-repeat;
```
Ihor-Karpyn, line 18: Fix this line, run the test again, and deploy changes. I gave u this [link](https://developer.mozilla.org/en-US/docs/Web/CSS/background-position) that u add `background-position: center;`

```suggestion
  background-image: url(images/star.svg);
  background-repeat: no-repeat;
  background-position: center;
```

Ihor-Karpyn (changes requested):

src/style.css
RedkoSerhii: I understand now, I thought you ask me write shortly %)
Fixed it.

src/index.html
```diff
@@ -8,6 +8,54 @@
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Stars</h1>
+
+    <ul class="block stars-1">
+      <li class="stars_star"></li>
```
ivan-kadykalo, line 13: Check requirements in readme
```suggestion
    <ul class="stars stars--1">
      <li class="stars_star"></li>
```

src/style.css
```diff
@@ -0,0 +1,34 @@
...
+.stars_star:last-child {
+  margin: 0;
+}
+
+.stars-2 .stars_star:nth-child(1),
+.stars-3 .stars_star:not(:nth-child(n + 3)),
+.stars-4 .stars_star:not(:nth-child(n + 4)),
+.stars-5 .stars_star:not(:nth-child(n + 5)),
+.stars-6 .stars_star:not(:nth-child(n + 6)),
+.stars-7 .stars_star:not(:nth-child(n + 7)) {
```
ivan-kadykalo, line 32: Would be better to use the number in the class name to set how many stars should be active. Why had you used :not operator here? You can do it as you did in the first row
`.stars--2 :nth-child(-n+2)`

src/style.css
```diff
@@ -0,0 +1,34 @@
...
+  background-image: url(images/star.svg);
+  background-repeat: no-repeat;
+  background-position: center;
+}
+
+.stars_star:last-child {
+  margin: 0;
+}
+
+.stars-2 .stars_star:nth-child(1),
```
ivan-kadykalo, line 27: We can write it in shorter
```suggestion
.stars--1 :nth-child(1),
```

ivan-kadykalo (changes requested):
Have you checked your tests?

src/index.html
RedkoSerhii: fixed it

src/style.css
RedkoSerhii: wow, I needed your advice early. Fixed. Looks great!

src/style.css
RedkoSerhii: fixed

ivan-kadykalo (approved):
GJ🔥

PR #2080, opened 2022-11-04
Develop

dolnaro:
- [DEMO LINK](https://dolnaro.github.io/layout_stars/)
- [TEST REPORT LINK](https://dolnaro.github.io/layout_stars/report/html_report/)


src/style.css
```diff
@@ -0,0 +1,26 @@
+body {
+  margin: 0;
+}
+
+.stars {
+  display: flex;
+}
+
+/* #region stars__star */
```
naziamoff, line 9: make sure to remove all the unnecessary comments before commiting

dolnaro: > 

```suggestion
/* #region stars__star */
```
So shouldn't I use this to wrap the CSS selector ever?

PR #2083, opened 2022-11-04
Develop

HiDmytro:
- [DEMO LINK](https://HiDmytro.github.io/layout_stars/)
- [TEST REPORT LINK](https://HiDmytro.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -4,13 +4,11 @@ body {
 
 .stars {
   display: flex;
-  justify-content: space-between;
-  align-items: center;
-  width: 96px;
 }
 
 .stars__star {
   background: url(./images/star.svg)2px 2px no-repeat;
+  margin-right: 4px;
   height: 16px;
   width: 16px;
 }


src/style.css
```diff
@@ -0,0 +1,22 @@
+body {
+  margin: 0;
+}
+
+.stars {
+  display: flex;
+}
+
+.stars__star {
+  background: url(./images/star.svg)2px 2px no-repeat;
```
ivanchuk-yurii, line 10:
```suggestion
  background: url(./images/star.svg) 2px 2px no-repeat;
```

src/style.css
```diff
@@ -0,0 +1,22 @@
+body {
+  margin: 0;
+}
+
+.stars {
```
ivanchuk-yurii, line 5: it's better to move all `starts` styles into separate file

ivanchuk-yurii (approved):

PR #2087, opened 2022-11-04
Edited

Paryax:
- [DEMO LINK](https://Paryax.github.io/layout_stars/)
- [TEST REPORT LINK](https://Paryax.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -9,7 +9,7 @@ body {
 .stars__star {
   width: 16px;
   height: 16px;
-  background-image: url(images/star.svg);
+  background-image: url(images/star-active.svg);
   background-repeat: no-repeat;
   background-position: center;
 
@@ -26,5 +26,5 @@ body {
 .stars--3 :nth-child(n + 4),
 .stars--4 :nth-child(n + 5),
 .stars--5 :nth-child(n + 6) {
-  background-image: url(images/star-active.svg);
+  background-image: url(images/star.svg);
 }
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,30 @@
+body {
+  margin: 0;
+}
+
+.stars {
+  display: flex;
+}
+
+.stars__star {
+  width: 16px;
+  height: 16px;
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,30 @@
...
+}
+
+.stars {
+  display: flex;
+}
+
+.stars__star {
+  width: 16px;
+  height: 16px;
+  background-image: url(images/star-active.svg);
```
ivan-kadykalo, line 12: Use inactive star here

src/style.css
```diff
@@ -0,0 +1,30 @@
...
+  margin-right: 0;
+}
+
+.stars--0 .stars__star:nth-child(n),
+.stars--1 .stars__star:nth-child(n + 2),
+.stars--2 .stars__star:nth-child(n + 3),
+.stars--3 .stars__star:nth-child(n + 4),
+.stars--4 .stars__star:nth-child(n + 5),
+.stars--5 .stars__star:nth-child(n + 6) {
+  background-image: url(images/star.svg);
```
ivan-kadykalo, line 29: And here use active stare

src/style.css
```diff
@@ -0,0 +1,30 @@
...
+  margin-right: 4px;
+}
+
+.stars__star:last-child {
+  margin-right: 0;
+}
+
+.stars--0 .stars__star:nth-child(n),
+.stars--1 .stars__star:nth-child(n + 2),
+.stars--2 .stars__star:nth-child(n + 3),
```
ivan-kadykalo, line 25: We can write it shorter
```suggestion
.stars--2 :nth-child(n + 3),
```

ivan-kadykalo (changes requested):
Demo links are not active. Delete <> braces from the link

src/style.css
Paryax: If this is done, then the asterisks will be located differently than in the task condition

ivan-kadykalo (approved):
🚀

PR #2116, opened 2022-11-08
layout stars

NikolaienkoN:
NikolaienkoN.github.io/layout_stars/
NikolaienkoN.github.io/layout_stars/report/html_report/

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,27 @@
+body {
+  margin: 0;
+}
+
+.stars {
+  display: flex;
+  flex-direction: row;
+  justify-content: start;
+}
+
+.stars__star {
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,27 @@
...
+  flex-direction: row;
+  justify-content: start;
+}
+
+.stars__star {
+  background-image: url(./images/star.svg);
+  background-repeat: no-repeat;
+  width: 16px;
+  height: 16px;
+  background-position: center;
```
ivanchuk-yurii, line 16: it's better to combine all into `background` property

ivanchuk-yurii (approved):

PR #2126, opened 2022-11-10
work done

Justfun1517:
https://Justfun1517.github.io/layout_stars/
https://Justfun1517.github.io/layout_stars/report/html_report/


liliia-livandovskaya (approved):
Great job! 🔥

PR #2130, opened 2022-11-13
added blocks of stars

IamLoren:
https://iamloren.github.io/layout_stars/
https://iamloren.github.io/layout_stars/report/html_report/


src/style.css
```diff
@@ -0,0 +1,59 @@
...
+  background-image: url(./images/star.svg);
+}
+
+.stars--2 .stars__star:nth-of-type(-n+2) {
+  background-image: url(./images/star-active.svg);
+}
+
+.stars--2 .stars__star:not(:nth-of-type(-n+2)) {
+  background-image: url(./images/star.svg);
+}
```
Bohdan-Maksymiuk, line 39: try to set active star by default and change where you need on not-active by combining selectors

Bohdan-Maksymiuk (changes requested):
Screenshot 2022-11-15 at 09 24 08

Good job, try to fix comments

src/style.css
IamLoren: wow! It was unexpected decision for me, but it works ) Thank you )

Virass (approved):
Good job 👍

PR #2148, opened 2022-11-17
Develop

Siriuse911:
- [DEMO LINK](https://siriuse911.github.io/layout_stars/)
- [TEST REPORT LINK](https://siriuse911.github.io/layout_stars/report/html_report/)


src/style.css
```diff
@@ -0,0 +1,31 @@
...
+  background-repeat: no-repeat;
+  background-position: center;
+  
+}
+
+.stars__star:not(:first-child) {
+  margin-left: 4px;
+}
+
+.stars__gold {
```
polosanya, line 27: Interesting solution 👍
But in this task we learn how to you use pseudo-classes in CSS, so please try to redo it using them.
It would be useful for you
https://developer.mozilla.org/en-US/docs/Web/CSS/Pseudo-classes#tree-structural_pseudo-classes

polosanya (changes requested):
Well done! Please check my comment

nazarmatsevych (approved):

PR #2180, opened 2022-11-27
Block stars

Andrey-Bikir:


src/style.css
```diff
@@ -0,0 +1,31 @@
+body {
+  margin: 0;
+  padding: 0;
```
ivanchuk-yurii, line 3:
```suggestion
```

src/style.css
```diff
@@ -0,0 +1,31 @@
...
+  display: flex;
+  list-style: none;
+  margin: 0;
+  padding: 0;
+}
+
+.stars__star {
+  background-image: url("images/star.svg");
+  background-repeat: no-repeat;
+  background-position: center;
```
ivanchuk-yurii, line 16: combine in `background` property

ivanchuk-yurii (changes requested):

qoa11a (approved):

PR #2186, opened 2022-11-29
first pull stars layout

Spec0s:
https://spec0s.github.io/layout_stars/
https://spec0s.github.io/layout_stars/report/html_report/


src/style.css
```diff
@@ -0,0 +1,41 @@
...
+  background-image: url(./images/star.svg);
+}
+
+.stars--3 > .stars__star:nth-child(n+4) {
+  background-image: url(./images/star.svg);
+}
+
+.stars--4 > .stars__star:nth-child(n+5) {
+  background-image: url(./images/star.svg);
+}
```
amitpatiljc, line 41: You can group these. and no need to combine
```suggestion
.stars--1 :nth-child(1),
.stars--2 :nth-child(-n+2),
.stars--3 :nth-child(-n+3),
.stars--4 :nth-child(-n+4),
.stars--5 :nth-child(-n+5), {
  background-image: url(./images/star.svg);
}
```

amitpatiljc (changes requested):

PR #2187, opened 2022-11-30
stars was added

Nikita211112:
https://Nikita211112.github.io/layout_stars/

https://Nikita211112.github.io/layout_stars/report/html_report/

I'm not sure that I correctly executed the code in css. From the rules for the task, I realized that you can’t add a second class to the block  'stars__star' like this (class='stars__star stars__star--active'  - for adding a yellow star)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,43 @@
+* {
+  box-sizing: border-box;
+}
+
+body {
+  margin: 0;
+}
+
+.stars {
+  display: flex;
+}
+
+.stars__star {
+  width: 16px;
+  height: 16px;
+  background-image: url(images/star.svg);
+  background-repeat: no-repeat;
+  background-position: center;
+}
+
+.stars__star + .stars__star {
+  margin-left: 4px;
+}
+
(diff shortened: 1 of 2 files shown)


src/style.css
```diff
@@ -0,0 +1,43 @@
...
+  background-image: url(images/star-active.svg);
+}
+
+.stars--4 > .stars__star:nth-child(-n + 4) {
+  background-image: url(images/star-active.svg);
+}
+
+.stars--5 > .stars__star:nth-child(-n + 5) {
+  background-image: url(images/star-active.svg);
+}
```
Viktor-Kost, line 43: Group 

```suggestion
.stars--1 > .stars__star:nth-child(-n + 1),
.stars--2 > .stars__star:nth-child(-n + 2),
.................... {
  background-image: url(images/star-active.svg);
}
```

Viktor-Kost (changes requested):

src/style.css
Nikita211112: > Group

ready

Viktor-Kost (approved):
GJ 👍

PR #2215, opened 2022-12-06
Develop

ihor-soloviov:
- [DEMO LINK](https://ihor-soloviov.github.io/layout_stars/)
- [TEST REPORT LINK](https://ihor-soloviov.github.io/layout_stars/report/html_report/)


src/style.css
```diff
@@ -0,0 +1,36 @@
+* {
+  margin: 0;
+}
```
Anna-Tkachenko, line 3: don't reset margins using * selector

src/style.css
```diff
@@ -0,0 +1,36 @@
...
+
+.stars--0 .stars__star,
+.stars--1 .stars__star,
+.stars--2 .stars__star,
+.stars--3 .stars__star,
+.stars--4 .stars__star,
+.stars--5 .stars__star {
+  background-position: center;
+  background-repeat: no-repeat;
```
Anna-Tkachenko, line 27: move these styles to the main stars__star class

Anna-Tkachenko (changes requested):

Anna-Tkachenko (approved):
⭐️

PR #2225, opened 2022-12-07
add stars block

Dashaverz:
https://Dashaverz.github.io/layout_stars/
https://Dashaverz.github.io/layout_stars/report/html_report/


src/style.css
```diff
@@ -0,0 +1,24 @@
...
+  display: flex;
+}
+
+.stars__star {
+  background: url(./images/star-active.svg);
+  width: 16px;
+  height: 16px;
+  background-repeat: no-repeat;
+  background-position: center;
+  margin-right: 4px;
```
Anna-Tkachenko, line 15: will be good to reset margin-right for the last star

Anna-Tkachenko (approved):
⭐️

PR #2226, opened 2022-12-07
Performed the task

maasiimka:


src/style.css
```diff
@@ -0,0 +1,25 @@
...
+}
+
+.stars {
+  display: flex;
+}
+
+.stars__star {
+  width: 16px;
+  height: 16px;
+  margin-right: 4px;
```
Anna-Tkachenko, line 13: will be better to reset margin for the last star

Anna-Tkachenko (approved):
⭐️

PR #2234, opened 2022-12-07
stars commit

serbinm:
- [DEMO LINK](https://serbinm.github.io/layout_stars/)
- [TEST REPORT LINK](https://serbinm.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,24 @@
+body {
+  margin: 0;
+}
+
+.stars {
+  display: flex;
+}
+
+.stars__star {
+  width: 20px;
+  height: 16px;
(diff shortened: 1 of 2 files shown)


src/style.css
```diff
@@ -0,0 +1,24 @@
...
+  margin: 0;
+}
+
+.stars {
+  display: flex;
+}
+
+.stars__star {
+  width: 20px;
+  height: 16px;
```
Anna-Tkachenko, line 11: will be better to set correct size for the stars (16px - 16px), and set correct margins between them

Anna-Tkachenko (approved):
⭐️

PR #2245, opened 2022-12-08
initial commit

artem-sobolevskyi:
- [DEMO LINK](https://artem-sobolevskyi.github.io/layout_stars/)
- [TEST REPORT LINK](https://artem-sobolevskyi.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -7,7 +7,6 @@ body {
   width: 16px;
   height: 16px;
   background-position: center;
-  /* background-size: contain; */
   margin-right: 4px;
   background-repeat: no-repeat;
 }
@@ -16,10 +15,7 @@ body {
   display: flex;
 }
 
-.stars--6 .stars__star {
-  background-image: url(./images/star-active.svg);
-}
-
+.stars--6 .stars__star,
 .stars--5 .stars__star {
   background-image: url(./images/star-active.svg);
 }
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,41 @@
+body {
+  margin: 0;
+}
+
+.stars__star {
+  background-image: url(./images/star.svg);
+  width: 16px;
+  height: 16px;
+  background-position: center;
+  /* background-size: contain; */
```
Anna-Tkachenko, line 10: delete unused comments

src/style.css
```diff
@@ -0,0 +1,41 @@
...
+  display: flex;
+}
+
+.stars--6 .stars__star {
+  background-image: url(./images/star-active.svg);
+}
+
+.stars--5 .stars__star {
+  background-image: url(./images/star-active.svg);
+}
```
Anna-Tkachenko, line 25: check checklist, there is the rule about the case, when you want to set the same property for many classes

Anna-Tkachenko (changes requested):

Anna-Tkachenko (changes requested):
check point 2 in checklist

artem-sobolevskyi:
Sorry, I thought you meant only the highlighted part of the code, so I thought you should only group elements with similar selectors..

Anna-Tkachenko (approved):
⭐️

PR #2252, opened 2022-12-08
stars

Tsvylenyeva:
- [DEMO LINK](https://Tsvylenyeva.github.io/layout_stars/)
- [TEST REPORT LINK](https://Tsvylenyeva.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,32 @@
+* {
+  margin: 0;
+  padding: 0;
+}
+
+.container {
+  background-color: white;
+  position: relative;
+}
+
+.yellow {
+  display: flex;
+  justify-content: flex-start;
+}
+
+.stars__star {
+  width: 16px;
+  height: 16px;
+  background-image: url(images/star.svg);
+  background-position: center center;
+  background-size: contain;
+  flex-wrap: wrap;
+  margin: 1px;
+}
+
+.stars--1 .stars__star:nth-child(1),
+.stars--2 .stars__star:nth-child(-n+2),
+.stars--3 .stars__star:nth-child(-n+3),
+.stars--4 .stars__star:nth-child(-n+4),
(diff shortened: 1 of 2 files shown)


src/style.css
```diff
@@ -0,0 +1,32 @@
...
+
+.container {
+  background-color: white;
+  position: relative;
+}
+
+.yellow {
+  display: flex;
+  justify-content: flex-start;
+}
```
Anna-Tkachenko, line 14: do you need this class? can you move these styles to the main `stars__star` class?

src/style.css
```diff
@@ -0,0 +1,32 @@
...
+
+.stars__star {
+  width: 16px;
+  height: 16px;
+  background-image: url(images/star.svg);
+  background-position: center center;
+  background-size: contain;
+  flex-wrap: wrap;
```
Anna-Tkachenko, line 22: do you need this property?

src/style.css
```diff
@@ -0,0 +1,32 @@
...
+.yellow {
+  display: flex;
+  justify-content: flex-start;
+}
+
+.stars__star {
+  width: 16px;
+  height: 16px;
+  background-image: url(images/star.svg);
+  background-position: center center;
```
Anna-Tkachenko, line 20: check background for the each star, right now background image is repeated, fix it

Anna-Tkachenko (changes requested):

src/style.css
```diff
@@ -0,0 +1,30 @@
+* {
+  margin: 0;
+  padding: 0;
+}
```
Denchepornyuk, line 4: don't do it, it's a bad practice to add `margin: 0;` and `padding: 0;` for tag `*`
Not all tags have default margin and padding, tag `*` works for all tags on the page

Denchepornyuk (changes requested):

PR #2262, opened 2022-12-10
Develop

RufatRM:
https://RufatRM.github.io/layout_stars/
https://RufatRM.github.io/layout_stars/report/html_report/

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,36 @@
+
+* {
+  margin: 0;
+}
+
+.stars {
+  display: flex;
+  list-style: none;
+  padding-left: 0;
+}
+
+.stars__star {
+  width: 16px;
+  height: 16px;
+  margin-right: 4px;
+
+  background-repeat: no-repeat;
+  background-size: auto;
+  background-position: center;
+}
+
+.stars--0 .stars__star,
+.stars--1 .stars__star:nth-child(n+2),
+.stars--2 .stars__star:nth-child(n+3),
+.stars--3 .stars__star:nth-child(n+4),
+.stars--4 .stars__star:nth-child(n+5) {
+  background-image: url(images/star.svg);
+}
+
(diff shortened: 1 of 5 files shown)


src/style.css
```diff
@@ -0,0 +1,36 @@
+
```
qoa11a, line 1: Redundant empty row.

src/style.css
```diff
@@ -0,0 +1,36 @@
...
+.stars {
+  display: flex;
+  list-style: none;
+  padding-left: 0;
+}
+
+.stars__star {
+  width: 16px;
+  height: 16px;
+  margin-right: 4px;
```
qoa11a, line 15: Last star shouldn't have margin-right.

src/style.css
```diff
@@ -0,0 +1,36 @@
...
+
+  background-repeat: no-repeat;
+  background-size: auto;
+  background-position: center;
```
qoa11a, line 19:
```suggestion
  background: url(images/star.svg) no-repeat center auto;
```

src/style.css
```diff
@@ -0,0 +1,36 @@
...
+}
+
+.stars--0 .stars__star,
+.stars--1 .stars__star:nth-child(n+2),
+.stars--2 .stars__star:nth-child(n+3),
+.stars--3 .stars__star:nth-child(n+4),
+.stars--4 .stars__star:nth-child(n+5) {
+  background-image: url(images/star.svg);
+}
```
qoa11a, line 28: You could use this image for every star. Than you just rewrite it with code below.

qoa11a (changes requested):
Good job! 👍 
Few small fixes.

src/style.css
RufatRM: Hi, `background: url(images/star.svg) no-repeat center auto;` for some reason this doesn't work.

PR #2280, opened 2022-12-14
Develop

maxonshev:
https://maxonshev.github.io/layout_stars/
https://maxonshev.github.io/layout_stars/report/html_report/


src/index.html
```diff
@@ -8,6 +8,54 @@
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Stars</h1>
+
+    <div class="stars stars--0">
+      <div class="stars_star"></div>
```
UnknownAI, line 13: for this class name should be two underscores instead of one, please change this everywhere

src/style.css
```diff
@@ -0,0 +1,28 @@
...
+  background-repeat: no-repeat;
+  background-position: center;
+  margin-right: 4px;
+}
+
+.stars_star:last-child {
+  margin-right: 0;
+}
+
+.stars--1 .stars_star:first-child,
```
UnknownAI, line 22: not a big mistake, but it is better to have the same structure and use `nth-child` instead `first-child`

UnknownAI (changes requested):

ViktorHavryliuk (approved):

SanyaBratashchuk (approved):

PR #2287, opened 2022-12-17
Stars

SashaKozoriz:


src/style.css
```diff
@@ -0,0 +1,28 @@
...
+  display: flex;
+}
+
+.stars__star {
+  margin-right: 4px;
+  height: 16px;
+  width: 16px;
+  background-image: url(images/star.svg);
+  background-position: center;
+  background-repeat: no-repeat;
```
ivanchuk-yurii, line 15: it's better to combine in `background` property

ivanchuk-yurii (approved):

PR #2292, opened 2022-12-19
layout stars was created

Daria13k:
https://Daria13k.github.io/layout_stars/
https://Daria13k.github.io/layout_stars/report/html_report/

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,40 @@
+* {
+  margin: 0;
+}
+
+body {
+  display: block;
+}
+
+.stars {
+  display: flex;
+}
+
+.stars__star {
+  width: 16px;
+  height: 16px;
+  margin-right: 4px;
+  background-repeat: no-repeat;
+  background-position: center;
+  background-image: url(images/star.svg);
+}
+
+.stars--1 .stars__star:nth-child(-n+1) {
+  background-image: url(images/star-active.svg);
+}
+
+.stars--2 .stars__star:nth-child(-n+2) {
+  background-image: url(images/star-active.svg);
+}
+
+.stars--3 .stars__star:nth-child(-n+3) {
+  background-image: url(images/star-active.svg);
+}
+
+.stars--4 .stars__star:nth-child(-n+4) {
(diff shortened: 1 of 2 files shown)


src/style.css
```diff
@@ -0,0 +1,40 @@
...
+  background-image: url(images/star-active.svg);
+}
+
+.stars--4 .stars__star:nth-child(-n+4) {
+  background-image: url(images/star-active.svg);
+}
+
+.stars--5 .stars__star:nth-child(-n+5) {
+  background-image: url(images/star-active.svg);
+}
```
witflash, line 40: You can combine all rules into short notation like this:
```
.... {
  // rules for all these cases
}
```

witflash (changes requested):
Amazing!
Try to fix my comment and your work will be pretty well! 👍🏼

src/style.css
Daria13k: I combined the rules by recommendation, thank you!

src/style.css
```diff
@@ -0,0 +1,28 @@
+* {
+  margin: 0;
+}
+
+body {
+  display: block;
+}
```
OlEzhA0, line 7: why do we need this?

src/style.css
```diff
@@ -0,0 +1,28 @@
+* {
+  margin: 0;
+}
```
OlEzhA0, line 3: Would be better to reset margin on `body` directly

OlEzhA0 (changes requested):
Hey! You are almost done
please check my comments

src/style.css
Daria13k: removed

src/style.css
Daria13k: replaced

src/style.css
```diff
@@ -0,0 +1,25 @@
+body {
+  display: block;
```
witflash, line 2: Redundant. You don't need to use `block` for `body`. It's a block element already.

witflash (approved):
Well done!
I left a small note for your notice.

PR #2293, opened 2022-12-19
stars were added

Pavlo-Petrashevskyi:
- [DEMO LINK](https:/<Pavlo-Petrashevskyi>.github.io/layout_stars/)
- [TEST REPORT LINK](https://<Pavlo-Petrashevskyi>.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -8,6 +8,76 @@
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Stars</h1>
+    <div class="stars stars--0">
+      <div class="stars__star"></div>
+
+      <div class="stars__star"></div>
+
+      <div class="stars__star"></div>
+
+      <div class="stars__star"></div>
+
+      <div class="stars__star"></div>
+    </div>
+    
+    <div class="stars stars--1">
+      <div class="stars__star stars__active"></div>
+
+      <div class="stars__star"></div>
+
+      <div class="stars__star"></div>
+
+      <div class="stars__star"></div>
+
+      <div class="stars__star"></div>
+    </div>
+
+    <div class="stars stars--2">
+      <div class="stars__star stars__active"></div>
+
+      <div class="stars__star stars__active"></div>
+
+      <div class="stars__star"></div>
+
+      <div class="stars__star"></div>
+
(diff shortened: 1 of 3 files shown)


src/index.html
```diff
@@ -8,6 +8,76 @@
...
+
+      <div class="stars__star"></div>
+
+      <div class="stars__star"></div>
+
+      <div class="stars__star"></div>
+    </div>
+    
+    <div class="stars stars--1">
+      <div class="stars__star stars__active"></div>
```
SanyaBratashchuk, line 24:
```suggestion
      <div class="stars__star"></div>
```
Remove it everywhere use nth-child instead

SanyaBratashchuk (changes requested):
Links is not working

src/index.html
Pavlo-Petrashevskyi: Done, but is it okay if I have written it in many selectors? Like in the "Update №2" or there is a better and easier way to write with 'nth-child'?

src/style.css
```diff
@@ -0,0 +1,29 @@
...
+}
+
+.stars--1 .stars__star:nth-child(1),
+.stars--2 .stars__star:nth-child(-n+2),
+.stars--3 .stars__star:nth-child(-n+3),
+.stars--4 .stars__star:nth-child(-n+4),
+.stars--5 .stars__star:nth-child(n) {
+  background-image: url("images\star-active.svg");
+  background-repeat: no-repeat;
+  background-position: center;
```
TarasHoliuk, line 28: remove it - DRY (you've already set it in `.stars__star`)

TarasHoliuk (changes requested):
I've added 1 small suggestion

src/index.html
TarasHoliuk: it's better to do it with 'nth-child' as you do :+1:

Pavlo-Petrashevskyi: Thanks)

src/style.css
Pavlo-Petrashevskyi: Done)

SanyaBratashchuk (approved):

PR #2298, opened 2022-12-20
Flexbox / Stars

valntyn:
[DEMO LINK](https://valntyn.github.io/layout_stars/)
[TEST REPORT LINK](https://valntyn.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,25 @@
+body {
+  margin: 0;
+}
+
+.stars {
+  margin: 0;
+  display: flex;
+}
+
+.star_box {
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,25 @@
+body {
+  margin: 0;
+}
+
+.stars {
+  margin: 0;
+  display: flex;
+}
+
+.star_box {
```
witflash, line 10: If you know about BEM methodology here is better to name an element as `.star__box`

witflash (approved):
Good work!

PR #2304, opened 2022-12-22
Task has been done

u9on:
[DEMO LINK](https://u9on.github.io/layout_stars/)
[TEST REPORT LINK](https://u9on.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,30 @@
+body {
+  margin: 0;
+}
+
+.stars {
+  display: flex;
+  justify-content: center;
+  align-items: center;
+}
+
+.stars__star {
+  background-image: url(./images/star.svg);
+  width: 16px;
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,30 @@
...
+.stars__star {
+  background-image: url(./images/star.svg);
+  width: 16px;
+  height: 16px;
+  margin-right: 4px;
+  background-repeat: no-repeat;
+  background-position: center;
+}
+
+.stars__star :last-child {
```
witflash, line 20: I'd like to use pseudo-classes without space between their parent classes. Like in these docs: https://developer.mozilla.org/en-US/docs/Web/CSS/:last-child

witflash (approved):
Excellent work! 👍🏼

PR #2314, opened 2022-12-26
Develop

mxwolfy:
- [DEMO LINK](https://mxwolfy.github.io/layout_stars/)
- [TEST REPORT LINK](https://mxwolfy.github.io/layout_stars/report/html_report/)


src/style.css
```diff
@@ -0,0 +1,40 @@
...
+  background-image: url(images/star-active.svg);
+}
+
+.stars--4 .star:nth-child(-n + 4) {
+  background-image: url(images/star-active.svg);
+}
+
+.stars--5 .star:nth-child(-n + 5) {
+  background-image: url(images/star-active.svg);
+}
```
Virass, line 40: Cool, but you can unit this logic into one block

Virass (changes requested):

src/style.css
mxwolfy: Thanks, ok  i understood

Viktor-Kost (approved):

PR #2318, opened 2022-12-28
added stars

vshepel:
[DEMO LINK](https://vshepel.github.io/layout_stars/)
[TEST REPORT LINK](https://vshepel.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,23 @@
+* {
+  margin: 0;
+  padding: 0;
+}
+
+.stars {
+  display: flex;
+}
+
+.stars__star {
+  width: 16px;
+  height: 16px;
+  margin-right: 4px;
+  background: url(images/star-active.svg) no-repeat 50% 50%;
+}
+
+.stars--0 > *:nth-child(n+1),
+.stars--1 > *:nth-child(n+2),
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,23 @@
+* {
+  margin: 0;
+  padding: 0;
```
TarasHoliuk, line 3: don't add these properties to `*` selector - reset it for `body` (and/or smth else) if you need

src/style.css
```diff
@@ -0,0 +1,23 @@
...
+
+.stars {
+  display: flex;
+}
+
+.stars__star {
+  width: 16px;
+  height: 16px;
+  margin-right: 4px;
+  background: url(images/star-active.svg) no-repeat 50% 50%;
```
TarasHoliuk, line 14:
```suggestion
  background: url(images/star-active.svg) no-repeat center;
```

TarasHoliuk (changes requested):

src/style.css
vshepel: code reduction?

witflash (approved):
Good job! I like it! 👍🏼

PR #2362, opened 2023-01-12
Solution stars

Zelinskyi-Serhii:
https://Zelinskyi-Serhii.github.io/layout_stars/
https://Zelinskyi-Serhii.github.io/layout_stars/report/html_report/

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,35 @@
+* {
+  box-sizing: border-box;
+  margin: 0;
+  padding: 0;
+}
+
+.container {
+  display: flex;
+  flex-direction: column;
+  justify-content: flex-start;
+  width: 96;
+  height: 96px;
+}
+
+.stars {
+  display: flex;
+  flex-direction: row;
+}
+
+.stars_star {
+  width: 16px;
+  height: 16px;
+  border-radius: 0.5px;
+  background-image: url(images/star.svg);
+  background-repeat: no-repeat;
+  background-position: center;
+}
+
+.stars_star:not(:last-child) {
+  margin-right: 4px;
+}
+
+.active_star {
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,37 @@
+* {
+  box-sizing: border-box;
+  margin: 0;
+  padding: 0;
```
OleksiiBlezniuk, line 4: Don't use `*` selector for reseting styles, only for `box-sizing`. Reset default styles of particular tag if needed

src/style.css
```diff
@@ -0,0 +1,37 @@
...
+}
+
+.container {
+  display: flex;
+  flex-direction: column;
+  width: 96;
```
OleksiiBlezniuk, line 10: This value is incorrect, you can't just set number, need to specify unit (px in our case)

src/style.css
```diff
@@ -0,0 +1,37 @@
...
+  height: 96px;
```
OleksiiBlezniuk, line 11: And you don't need this rules, your container already has this size because of its children

src/style.css
```diff
@@ -0,0 +1,37 @@
...
+}
+
+.stars {
+  display: flex;
+}
+
+.stars__star {
+  width: 16px;
+  height: 16px;
+  border-radius: 0.5px;
```
OleksiiBlezniuk, line 21:
```suggestion
  border-radius: 1px;
```

OleksiiBlezniuk: Don't use values smaller that 1px

OleksiiBlezniuk (changes requested):

OleksiiBlezniuk (approved):
Well done! ⭐

PR #2367, opened 2023-01-13
Star Boxes

Vlados3611:
- [DEMO LINK](https://Vlados3611.github.io/layout_stars/)
- [TEST REPORT LINK](https://Vlados3611.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,61 @@
+* {
+  margin: 0;
+}
+
+.star {
+  display: flex;
+  flex-direction: row;
+}
+
+.stars__star {
+  width: 16px;
+  height: 16px;
+  background: url(./images/star.svg);
+  background-repeat: no-repeat;
+  margin-right: 4px;
+  background-position: 50%;
+}
+
+.star > .stars__star:last-child {
+  margin-right: 0;
+}
+
+.stars--1 > .stars__star:nth-of-type(1) {
+  background: url(./images/star-active.svg);
+  background-repeat: no-repeat;
+  background-position: 50%;
+}
+
+.stars--2 > .stars__star:nth-of-type(1),
+.stars--2 > .stars__star:nth-of-type(2) {
+  background: url(./images/star-active.svg);
+  background-repeat: no-repeat;
+  background-position: 50%;
+}
+
+.stars--3 > .stars__star:nth-of-type(1),
+.stars--3 > .stars__star:nth-of-type(2),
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,61 @@
+* {
+  margin: 0;
+}
```
OleksiiBlezniuk, line 3: Don't use `*` selector for default styles reset

src/style.css
```diff
@@ -0,0 +1,61 @@
...
+  height: 16px;
+  background: url(./images/star.svg);
+  background-repeat: no-repeat;
+  margin-right: 4px;
+  background-position: 50%;
+}
+
+.star > .stars__star:last-child {
+  margin-right: 0;
+}
```
OleksiiBlezniuk, line 21: Don't need to increase specificity here, just use pseudoclass

src/style.css
```diff
@@ -0,0 +1,61 @@
...
+
+.stars--1 > .stars__star:nth-of-type(1) {
+  background: url(./images/star-active.svg);
+  background-repeat: no-repeat;
+  background-position: 50%;
```
OleksiiBlezniuk, line 26: But these styles are already set to `stars__star`, no need to duplicate

src/style.css
```diff
@@ -0,0 +1,61 @@
...
+
+.stars--5 > .stars__star:nth-of-type(1),
+.stars--5 > .stars__star:nth-of-type(2),
+.stars--5 > .stars__star:nth-of-type(3),
+.stars--5 > .stars__star:nth-of-type(4),
+.stars--5 > .stars__star:nth-of-type(5) {
+  background: url(./images/star-active.svg);
+  background-repeat: no-repeat;
+  background-position: 50%;
+}
```
OleksiiBlezniuk, line 61: Actually this code can be shortened to 7 lines. Let's use correct condition in `nth-of-type` and combine rules with same properties to one

OleksiiBlezniuk (changes requested):

src/style.css
```diff
@@ -0,0 +1,28 @@
...
+
+.star {
+  display: flex;
+}
+
+.stars__star {
+  width: 16px;
+  height: 16px;
+  background-image: url(./images/star.svg);
+  background-position: 50%;
```
illiarerun, line 13: `center` would be much more clear

illiarerun (approved):

PR #2374, opened 2023-01-16
stars was added

Vo7kov:
https://vo7kov.github.io/layout_stars/
https://vo7kov.github.io/layout_stars/report/html_report/
Can you explain me please, what should I do? If I set the `background-position: center;` the test, which checks all stars is correct, but all another tests are incorrect, but, if I set the `background-position: center left;` there are no tests, which are correct, but when I'm looking to any test except complete test, I see that there is no difference. Explain me please, what should I do?

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,65 @@
+body {
+  margin: 0;
+  padding: 0;
+}
+
+.stars {
+  display: flex;
+  margin-right: 4px;
+}
+
+.stars__star {
+  background-image: url(./images/star.svg);
+  background-repeat: no-repeat;
+  background-position: center;
+  width: 16px;
+  height: 16px;
+}
+
+.stars__star:nth-child(n+1) {
+  margin-right: 4px;
+}
+
+.stars__star:last-child {
+  margin: 0;
+}
+
+.stars--1 .stars__star:first-child {
+  background-image: url(./images/star-active.svg);
+  background-repeat: no-repeat;
+  background-position: center;
+  width: 16px;
+  height: 16px;
+}
+
+.stars--2 .stars__star:nth-child(-n + 2) {
+  background-image: url(./images/star-active.svg);
+  background-repeat: no-repeat;
+  background-position: center;
+  width: 16px;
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,65 @@
...
+.stars__star:nth-child(n+1) {
+  margin-right: 4px;
+}
+
+.stars__star:last-child {
+  margin: 0;
+}
+
+.stars--1 .stars__star:first-child {
+  background-image: url(./images/star-active.svg);
```
mykhalenych, line 28: use only background-image delete everything else

src/style.css
```diff
@@ -0,0 +1,65 @@
...
+
+.stars__star:nth-child(n+1) {
+  margin-right: 4px;
+}
+
+.stars__star:last-child {
+  margin: 0;
+}
+
+.stars--1 .stars__star:first-child {
```
mykhalenych, line 27:
```suggestion
.stars--1 :first-child,
.stars--2 :nth-child(-n+2),
.stars--3 > div:nth-child(-n + 3),
.stars--4 > div:nth-child(-n + 4),
.stars--5 > div:nth-child(n) {

}

```

mykhalenych (changes requested):

Vo7kov:
https://vo7kov.github.io/layout_stars/
https://vo7kov.github.io/layout_stars/report/html_report/

src/style.css
```diff
@@ -0,0 +1,33 @@
+body {
+  margin: 0;
+  padding: 0;
+}
+
+.stars {
+  display: flex;
+  margin-right: 4px;
```
witflash, line 8: You don't need this rule.
Remove it and your tests will pass, I think

witflash (changes requested):
Pay attention to my comment, please

Vo7kov:
https://vo7kov.github.io/layout_stars/
https://vo7kov.github.io/layout_stars/report/html_report/

mykhalenych (approved):

PR #2390, opened 2023-01-21
stars-block was added

eugene-kulish:


eugene-kulish:
https://eugene-kulish.github.io/layout_stars/
https://eugene-kulish.github.io/layout_stars/report/html_report/

src/style.css
```diff
@@ -0,0 +1,32 @@
...
+.stars {
+  display: flex;
+}
+
+.stars__star:not(:last-child) {
+  margin-right: 4px;
+}
+
+.stars--5 .stars__star,
+.stars--4 .stars__star:not(:last-child),
```
TarasHoliuk, line 27: better to be consistent and write (-n + 4) instead of `last-child`
such a way of writing will allow you to use scss loops in future tasks (FYI)

TarasHoliuk (approved):

PR #2393, opened 2023-01-22
Stars were done

NataliaSavenets:
**[DEMO LINK](https://NataliaSavenets.github.io/layout_stars/)
[TEST REPORT LINK](https://NataliaSavenets.github.io/layout_stars/report/html_report/)**


src/style.css
```diff
@@ -0,0 +1,25 @@
...
+.stars {
+  display: flex;
+}
+
+.stars__star {
+  background-image: url(./images/star-active.svg);
+  width: 16px;
+  height: 16px;
+  margin-right: 4px;
+  color: #e5e5e5;
```
OleksiiBlezniuk, line 14: Why do you need color for image? 🙂 Let's remove

src/style.css
```diff
@@ -0,0 +1,25 @@
...
+  background-repeat: no-repeat;
+  background-position: center;
+}
```
OleksiiBlezniuk, line 17: What about last element? Better to reset margins for last elements

OleksiiBlezniuk: And it will be more logical to set this star inactive by default, but it's not critical

OleksiiBlezniuk (changes requested):

illiarerun (approved):

PR #2394, opened 2023-01-22
create stars

BilokonYaroslav:
- [DEMO LINK](https://BilokonYaroslav.github.io/layout_stars/)
- [TEST REPORT LINK](https://BilokonYaroslav.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -8,49 +8,47 @@
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <div class="counteiner">
-      <div class="stars stars--0">
-        <div class="stars__star"></div>
-        <div class="stars__star"></div>
-        <div class="stars__star"></div>
-        <div class="stars__star"></div>
-        <div class="stars__star"></div>
-      </div>
-      <div class="stars stars--1">
-        <div class="stars__star"></div>
-        <div class="stars__star"></div>
-        <div class="stars__star"></div>
-        <div class="stars__star"></div>
-        <div class="stars__star"></div>
-      </div>
-      <div class="stars stars--2">
-        <div class="stars__star"></div>
-        <div class="stars__star"></div>
-        <div class="stars__star"></div>
-        <div class="stars__star"></div>
-        <div class="stars__star"></div>
-      </div>
-      <div class="stars stars--3">
-        <div class="stars__star"></div>
-        <div class="stars__star"></div>
-        <div class="stars__star"></div>
-        <div class="stars__star"></div>
-        <div class="stars__star"></div>
-      </div>
-      <div class="stars stars--4">
-        <div class="stars__star"></div>
-        <div class="stars__star"></div>
-        <div class="stars__star"></div>
-        <div class="stars__star"></div>
-        <div class="stars__star"></div>
-      </div>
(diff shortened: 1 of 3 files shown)


src/index.html
```diff
@@ -8,6 +8,49 @@
...
+        <div class="stars__star"></div>
+        <div class="stars__star"></div>
+      </div>
+      <div class="stars stars--1">
+        <div class="stars__star active"></div>
+        <div class="stars__star"></div>
+        <div class="stars__star"></div>
+        <div class="stars__star"></div>
+        <div class="stars__star"></div>
+      </div>
```
Virass, line 25: Change logic for set active stars amount. Set active stars use only second class for parrent, like stars--1... stars--3... etc...
And write css like 
`.stars--1 .stars__star:nth-child(-n+1),
	... continue this logic
	`

Virass (changes requested):

src/style.css
```diff
@@ -0,0 +1,40 @@
...
+}
+
+.stars__star {
+  height: 16px;
+  width: 16px;
+  background: url(images/star.svg) no-repeat center;
+  margin-right: 4px;
+}
+
+.stars--1 .stars__star:nth-child(-n+1) {
```
mykhalenych, line 22:
```suggestion
.stars--1 .stars__star:nth-child(-n+1),
.stars--2 .stars__star:nth-child(-n+2),
.stars--3 .stars__star:nth-child(-n+3),
... {
  background: url(images/star-active.svg) no-repeat center;
  }
```

src/style.css
```diff
@@ -0,0 +1,40 @@
...
+.stars {
+  height: 16px;
+  display: flex;
+}
+
+.stars__star {
+  height: 16px;
+  width: 16px;
+  background: url(images/star.svg) no-repeat center;
+  margin-right: 4px;
```
mykhalenych, line 19: last star shouldn't has margin-right

src/style.css
```diff
@@ -0,0 +1,40 @@
+* {
+  margin: 0;
+  padding: 0;
+}
+
+.counteiner {
+  display: inline-block;
```
mykhalenych, line 7: it is redundant

mykhalenych (changes requested):

src/style.css
BilokonYaroslav: without without margin-right the test does not pass

margin-right
margin-right2

VitaliyBondarenko1982: Not fixed.
Remove div with class counteiner and set last star margin-right: 0.
If you will do both fixes tests should pass

VitaliyBondarenko1982 (changes requested):

src/style.css
```diff
@@ -0,0 +1,28 @@
+* {
+  margin: 0;
+  padding: 0;
+}
```
polosanya, line 4: Don't reset margins and paddings using (*) selector. It applies to all elements, which is not good for perfomance.
Better specify the exact elements

polosanya (approved):
Good job! I'm approving it, but please pay attention to my comment)

PR #2435, opened 2023-01-26
Develop

vadosik15:
- [DEMO LINK](https://vadosik15github.io/layout_stars/)
- [TEST REPORT LINK](https://vadosik15.github.io/layout_stars/report/html_report/)


src/style.css
```diff
@@ -0,0 +1,33 @@
...
+  margin: 0;
+}
+
+.stars__star {
+  height: 16px;
+  width: 16px;
+  margin-right: 4px;
+  background: url(./images/star.svg);
+  background-position: center !important;
+  background-repeat: no-repeat !important;
```
Andr1yK, line 15: wow... why do u need `important`?

src/style.css
```diff
@@ -0,0 +1,33 @@
...
+  display: flex;
+}
+
+.stars--1 > .stars__star:first-child,
+.stars--2 > .stars__star:nth-child(-n + 2),
+.stars--3 > .stars__star:nth-child(-n + 3),
+.stars--4 > .stars__star:nth-child(-n + 4),
+.stars--5 > .stars__star:nth-child(-n + 5) {
+  background: url(./images/star-active.svg);
+  background-position: 50%;
```
Andr1yK, line 32: already added in .stars__star

Andr1yK (changes requested):

src/style.css
vadosik15: image

i do not know why but (background-position) is not working for active stars

vadosik15: sory i find mistake me should to use background-image not background

PR #2436, opened 2023-01-26
Develop

yuliia-vereshko:
- [DEMO LINK](https://yuliia-vereshko>.github.io/layout_stars/)
- [TEST REPORT LINK](https://yuliia-vereshko.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,36 @@
+* {
+  box-sizing: border-box;
+}
+
+body {
+  margin: 0;
+}
+
+.stars {
+  display: flex;
+  margin: 0;
+  padding: 0;
+  justify-content: flex-start;
+
+}
+
+.stars__star {
+  width: 16px;
+  height: 16px;
+  margin-right: 4px;
+  background-image: url(./images/star.svg);
+  background-repeat: no-repeat;
+  background-position: center;
+}
+
+.stars--1 :nth-child(1),
+.stars--2 :nth-child(-n+2),
+.stars--3 :nth-child(-n+3),
+.stars--4 :nth-child(-n+4),
+.stars--5 :nth-child(-n+5) {
+  background-image: url(images/star-active.svg);
+}
+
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,36 @@
...
+  margin: 0;
+}
+
+.stars {
+  display: flex;
+  margin: 0;
+  padding: 0;
+  justify-content: flex-start;
+
+}
```
OleksiiBlezniuk, line 15:
```suggestion
.stars {
  display: flex;
  justify-content: flex-start;
}
```

OleksiiBlezniuk: divs don't have initial margins or paddings, so it can be removed

src/style.css
```diff
@@ -0,0 +1,36 @@
...
+.stars--2 :nth-child(-n+2),
+.stars--3 :nth-child(-n+3),
+.stars--4 :nth-child(-n+4),
+.stars--5 :nth-child(-n+5) {
+  background-image: url(images/star-active.svg);
+}
+
+.stars__star:last-child {
+  margin-right: 0;
+}
```
OleksiiBlezniuk, line 36:
```suggestion
.stars__star {
  width: 16px;
  height: 16px;
  margin-right: 4px;
  background-image: url(./images/star.svg);
  background-repeat: no-repeat;
  background-position: center;
}

.stars__star:last-child {
  margin-right: 0;
}

.stars--1 :nth-child(1),
.stars--2 :nth-child(-n+2),
.stars--3 :nth-child(-n+3),
.stars--4 :nth-child(-n+4),
.stars--5 :nth-child(-n+5) {
  background-image: url(images/star-active.svg);
}
```

OleksiiBlezniuk: that way is more clear as for me

OleksiiBlezniuk (approved):
GJ, but avoid setting unnecessary properties to classes

PR #2440, opened 2023-01-26
add solution - layout_stars

Vepr984:
[DEMO LINK](https://vepr984.github.io/layout_stars/)
[TEST REPORT LINK](https://vepr984.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,38 @@
+/* * {
+  box-sizing: border-box;
+} */
+
+body {
+  margin: 0;
+  padding: 0;
+}
+
+ul {
+  list-style: none;
+  padding: 0;
+}
+
+.stars {
+  display: flex;
+  margin: 0;
+}
+
+.stars :not(:last-child) {
+  margin-right: 4px;
+}
+
+.stars__star {
+  width: 16px;
+  height: 16px;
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,38 @@
+/* * {
```
mykhalenych, line 1: delete comment

src/style.css
```diff
@@ -0,0 +1,38 @@
...
+  box-sizing: border-box;
+} */
+
+body {
+  margin: 0;
+  padding: 0;
+}
+
+ul {
+  list-style: none;
```
mykhalenych, line 11: maybe better to use div?

mykhalenych (changes requested):

src/style.css
Vepr984: good note, thanks

src/style.css
```diff
@@ -0,0 +1,29 @@
+body {
+  margin: 0;
+  padding: 0;
+}
+
+.stars {
+  display: flex;
+  margin: 0;
+}
```
IvanFesenko, line 9: Useless margin

IvanFesenko (approved):
Good job.

PR #2445, opened 2023-01-27
try to push

BozhenaBodnar:
- [DEMO LINK](https://BozhenaBodnar.github.io/layout_stars/)
- [TEST REPORT LINK](https://BozhenaBodnar.github.io/layout_stars/report/html_report/)


src/style.css
```diff
@@ -0,0 +1,43 @@
...
+  background-repeat: no-repeat;
+}
+
+.stars--1 .stars__star:first-child {
+  background-image: url("images/star-active.svg");
+}
+
+.stars--2 .stars__star:nth-child(-n+2) {
+  background-image: url("images/star-active.svg");
+}
```
Andr1yK, line 31: let's merge same styles for classes 

```suggestion
.stars--1 .stars__star:first-child,
.stars--2 .stars__star:nth-child(-n+2) {
  background-image: url("images/star-active.svg");
}
```

Andr1yK (changes requested):
Why we need this folder in this PR?
image

src/style.css
BozhenaBodnar: fixed:)

Andr1yK: But you didn't add any commits....🥲

image

PR #2452, opened 2023-01-29
stars add

Vitalii-Fediaiev:
[DEMO LINK](https://vitalii-fediaiev.github.io/layout_stars/)
[TEST REPORT LINK](https://vitalii-fediaiev.github.io/layout_stars/report/html_report/)


src/style.css
```diff
@@ -0,0 +1,29 @@
+* {
+  box-sizing: border-box;
+  padding: 0;
+  margin: 0;
+}
```
polosanya, line 5:
```suggestion
* {
  box-sizing: border-box;
}
```
Don't reset the margins and paddings using (*) selector. It applies to all elements, which is really bad for performance. Better to specify the exact elements

polosanya (changes requested):
Good job! I'm approving it, but please pay attention to my comment

Anastasiia-Svintsova (approved):

PR #2455, opened 2023-01-30
Stars layout completed

ArthurKrepko:
- [DEMO LINK](https://ArthurKrepko.github.io/layout_stars/)
- [TEST REPORT LINK](https://ArthurKrepko.github.io/layout_stars/report/html_report/)


src/index.html
```diff
@@ -8,6 +8,131 @@
...
   <body>
-    <h1>Stars</h1>
+    <div class="stars stars--0">
+      <div class="star__container">
+        <div class="stars__star"></div>
+      </div>
+
+      <div class="star__container">
+        <div class="stars__star"></div>
+      </div>
```
Andr1yK, line 18: why do we need those star__containers?

src/style.css
```diff
@@ -0,0 +1,36 @@
+* {
+  margin: 0;
```
Andr1yK, line 2: https://mateacademy.slack.com/archives/C04EZF03G10/p1674665677527709

Andr1yK (changes requested):

src/index.html
ArthurKrepko: because of positioning of bg-image. Image of star doesn't look like in Figma Example. In example it's width = 16px and height = 16px, but in default size width = 13px and height = 12px

illiarerun: You can use `background-position` for it, remove these containers

illiarerun (changes requested):

illiarerun (approved):

PR #2458, opened 2023-01-30
Ulcja.Stars

Ulcja:
- [DEMO LINK](https://Ulcja.github.io/layout_stars/)
- [TEST REPORT LINK](https://Ulcja.github.io/layout_stars/report/html_report/)


src/style.css
```diff
@@ -0,0 +1,29 @@
+* {
+  margin: 0;
+  padding: 0;
+}
```
DarkMistyRoom, line 4: Don't use * selector for zeroing out your margins or paddings. It's still inefficient for browser to read your web document

DarkMistyRoom (changes requested):

tetiana-ivanchenko (approved):

PR #2461, opened 2023-01-31
Added Stars

Kapidastra:
[DEMO LINK](https://kapidastra.github.io/layout_stars/)
[TEST REPORT LINK](https://kapidastra.github.io/layout_stars/report/html_report/)


Volodymyr-Melnyk-Fron10der (changes requested):
Hi! Pay more attention for the task. The count of active starts has to be controlled by modificator `stars--x`, not by `.block` class name

image

src/style.css
```diff
@@ -0,0 +1,31 @@
...
+  height: 16px;
+  width: 16px;
+  margin-right: 4px;
+  background-repeat: no-repeat;
+  background-position: center;
+}
+
+.stars {
+  display: flex;
+  flex-direction: row;
```
tetiana-ivanchenko, line 20: it's default value

tetiana-ivanchenko (changes requested):

src/style.css
Kapidastra: Ok, fixed it.

SanyaBratashchuk (approved):

PR #2474, opened 2023-02-04
create stars

semchuk-rodion:
https://semchuk-rodion.github.io/layout_stars/
https://semchuk-rodion.github.io/layout_stars/report/html_report/


src/style.css
```diff
@@ -0,0 +1,25 @@
...
+
+.stars {
+  display: flex;
+}
+
+.stars__star {
+  background-image: url(./images/star.svg);
+  height: 16px;
+  width: 16px;
+  border-radius: 0.5px;
```
ivanchuk-yurii, line 13: it's redundant

src/style.css
```diff
@@ -0,0 +1,25 @@
+body {
+  margin: 0;
+}
+
+.stars {
+  display: flex;
+}
+
+.stars__star {
+  background-image: url(./images/star.svg);
```
ivanchuk-yurii, line 10: it's better to combine all background properties into one `background`

ivanchuk-yurii (approved):

PR #2478, opened 2023-02-05
stars

miamate:
- [DEMO LINK](https://miamate.github.io/layout_stars/)
- [TEST REPORT LINK](https://miamate.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,53 @@
+body {
+  margin: 0;
+}
+
+.rating {
+  display: flex;
+  flex-direction: column;
+}
+
+.stars {
+  display: flex;
+  flex-direction: row;
+}
+
+.stars__star {
+  display: flex;
+  margin-right: 4px;
+  height: 16px;
+  width: 16px;
+  background-image: url(./images/star.svg);
+  background-repeat: no-repeat;
+  background-position: center;
+}
+
+.stars--1 .stars__star:nth-child(-n+1) {
+  background-image: url(./images/star-active.svg);
+  background-repeat: no-repeat;
+  background-position: center;
+}
+
+.stars--2 .stars__star:nth-child(-n+2) {
+  background-image: url(./images/star-active.svg);
+  background-repeat: no-repeat;
+  background-position: center;
+}
+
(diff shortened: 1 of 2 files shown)


src/style.css
```diff
@@ -0,0 +1,53 @@
...
+  background-image: url(./images/star-active.svg);
+  background-repeat: no-repeat;
+  background-position: center;
+}
+
+.stars--5 .stars__star:nth-child(-n+5) {
+  background-image: url(./images/star-active.svg);
+  background-repeat: no-repeat;
+  background-position: center;
+}
```
agniya-i, line 53: combine these selectors to not repeat code

src/style.css
```diff
@@ -0,0 +1,53 @@
...
+  width: 16px;
+  background-image: url(./images/star.svg);
+  background-repeat: no-repeat;
+  background-position: center;
+}
+
+.stars--1 .stars__star:nth-child(-n+1) {
+  background-image: url(./images/star-active.svg);
+  background-repeat: no-repeat;
+  background-position: center;
```
agniya-i, line 28:
```suggestion
```

you can remove `background-repeat` and `background-position`, because it's already defined in `.stars__star`

agniya-i (changes requested):

src/style.css
miamate: you're right! looks definitely better without duplicates and still works :)

src/style.css
```diff
@@ -0,0 +1,43 @@
...
+.stars--3 .stars__star:nth-child(-n+3) {
+  background-image: url(./images/star-active.svg);
+}
+
+.stars--4 .stars__star:nth-child(-n+4) {
+  background-image: url(./images/star-active.svg);
+}
+
+.stars--5 .stars__star:nth-child(-n+5) {
+  background-image: url(./images/star-active.svg);
```
polosanya, line 42: image

polosanya (changes requested):
Almost done!

polosanya (approved):
Well done!

PR #2484, opened 2023-02-06
commit message

VasilPan:
- [DEMO LINK](https://VasilPan.github.io/layout_stars/)
- [TEST REPORT LINK](https://VasilPan.github.io/layout_stars/report/html_report/)


src/style.css
```diff
@@ -0,0 +1,34 @@
+* {
+  margin: 0;
```
loralevitska, line 2: It is advisable to use only `box-sizing` for this selector. Move it `body`

src/style.css
```diff
@@ -0,0 +1,34 @@
...
+  background-position: center;
+  margin-right: 4px;
+}
+
+.stars--0,
+.stars--1,
+.stars--2,
+.stars--3,
+.stars--4,
+.stars--5 {
```
loralevitska, line 23: Combine with  1 common class

src/index.html
```diff
@@ -8,6 +8,50 @@
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Stars</h1>
+    <div class="stars">
+      <div class="stars--0">
+        <div class="stars_star"></div>
```
loralevitska, line 13: Fix everywhere
```suggestion
        <div class="stars__star"></div>
```

src/index.html
```diff
@@ -8,6 +8,50 @@
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Stars</h1>
+    <div class="stars">
```
loralevitska, line 11: BTW you don't need wrapper here. You may move this class to each block with modifier `stars--{number}`

loralevitska (changes requested):
A little more and it will be completed

src/index.html
loralevitska: 1. Okay, maybe I didn't explain it that way. It is better to remove that wrapper, it is not needed, all the blocks will be under each other anyway.
2. According to BEM, it is not very readable when the block modifier is embedded in other sub-blocks. It is better to use the star class instead of the row class

src/index.html
```diff
@@ -8,6 +8,50 @@
...
+      <div class="stars--0 row">
```
loralevitska, line 12:
```suggestion
      <div class="star stars--0">
```

loralevitska (approved):

PR #2485, opened 2023-02-06
Develop

maks-golovatyuk:
https://maks-golovatyuk.github.io/layout_stars/
https://maks-golovatyuk.github.io/layout_stars/report/html_report/


polosanya (approved):
Good job!

PR #2486, opened 2023-02-07
stars first try

julius0724:
[DEMO LINK](https://github.com/julius0724/layout_stars)
[TEST REPORT LINK](https://github.com/julius0724/layout_stars/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -8,6 +8,49 @@
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Stars</h1>
+    <div class="rate">
+      <div class="stars stars--0">
+        <div class="stars__star"></div>
+        <div class="stars__star"></div>
+        <div class="stars__star"></div>
+        <div class="stars__star"></div>
+        <div class="stars__star"></div>
+      </div>
+      <div class="stars stars--1">
+        <div class="stars__star"></div>
+        <div class="stars__star"></div>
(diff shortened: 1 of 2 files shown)


src/index.html
```diff
@@ -8,6 +8,49 @@
...
+    <div class="rate">
+      <div class="stars stars--0">
+        <div class="stars__star"></div>
+        <div class="stars__star"></div>
+        <div class="stars__star"></div>
+        <div class="stars__star"></div>
+        <div class="stars__star"></div>
+      </div>
+      <div class="stars stars--1">
+        <div class="stars__star"></div>
```
OleziO, line 20: It is better to name the element` stars__item`, it will be more correct. Also, it is better to list elements of the same type (`<ul>`)

src/style.css
```diff
@@ -0,0 +1,25 @@
...
+  margin-right: 4px;
+  display: flex;
+  width: 16px;
+  height: 16px;
+  background-image: url(./images/star.svg);
+  background-repeat: no-repeat;
+  background-position: center;
+}
+
+.stars--1 .stars__star:nth-child(1),
```
OleziO, line 19: Better to use `(-n+1)` instead of `1`.

OleziO (approved):

OleziO (approved):

AlisaVlasova (approved):

PR #2491, opened 2023-02-07
Solution

superzuc:
https://superzuc.github.io/layout_stars/

https://superzuc.github.io/layout_stars/report/html_report/


src/index.html
```diff
@@ -8,6 +8,111 @@
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Stars</h1>
+    <div class="big_star_container">
+      <div class="star_container stars--0">
```
polosanya, line 12:
```suggestion
      <div class="stars stars--0">
```
image

src/index.html
```diff
@@ -8,6 +8,111 @@
...
+        </div>
+        <div class="star_deactivated">
+
+        </div>
+        <div class="star_deactivated">
+
+        </div>
+        <div class="star_deactivated">
+
+        </div>
```
polosanya, line 27:
```suggestion
        <div class="stars__star"></div>
        <div class="stars__star"></div>
        <div class="stars__star"></div>
        <div class="stars__star"></div>
        <div class="stars__star"></div>
```
Fix all cases
image

polosanya (changes requested):
I don't know how tests passed, but it shouldn't look like this)
image
Fix it according to the example:
image
Also, you don't need to hardcore the classes for activated\deactivated stars. Just use[ :nth-child](https://developer.mozilla.org/en-US/docs/Web/CSS/:nth-child) selector for stars container

superzuc:
I really don't know why the DEMO LINK appears in this way, as you can see the width of the container is less than 100px but it doesn't run as i see in my project, i think that i've solved all the problems, sorry for that

src/index.html
```diff
@@ -8,6 +8,56 @@
...
+      <div class="stars_star"></div>
+      <div class="stars_star"></div>
+      <div class="stars_star"></div>
+      <div class="stars_star"></div>
+    </div>
+
+      </div>
+
+
   </body>
```
polosanya, line 62:
```suggestion
    </div>
  </body>
```
What's the purpose of this closing tag?

src/index.html
```diff
@@ -8,6 +8,56 @@
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Stars</h1>
+    <div class="star_container stars--0">
+      <div class="stars_star"></div>
```
polosanya, line 12:
```suggestion
    <div class="stars stars--0">
      <div class="stars__star"></div>
```
Add a default class of stars without modifier too.
And provide two underlines, so it's clear that it is an element of stars block

src/style.css
```diff
@@ -0,0 +1,64 @@
...
+  width: 16px;
+  height: 16px;
+}
+
+.stars--1 .stars_star:nth-child(-n + 1) {
+  background-image: url(images/star-active.svg);
+  background-position: center center;
+  background-repeat: no-repeat;
+  width: 16px;
+  height: 16px;
```
polosanya, line 31: You can change only background-image.
No need to duplicate other properties

src/style.css
```diff
@@ -0,0 +1,64 @@
...
+
+.stars--4 .stars_star:nth-child(-n + 4) {
+  background-image: url(images/star-active.svg);
+  background-position: center center;
+  background-repeat: no-repeat;
+  width: 16px;
+  height: 16px;
+}
+
+.stars--5 .stars_star:nth-child(-n + 5) {
```
polosanya, line 58: image

polosanya (changes requested):
Almost done!

src/style.css
```diff
@@ -0,0 +1,28 @@
...
+.stars {
+  display: flex;
+  flex-direction: row;
+  justify-content: space-between;
+  align-items: center;
+  padding-right: 4px;
+  width: 96px;
+}
+
+.stars_star {
```
Volodymyr-Melnyk-Fron10der, line 14: according to the BEM
```suggestion
.stars__star {
```

Volodymyr-Melnyk-Fron10der (changes requested):
Almost perfect, just one BEM error

Volodymyr-Melnyk-Fron10der (approved):
Looks good👍

polosanya (approved):

PR #2509, opened 2023-02-14
added stars

valeriy-mikheyev:
- [DEMO LINK](https://valeriy-mikheyev.github.io/layout_stars/)
- [TEST REPORT LINK](https://valeriy-mikheyev.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,27 @@
+* {
+  margin: 0;
+  padding: 0;
+}
+
+.stars {
+  display: flex;
+}
+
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,27 @@
+* {
+  margin: 0;
+  padding: 0;
```
vlad-demchuk, line 3: don't use `*` for reset

vlad-demchuk (approved):

PR #2517, opened 2023-02-16
+Stars

DefaultGitUser:
- [DEMO LINK](https://defaultgituser.github.io/layout_stars/)
- [TEST REPORT LINK](https://defaultgituser.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,28 @@
+* {
+  margin: 0;
+}
+
+.stars {
+  display: flex;
+}
+
+.stars .stars__star {
+  margin-right: 4px;
+  width: 16px;
+  height: 16px;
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,28 @@
+* {
+  margin: 0;
+}
+
+.stars {
+  display: flex;
+}
+
+.stars .stars__star {
+  margin-right: 4px;
```
sTorba24, line 10: write margins after `width` `height`

sTorba24 (approved):
Good work!

src/style.css
sTorba24: https://codeguide.academy/html-css.html#css-order

here you can read about ordering for properties

PR #2520, opened 2023-02-16
added stars

OlenaGor:
https://OlenaGor.github.io/layout_stars/
https://OlenaGor.github.io/layout_stars/report/html_report/


src/style.css
```diff
@@ -0,0 +1,41 @@
...
+  height: 16px;
+  width: 16px;
+  margin-right: 4px;
+}
+
+.star:last-child {
+  margin-right: 0;
+}
+
+.stars--1 :nth-child(1) {
```
mykhalenych, line 23:
```suggestion
.stars--1 :nth-child(1),
.stars--2 :nth-child(-n+2),
... {
  background-image: url(./images/star-active.svg);
}
```

src/style.css
```diff
@@ -0,0 +1,41 @@
+* {
+  margin: 0;
```
mykhalenych, line 2: reset margins only for specific tags not for all
`*` selector better to use only for box-sizing

mykhalenych (changes requested):

VitaliyBondarenko1982 (approved):

VitaliyBondarenko1982 (approved):

PR #2523, opened 2023-02-17
Develop

ShplmnVtlk:
https://ShplmnVtlk.github.io/layout_stars/

https://ShplmnVtlk.github.io/layout_stars/report/html_report/


OleziO (changes requested):
Empty demo page, fix it.

PR #2542, opened 2023-02-21
added stars

maksym-didyk:


src/style.css
```diff
@@ -0,0 +1,24 @@
...
+  display: flex;
+}
+
+.stars--1 :nth-child(-n + 1),
+.stars--2 :nth-child(-n + 2),
+.stars--3 :nth-child(-n + 3),
+.stars--4 :nth-child(-n + 4),
+.stars--5 :nth-child(-n + 5) {
+  background: no-repeat url(images/star-active.svg);
+  background-position: 50% 50%;
```
Olexandr6, line 15: You don't need the property here
```suggestion
```

Olexandr6, line 15: Better to use just center, it's the same
```suggestion
  background-position: center;
```

src/style.css
```diff
@@ -0,0 +1,24 @@
...
+}
+
+.stars__star {
+  height: 16px;
+  width: 16px;
+  background: no-repeat url(images/star.svg);
+  background-position: 50% 50%;
```
Olexandr6, line 22:
```suggestion
  background-position: center;
```

src/style.css
Olexandr6: Just move `.stars__star` selector above, and you don't need to repeat `background-position` here

src/style.css
```diff
@@ -0,0 +1,24 @@
...
+  margin-right: 4px;
```
Olexandr6, line 23: There shouldn't be margins after last start
Screenshot 2023-02-22 at 19 38 58

Olexandr6 (changes requested):
Almost done👍 Now, let's work on improvements:

src/style.css
maksym-didyk: Done

src/style.css
maksym-didyk: Done

src/style.css
maksym-didyk: Done
Added .stars__star:last-child

src/style.css
maksym-didyk: It doen't work without here

maksym-didyk:
I've have fixed bugs. 
Check it, pls.

Olexandr6 (approved):

PR #2594, opened 2023-02-23
stars are shining

stnayn:
- [DEMO LINK](https://stnayn.github.io/layout_stars/)
- [TEST REPORT LINK](https://stnayn.github.io/layout_stars/report/html_report/)


src/style.css
```diff
@@ -0,0 +1,26 @@
+* {
+  margin: 0;
+  padding: 0;
```
tetiana-ivanchenko, line 3: don't reset indentation for the *

tetiana-ivanchenko (changes requested):

src/style.css
stnayn: Can you explain to me why I cannot use * for resetting styles?

sTorba24: because * is selecting everything, it's bad practice to reset styles so

sTorba24 (approved):
Approved :) I left you a comment why to not use * selector for resetting

PR #2600, opened 2023-02-24
implemented rows of stars

Titech0:
[DEMO LINK](https://titech0.github.io/layout_stars/)
[TEST REPORT LINK](https://titech0.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,30 @@
+.body {
+  margin: 0;
+  --active-star: url(./images/star-active.svg);
+}
+
+.stars__star {
+  background-image: url(./images/star.svg);
+  background-repeat: no-repeat;
+  background-position: center center;
+
+  margin-right: 4px;
+  height: 16px;
+  width: 16px;
+}
+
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,30 @@
...
+  background-position: center center;
+
+  margin-right: 4px;
+  height: 16px;
+  width: 16px;
+}
+
+.stars__star:last-child {
+  margin-right: 0;
+}
```
etojeDenys, line 18:
```suggestion
.stars__star:not(:last-child) {
  margin-right: 4px;
}
```

etojeDenys (changes requested):

src/style.css
```diff
@@ -0,0 +1,29 @@
+.body {
+  margin: 0;
+  --active-star: url(./images/star-active.svg);
```
vlad-demchuk, line 3: move variable to `:root` pseudo selector

vlad-demchuk (approved):

PR #2605, opened 2023-02-25
1st commit

medolga:
- [DEMO LINK](https://medolga.github.io/layout_stars/)
- [TEST REPORT LINK](https://medolga.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,4 +1,4 @@
-* {
+body {
   margin: 0;
   padding: 0;
 }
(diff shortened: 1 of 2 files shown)


src/style.css
```diff
@@ -0,0 +1,27 @@
+* {
+  margin: 0;
+  padding: 0;
+}
```
TarasHoliuk, line 4: do not reset such values in `*` selector - use it for setting `box-sizing`
if you want to reset some default styles do it for a list of tags you really need, i. e. 
```
div,
body {
  margin: 0
}
```

TarasHoliuk (approved):
GJ :+1: 
I've added only one comment, just FYI

PR #2608, opened 2023-02-26
starsone

oleksandr-rudiuk:


src/style.css
```diff
@@ -0,0 +1,34 @@
...
+  flex-direction: column;
+}
+
+.stars--0,
+.stars--1,
+.stars--2,
+.stars--3,
+.stars--4,
+.stars--5 {
+  display: flex;
```
witflash, line 16: You don't need to add this rule for every modifier. You can just place it inside `.stars` selector if you need it everywhere.

src/index.html
```diff
@@ -8,6 +8,79 @@
...
+        </div>
+        <div class="box">
+        </div>
+        <div class="box">
+        </div>
+        <div class="box">
+        </div>
+      </div>
+      <div class="stars--1">
+        <div class="box">
```
witflash, line 25: This element should be `stars__star` not `box`.
image

witflash (changes requested):
Take attention to my comments, please

SanyaBratashchuk (approved):

PR #2620, opened 2023-02-28
task done

artem-ivashchenko:
[DEMO LINK] (https://artem-ivashchenko.github.io/layout_stars/)
[TEST REPORT LINK] (https://artem-ivashchenko.github.io/layout_stars/report/html_report/)

How could i implement stars better with nth-child property? Just curious :)


polosanya (approved):
Well done)

PR #2623, opened 2023-02-28
first commit

KhromchenkovOleksandr:


src/style.css
```diff
@@ -0,0 +1,26 @@
+* {
+  margin: 0;
+}
```
vlad-demchuk, line 3: don't reset margins for all elements
better to specify some elements you need to reset

src/style.css
```diff
@@ -0,0 +1,26 @@
...
+  background-repeat: no-repeat;
+}
+
+.star:not(:last-child) {
+  margin-right: 4px;
+}
+
+.star__active {
+  background-image: url(./images/star-active.svg);
+  /* width: 16px; */
```
vlad-demchuk, line 25: remove commnets

vlad-demchuk (changes requested):

src/style.css
KhromchenkovOleksandr: - Reset browser's default `margin`    I rad this in the task

src/style.css
```diff
@@ -0,0 +1,25 @@
...
+  display: flex;
+  background-position: center;
+  background-repeat: no-repeat;
+}
+
+.star:not(:last-child) {
+  margin-right: 4px;
+}
+
+.star__active {
```
OleziO, line 23: You don't need this class. You need to set stars from container class number with `:nth-child`

OleziO (changes requested):

src/style.css
```diff
@@ -0,0 +1,43 @@
...
+  background-image: url(./images/star-active.svg);
+}
+
+.stars--4 .star:nth-child(-n+4) {
+  background-image: url(./images/star-active.svg);
+}
+
+.stars--5 .star:nth-child(-n+5) {
+  background-image: url(./images/star-active.svg);
+}
```
IvanFesenko, line 43: You can use comma
```suggestion
.stars--0 .star:nth-child(-n),
.stars--1 .star:nth-child(-n+1),
.stars--2 .star:nth-child(-n+2),
.stars--3 .star:nth-child(-n+3),
.stars--4 .star:nth-child(-n+4),
.stars--5 .star:nth-child(-n+5) {
  background-image: url(./images/star-active.svg);
}
```

IvanFesenko (approved):
Check my comment to avoid mistakes in the future

src/style.css
KhromchenkovOleksandr: usefull

PR #2633, opened 2023-03-04
Layout Stars

vintwp:
https://vintwp.github.io/layout_stars/
https://vintwp.github.io/layout_stars/report/html_report/

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -4,7 +4,6 @@
 }
 
 .stars {
-  margin: 0;
   padding: 0;
   list-style: none;
   display: flex;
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,31 @@
+* {
+  box-sizing: border-box;
+  margin: 0;
```
nazarmatsevych, line 3: Don't use * selector for zeroing margins or paddings,
you can move it to body

nazarmatsevych (approved):

PR #2644, opened 2023-03-07
position stars with flexbox

Zibi95:
[DEMO LINK](https://Zibi95.github.io/layout_stars/)
[TEST REPORT LINK](https://Zibi95.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -8,6 +8,52 @@
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Stars</h1>
+    <div class="stars--0">
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+    </div>
+
+    <div class="stars--1">
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+    </div>
+
+    <div class="stars--2">
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+    </div>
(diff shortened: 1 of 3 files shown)


src/index.html
```diff
@@ -8,6 +8,52 @@
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Stars</h1>
+    <div class="stars--0">
```
Radoslaw-Czerniawski, line 11: From the requirements those blocks should have **additional** class stars--number. As those are the same types of blocks, i suggest `class="stars stars--0"`. Thanks to this you won't have to list all stars-number classes for common styles for those containers.

src/style.css
```diff
@@ -0,0 +1,69 @@
+* {
```
Radoslaw-Czerniawski, line 1: As mentioned in some other review don't use `*` for resetting styles. You don't have control what tags are reset, this may lead to some undesired behaviour. Consider creating separate css file for reset purposes and importing it here. There in the reset.css, list all tags and reset their desired styles. Ideally you would use something like [Meyer Reset](https://meyerweb.com/eric/tools/css/reset/)

src/style.css
```diff
@@ -0,0 +1,69 @@
...
+  margin: 0;
+  box-sizing: border-box;
+}
+
+.container {
```
Radoslaw-Czerniawski, line 6: Does this class even exist?

src/style.css
```diff
@@ -0,0 +1,69 @@
...
+  width: 96px;
+}
+
+.stars--0,
```
Radoslaw-Czerniawski, line 10: As mentioned in the html structure, If you would give the containing divs also `stars` class then you wouldn't need to list all of those classes here.

Radoslaw-Czerniawski (changes requested):
Please again take a look at tasks requirements. The layout is very good however use of classes is still not exactly correct.

Radoslaw-Czerniawski (approved):
Very promising prs so far 😄

mkyloa (approved):

PR #2665, opened 2023-03-10
add task solution

SuaBy514:
- [DEMO LINK](https://suaby514.github.io/layout_stars/)
- [TEST REPORT LINK](https://suaby514.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,50 @@
+body {
+  margin: 0;
+}
+
+.stars {
+  list-style: none;
+  display: flex;
+  margin: 0;
+  padding: 0;
+}
+
+.stars__star {
+  background-image: url("images/star.svg");
+  background-repeat: no-repeat;
+  background-position: 50%;
+  height: 16px;
+  width: 16px;
+  margin-left: 4px;
+}
+
+.stars--0 :first-child,
+.stars--1 :first-child,
+.stars--2 :first-child,
+.stars--3 :first-child,
+.stars--4 :first-child,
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,50 @@
...
+.stars--5 :nth-child(2),
+
+.stars--3 :nth-child(3),
+.stars--4 :nth-child(3),
+.stars--5 :nth-child(3),
+
+.stars--4 :nth-child(4),
+.stars--5 :nth-child(4),
+
+.stars--5 :nth-child(5) {
```
michalcywka, line 48: Try with `:nth-child(-n+1)` and change number in the parentheses accordingly :)

src/style.css
```diff
@@ -0,0 +1,50 @@
...
+}
+
+.stars--0 :first-child,
+.stars--1 :first-child,
+.stars--2 :first-child,
+.stars--3 :first-child,
+.stars--4 :first-child,
+.stars--5 :first-child {
+  margin-left: 0;
+}
```
michalcywka, line 28: Maybe it would be better to use `stars__star:last_child` :)

michalcywka (changes requested):
Looks good, but we can make css a bit better :)

src/style.css
SuaBy514: :last-child causes this
image

This is because the margin is on the left
:first-child produces expected results, thank you

d-wasilewski (approved):

PR #2668, opened 2023-03-10
stars solution v1

bjeziorek:
- [DEMO LINK](https://bjeziorek.github.io/layout_stars/)
- [TEST REPORT LINK](https://bjeziorek.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,48 @@
+* {
+  box-sizing: border-box;
+}
+
+body {
+  margin: 0;
+}
+
+.stars {
+  display: flex;
+}
+
+.stars > div {
+  background-image: url(images/star-active.svg);
+  background-repeat: no-repeat;
+  background-position: center;
+  height: 16px;
+  width: 16px;
+  display: inline-block;
+}
+
+.stars :not(:last-of-type) {
+  margin-right: 4px;
+}
+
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,48 @@
...
+  background-image: url(images/star.svg);
+}
+
+.stars--4 > div:nth-child(n + 5) {
+  background-image: url(images/star.svg);
+}
+
+.stars--5 > div:nth-child(n + 6) {
+  background-image: url(images/star.svg);
+}
```
michalcywka, line 48: See checklist point 2. Try to use a class instead of `div`.
Also think about starting with `(-n+1)`

michalcywka (changes requested):

src/style.css
```diff
@@ -0,0 +1,48 @@
...
+  background-image: url(images/star-active.svg);
+}
+
+.stars--4 .stars__star:nth-child(-n + 4) {
+  background-image: url(images/star-active.svg);
+}
+
+.stars--5 .stars__star:nth-child(-n + 5) {
+  background-image: url(images/star-active.svg);
+}
```
d-wasilewski, line 48: Group it, checklist 2

d-wasilewski (changes requested):

PR #2671, opened 2023-03-10
add task solution

SerhiiVoitiuk:
- [DEMO LINK](https://SerhiiVoitiuk.github.io/layout_stars/)
- [TEST REPORT LINK](https://SerhiiVoitiuk.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,37 @@
+body {
+  margin: 0;
+}
+
+.stars {
+  display: flex;
+
+}
+
+.stars__star {
+  background-image: url(./images/star.svg);
+  background-repeat: no-repeat;
+  background-position: center;
+  width: 16px;
+  height: 16px;
+  margin-right: 4px;
+}
+
+.stars--1 > .stars__star:nth-child(1) {
+  background-image: url(./images/star-active.svg);
+}
+
+.stars--2 > .stars__star:nth-child(-n+2) {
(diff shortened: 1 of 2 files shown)


src/style.css
```diff
@@ -0,0 +1,37 @@
...
+  background-image: url(./images/star-active.svg);
+}
+
+.stars--4 > .stars__star:nth-child(-n+4) {
+  background-image: url(./images/star-active.svg);
+}
+
+.stars--5 > .stars__star:nth-child(n) {
+  background-image: url(./images/star-active.svg);
+}
```
etojeDenys, line 37: join it with ','

etojeDenys (changes requested):
nit:

src/style.css
SerhiiVoitiuk: Поправив

src/style.css
```diff
@@ -0,0 +1,25 @@
...
+.stars__star {
+  background-image: url(./images/star.svg);
+  background-repeat: no-repeat;
+  background-position: center;
+  width: 16px;
+  height: 16px;
+  margin-right: 4px;
+}
+
+.stars--1 > .stars__star:nth-child(1),
```
OleziO, line 19: Try to stick to a similar code style, so I recommend replacing `1` with `-n + 1`

OleziO (approved):

PR #2672, opened 2023-03-11
add layout stars

MartaChobaniuk:
- [DEMO LINK](https://martachobaniuk.github.io/layout_stars/)
- [TEST REPORT LINK](https://martachobaniuk.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,4 +1,4 @@
-* {
+body {
   margin: 0;
 }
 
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,28 @@
+* {
+  margin: 0;
+}
```
nazarmatsevych, line 3: image

nazarmatsevych (changes requested):
GJ!

No need to set margin-right for the last elements in a row 
image

PR #2684, opened 2023-03-15
add task solution

b-buchkevych:
[DEMO LINK](https://b-buchkevych.github.io/layout_stars/)
[TEST REPORT LINK](https://b-buchkevych.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,17 +1,9 @@
-* {
+body {
   margin: 0;
-}
-
-.container {
-  height: 96px;
-  width: 96px;
-  background-color: #fff;
 }
 
 .star {
   display: flex;
-  justify-content: center;
-  align-items: center;
 }
 
 .stars__star {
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,36 @@
+* {
+  margin: 0;
+}
```
sTorba24, line 3: Don't reset margins for *, do it for tags directly, in this case you need it for body only

src/style.css
```diff
@@ -0,0 +1,36 @@
...
+
+.container {
+  height: 96px;
+  width: 96px;
+  background-color: #fff;
+}
```
sTorba24, line 9: you can remove it

src/style.css
```diff
@@ -0,0 +1,36 @@
...
+
+.star {
+  display: flex;
+  justify-content: center;
+  align-items: center;
```
sTorba24, line 14: this too

sTorba24 (changes requested):
GJ:) Lets improve it

Viktor-Kost (approved):

PR #2687, opened 2023-03-15
added task solution

mariuszgit:
https://mariuszgit.github.io/layout_stars/
https://mariuszgit.github.io/layout_stars/report/html_report/

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,43 @@
+* {
+  box-sizing: border-box;
+}
+
+body {
+  margin: 0;
+}
+
+.stars {
+  display: flex;
+  justify-content: space-between;
+  width: 100px;
+  /*dont undersand why need this padding, but its required to pass tests*/
+  padding-right: 4px;
+}
+
+.stars__star {
+  background-image: url(./images/star.svg);
+  background-position: center;
+  background-repeat: no-repeat;
+  height: 16px;
+  width: 16px;
+}
+
+.stars--1 > .stars__star:nth-child(-n+1) {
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,43 @@
+* {
```
Radoslaw-Czerniawski, line 1: As in the header example try to use something like [Meyer Reset](https://meyerweb.com/eric/tools/css/reset/), you don't have control over what tags you reset by using *. I suggest creating a separate css file for the sake of reseting and importing it to this file

src/style.css
```diff
@@ -0,0 +1,43 @@
...
+
+.stars__star {
+  background-image: url(./images/star.svg);
+  background-position: center;
+  background-repeat: no-repeat;
+  height: 16px;
+  width: 16px;
+}
+
+.stars--1 > .stars__star:nth-child(-n+1) {
```
Radoslaw-Czerniawski, line 25: You don't need > selectors, it should work with just space for example
```
.stars--2 .stars__star:nth-child(-n+2)
```
https://developer.mozilla.org/en-US/docs/Web/CSS/Descendant_combinator

Radoslaw-Czerniawski: Also you could shorten those rules by not declaring them separately but rather by listing all classes for one rule
```
}
```

Radoslaw-Czerniawski (approved):
Good pr :D, some room for improvement however nothing explicitly wrong.

src/style.css
mariuszgit: Good idea 👍🏼

PR #2699, opened 2023-03-17
Sabadyr: "layout_stars"

Sabadyr:
https://sabadyr.github.io/layout_stars/
https://sabadyr.github.io/layout_stars/report/html_report/


src/index.html
```diff
@@ -8,6 +8,61 @@
...
+    <ul>
+      <li class="stars stars--0">
+        <ul class="stars-column stars-column-0">
+          <li class="stars__star"></li>
+          <li class="stars__star"></li>
+          <li class="stars__star"></li>
+          <li class="stars__star"></li>
+          <li class="stars__star"></li>
+        </ul>
+      </li>
```
IvanFesenko, line 20: Better use div for such cases. `stars-column` - is redudant use `stars` modifier

IvanFesenko (changes requested):
Over all good job. Please refactor using div and remove extra `stars-column` wrapper

src/index.html
Sabadyr: "Better use div for such cases"
can you pls explain why?

IvanFesenko (approved):
Check my comment to avoid same mistakes in the future. If you have questions - ask them in fe_chat

PR #2700, opened 2023-03-17
stars

pos1bl:
- [DEMO LINK](https://pos1bl.github.io/layout_stars/)
- [TEST REPORT LINK](https://pos1bl.github.io/layout_stars/report/html_report/)

On tests i have errors, but if you check both websites they are simillar.

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,25 @@
+body {
+  margin: 0;
+}
+
+.stars {
+  display: flex;
+  justify-content: space-between;
+  width: 96px;
+}
+
+.stars__star {
+  background-image: url(images/star.svg);
+  background-repeat: no-repeat;
+  background-position: center;
+  width: 16px;
+  height: 16px;
+}
+
+.stars--1 .stars__star:nth-child(-n + 1),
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,25 @@
+body {
+  margin: 0;
+}
+
+.stars {
+  display: flex;
+  justify-content: space-between;
+  width: 96px;
```
nazarmatsevych, line 8: Try not to hardcode width and remove `justify-content: space-between;` rule and set correct margin-right (4px) for each star except last star in a row and re-run your tests

nazarmatsevych (changes requested):

src/style.css
pos1bl: i remake and tests are corrects, but can you explain pls, why idea to hardcode witdth and add justify-content: space-between; is bad?

VitaliyBondarenko1982: good practise to allow content make size of container
in this case it is better solution. for example if will need to add more stars, we can add and will not change styles.

VitaliyBondarenko1982 (approved):

PR #2704, opened 2023-03-17
initial commit. task solution

xpopovdmitryx:


src/style.css
```diff
@@ -0,0 +1,62 @@
+* {
+  margin: 0;
+  padding: 0;
```
witflash, line 3: Don't use any rule inside `*` selector except for `border-box`.
Reset margin and padding directly where you need.

src/style.css
```diff
@@ -0,0 +1,62 @@
...
+  background-image: url(./images/star-active.svg);
+  background-size: auto;
+  background-repeat: no-repeat;
+}
+
+.stars--5 .stars__star:nth-of-type(n) {
+  background-image: url(./images/star-active.svg);
+  background-size: auto;
+  background-repeat: no-repeat;
+}
```
witflash, line 62: Don't duplicate these rules. You can combine these selectors (with the comma) and use rules just once.

witflash (changes requested):
Check my comments, please.
And try to fix all failed test which you can see in your TEST REPORT. Ideally you should do your work with all passed tests.

polosanya (approved):
Good job!

PR #2705, opened 2023-03-18
task solved

Nazarht:
- [DEMO LINK](https://nazarht.github.io/layout_stars/)
- [TEST REPORT LINK](https://nazarht.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,71 @@
+* {
+  box-sizing: border-box;
+}
+
+body {
+  margin: 0;
+  padding: 0;
+}
+
+.stars {
+  display: flex;
+  flex-direction: row;
+}
+
+.stars_star {
+  height: 16px;
+  width: 16px;
+  background-repeat: no-repeat;
+  background-position: center;
+  margin: 0 2px;
+}
+
+.stars--0 .stars_star {
+  background-image: url(./images/star.svg);
+}
+
+.stars--1 .stars_star:first-child {
+  background-image: url(./images/star-active.svg);
+}
+
+.stars--1 .stars_star:nth-child(n + 2) {
+  background-image: url(./images/star.svg);
+}
+
+.stars--2 .stars_star:nth-child(-n + 2) {
+  background-image: url(./images/star-active.svg);
+}
+
+.stars--2 .stars_star:nth-child(n + 3) {
+  background-image: url(./images/star.svg);
+}
+
+.stars--3 .stars_star:nth-child(-n + 3) {
+  background-image: url(./images/star-active.svg);
+}
+
+.stars--3 .stars_star:nth-child(n + 4) {
+  background-image: url(./images/star.svg);
+}
+
+.stars--4 .stars_star:nth-child(-n + 4) {
+  background-image: url(./images/star-active.svg);
+}
+
+.stars--4 .stars_star:nth-child(n + 5) {
+  background-image: url(./images/star.svg);
+}
+
+.stars--5 .stars_star:nth-child(-n + 5) {
+  background-image: url(./images/star-active.svg);
+}
+
+.stars .stars_star:first-child {
+  margin-right: 2px;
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,71 @@
+* {
+  box-sizing: border-box;
+}
+
+body {
+  margin: 0;
+  padding: 0;
```
ivanchuk-yurii, line 7: body doesn't have default padding

src/style.css
```diff
@@ -0,0 +1,71 @@
...
+
+.stars .stars_star:first-child {
+  margin-right: 2px;
+  margin-left: 0;
+}
+
+.stars .stars_star:last-child {
+  margin-left: 2px;
+  margin-right: 0;
+}
```
ivanchuk-yurii, line 71: add margin to one side instead of both

src/style.css
```diff
@@ -0,0 +1,71 @@
...
+  margin: 0 2px;
+}
+
+.stars--0 .stars_star {
+  background-image: url(./images/star.svg);
+}
+
+.stars--1 .stars_star:first-child {
+  background-image: url(./images/star-active.svg);
+}
```
ivanchuk-yurii, line 29:
```suggestion
.stars--0 .stars_star,
.stars--1 .stars_star:first-child {
  background-image: url(./images/star-active.svg);
}
```
can be simplified

ivanchuk-yurii (changes requested):

src/style.css
Nazarht: but I have global styles for each element to add margin on both sides, so a have to override those styles with margin only on one side.
If I would've add only 4px margin-right to one side I still had to override the last element not to have it

Nazarht: I guess I understood how it could be done

src/style.css
```diff
@@ -0,0 +1,36 @@
...
+.stars {
+  display: flex;
+  flex-direction: row;
+}
+
+.stars_star {
+  height: 16px;
+  width: 16px;
+  background-repeat: no-repeat;
+  background-position: center;
```
ivanchuk-yurii, line 18:
```suggestion
  background: center no-repeat url(./images/star.svg);
```

src/style.css
```diff
@@ -0,0 +1,36 @@
...
+  margin-right: 4px;
+}
+
+.stars--0 .stars_star,
+.stars--1 .stars_star:nth-child(n + 2),
+.stars--2 .stars_star:nth-child(n + 3),
+.stars--3 .stars_star:nth-child(n + 4),
+.stars--4 .stars_star:nth-child(n + 5) {
+  background-image: url(./images/star.svg);
```
ivanchuk-yurii, line 27: it will be moved to `.stars_star`

src/style.css
```diff
@@ -0,0 +1,36 @@
...
+  display: flex;
+  flex-direction: row;
+}
+
+.stars_star {
+  height: 16px;
+  width: 16px;
+  background-repeat: no-repeat;
+  background-position: center;
+  margin-right: 4px;
```
ivanchuk-yurii, line 19: remove `margin` from last element

ivanchuk-yurii (changes requested):

src/style.css
OleziO: Not resolved

src/style.css
```diff
@@ -0,0 +1,31 @@
...
+  width: 16px;
+  margin-right: 4px;
+  background: no-repeat center url(./images/star.svg);
+}
+
+.stars_star:last-child {
+  margin-right: 0;
+}
+
+.stars--1 .stars_star:first-child,
```
OleziO, line 25: Make all code in one style:

replace with -n + 1

OleziO (changes requested):

vlad-demchuk (approved):

PR #2707, opened 2023-03-19
solution

VadymKapeliusnyi:
- [DEMO LINK](https://VadymKapeliusnyi.github.io/layout_stars/)
- [TEST REPORT LINK](https://VadymKapeliusnyi.github.io/layout_stars/report/html_report/)


src/style.css
```diff
@@ -0,0 +1,53 @@
+* {
+  margin: 0;
+  padding: 0;
```
vlad-demchuk, line 3: do not reset margins and paddings for `*` selector

src/style.css
```diff
@@ -0,0 +1,53 @@
...
+  background-image: url("./images/star.svg");
+}
+
+.stars--1 .stars__star:nth-child(-n + 1) {
+  background-image: url("./images/star-active.svg");
+}
+
+.stars--2 .stars__star:nth-child(-n + 2) {
+  background-image: url("./images/star-active.svg");
+}
```
vlad-demchuk, line 41:
```suggestion
.stars--1 .stars__star:nth-child(-n + 1),
.stars--2 .stars__star:nth-child(-n + 2) {
  background-image: url("./images/star-active.svg");
}
```

vlad-demchuk: do not repeat yourself, use comma for a few classes

vlad-demchuk (changes requested):
almost all test are failed, try to fix them

PR #2711, opened 2023-03-20
add stars task

msukhovych:
DEMO https://msukhovych.github.io/layout_stars/
TEST https://msukhovych.github.io/layout_stars/report/html_report/


polosanya (approved):
Good job!

PR #2731, opened 2023-03-27
stars

liza-strykharchuk:


src/style.css
```diff
@@ -0,0 +1,26 @@
...
+}
+
+.stars {
+  display: flex;
+  justify-content: center;
+}
+
+.stars__star {
+  background-position: center;
+  flex-direction: row;
```
etojeDenys, line 12: i think you need to move it to `.stars`

etojeDenys (changes requested):
looks good, but add demo link

src/style.css
liza-strykharchuk: I try to do this, but it returm an error

liza-strykharchuk: and i have demo link in the project

YegorKorenkov, line 12: flex-direction nothing do without display:flex please remove it. And then you add display: flex, flex-direction: row by default

YegorKorenkov (approved):
Look good, but next time provide demo and test links in description

PR #2734, opened 2023-03-28
add task solution

NikaNika12:
[DEMO LINK](https://NikaNika12.github.io/layout_stars/)
[TEST REPORT LINK](https://NikaNika12.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,25 @@
+* {
+  margin: 0;
+  box-sizing: border-box;
+}
+
+.stars {
+  display: flex;
+}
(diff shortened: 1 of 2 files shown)


src/style.css
```diff
@@ -0,0 +1,25 @@
+* {
+  margin: 0;
```
loralevitska, line 2: Only `box-sizing` property allowed for `*` selector
```suggestion
```

loralevitska (approved):
Good job!

PR #2746, opened 2023-04-02
add task solution

mrsargos:
- [DEMO LINK](https://mrsargos.github.io/layout_stars/)
- [TEST REPORT LINK](https://mrsargos.github.io/layout_stars/report/html_report/)


nazarmatsevych (approved):
Great, just small suggestion
No need to add margin-right to last elements in a row
image

PR #2747, opened 2023-04-02
solution

Deniskakaka:
https://deniskakaka.github.io/layout_stars/

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -1,13 +1,64 @@
-<!doctype html>
+<!DOCTYPE html>
 <html lang="en">
   <head>
     <meta charset="UTF-8">
-    <meta name="viewport"
-          content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0">
+    <meta
+      name="viewport"
+      content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0"
+    >
     <title>Stars</title>
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Stars</h1>
+    <div class="content">
+      <div class="stars--0">
+        <div class="star item"></div>
+        <div class="star item"></div>
+        <div class="star item"></div>
+        <div class="star item"></div>
+        <div class="star item"></div>
+        <div class="star item"></div>
+      </div>
+      <div class="stars--1">
+        <div class="light-star item"></div>
+        <div class="star item"></div>
+        <div class="star item"></div>
+        <div class="star item"></div>
+        <div class="star item"></div>
+        <div class="star item"></div>
+      </div>
+      <div class="stars--2">
+        <div class="light-star item"></div>
+        <div class="light-star item"></div>
+        <div class="star item"></div>
+        <div class="star item"></div>
+        <div class="star item"></div>
+        <div class="star item"></div>
+      </div>
+      <div class="stars--3">
+        <div class="light-star item"></div>
+        <div class="light-star item"></div>
+        <div class="light-star item"></div>
+        <div class="star item"></div>
+        <div class="star item"></div>
+        <div class="star item"></div>
+      </div>
+      <div class="stars--4">
+        <div class="light-star item"></div>
+        <div class="light-star item"></div>
+        <div class="light-star item"></div>
+        <div class="light-star item"></div>
+        <div class="star item"></div>
+        <div class="star item"></div>
+      </div>
+      <div class="stars--5">
+        <div class="light-star item"></div>
+        <div class="light-star item"></div>
+        <div class="light-star item"></div>
+        <div class="light-star item"></div>
+        <div class="light-star item"></div>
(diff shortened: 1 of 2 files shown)


src/index.html
```diff
@@ -1,13 +1,64 @@
...
+    <meta
+      name="viewport"
+      content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0"
+    >
     <title>Stars</title>
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Stars</h1>
+    <div class="content">
```
VitaliyBondarenko1982, line 13:
```suggestion
```
redundant

src/index.html
```diff
@@ -1,13 +1,64 @@
...
+      <div class="stars--0">
+        <div class="star item"></div>
+        <div class="star item"></div>
+        <div class="star item"></div>
+        <div class="star item"></div>
+        <div class="star item"></div>
+        <div class="star item"></div>
+      </div>
```
VitaliyBondarenko1982, line 21: Screenshot from 2023-04-02 15-25-00
should be five stars in row
stars wil be responsible for general styles of row, stars--0 stars--1...stars--5 modifiers wil be responsible only for amount of active stars
```suggestion
      <div class="stars stars--0">
        <div class="stars__star"></div>
        <div class="stars__star"></div>
        <div class="stars__star"></div>
        <div class="stars__star"></div>
        <div class="stars __star"></div>
      </div>
```

src/style.css
```diff
@@ -0,0 +1,31 @@
+.content {
+  display: flex;
+  flex-direction: column;
+  width: 106px;
+  justify-content: space-between;
+}
```
VitaliyBondarenko1982, line 6:
```suggestion
```
redundant

src/index.html
```diff
@@ -1,13 +1,64 @@
...
+      <div class="stars--1">
+        <div class="light-star item"></div>
+        <div class="star item"></div>
+        <div class="star item"></div>
+        <div class="star item"></div>
+        <div class="star item"></div>
+        <div class="star item"></div>
+      </div>
+      <div class="stars--2">
+        <div class="light-star item"></div>
```
VitaliyBondarenko1982, line 31: for each star should be just one class
for example stars__star.
for amount active stars should be responsible modifier class on parent element
```suggestion
        <div class="stars__star"></div>
```

src/style.css
```diff
@@ -0,0 +1,31 @@
...
+
+.item {
```
VitaliyBondarenko1982, line 8: Reset browser's default margin

```suggestion
body {
  margin: 0;
}

.item {
```

VitaliyBondarenko1982 (changes requested):
please check task description and rewrite styles
Screenshot from 2023-04-02 15-32-28

also push changes in readme.md and provide demo link for TEST REPORT in PR description

src/index.html
Deniskakaka: why if i need active or disable star? it just than use css selector

src/style.css
```diff
@@ -0,0 +1,21 @@
...
+  flex-direction: row;
+  background-repeat: no-repeat;
+  width: 16px;
+  height: 16px;
+  background-image: url("./images/star.svg");
+}
+
+.active {
+  background-image: url("./images/star-active.svg");
+}
```
VitaliyBondarenko1982, line 16:
```suggestion
```
redundant
stars--n modifier on each block should be responsible for amount of active stars
something like that
```
...
}
```

src/index.html
```diff
@@ -1,13 +1,56 @@
...
+      <div class="stars__star active"></div>
+      <div class="stars__star"></div>
+    </div>
+    <div class="stars stars--5">
+      <div class="stars__star active"></div>
+      <div class="stars__star active"></div>
+      <div class="stars__star active"></div>
+      <div class="stars__star active"></div>
+      <div class="stars__star active"></div>
+    </div>
```
VitaliyBondarenko1982, line 54:
```suggestion
    <div class="stars stars--5">
      <div class="stars__star"></div>
      <div class="stars__star"></div>
      <div class="stars__star"></div>
      <div class="stars__star"></div>
      <div class="stars__star"></div>
    </div>
```
check comment below and remove all active classes.
they are redundant
for amount of active stars should be responsible stars--n modifier

VitaliyBondarenko1982 (changes requested):
also push changes in readme.md file and provide test report link

src/index.html
Deniskakaka: in js you will be this approach, no you will be use classNames and use class active, why not use this so

VitaliyBondarenko1982: because it is part of task description for learning modifiers and nth-child selector.
Screenshot from 2023-04-02 18-01-13

also in js (react) in real world from server we can receive amount of active cards and can make dynamically modifier class which will be responsible for amount of active cards.
it is very common and more flexible approach to make parent responsible for child view.

Deniskakaka: я убрал уже, что еще не так

PR #2750, opened 2023-04-03
Develop

LizKumber:
https://LizKumber.github.io/layout_stars/
https://LizKumber.github.io/layout_stars/report/html_report/

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -8,6 +8,73 @@
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Stars</h1>
+    <div class="block">
+      <div class="stars stars--0">
+        <div class="stars__star"></div>
+        
+        <div class="stars__star"></div>
+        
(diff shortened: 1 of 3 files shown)


src/index.html
```diff
@@ -8,6 +8,73 @@
...
+        <div class="stars__star"></div>
+        
+        <div class="stars__star"></div>
+        
+        <div class="stars__star"></div>
+        
+        <div class="stars__star"></div>
+      </div>
+      <div class="stars stars--1">
+        <div class="stars__star star_active"></div>
```
nazarmatsevych, line 24: Due to the task description you dont need to add additional classes like star__active
image

Instead you can use :nth-child to set appropriate image

nazarmatsevych (changes requested):
no need to add margin-right to the last elements in a row
image

vlad-demchuk (approved):

PR #2751, opened 2023-04-03
Stars

levchenko-dmytro:
- [DEMO LINK](https://levchenko-dmytro.github.io/layout_stars/)
- [TEST REPORT LINK](https://levchenko-dmytro.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -7,20 +7,11 @@ body {
 }
 
 .stars {
+  display: flex;
   background-color: #fff;
 }
 
-.stars--0,
-.stars--1,
-.stars--2,
-.stars--3,
-.stars--4,
-.stars--5 {
-  display: flex;
-}
-
 .stars__star {
-  display: flex;
   height: 16px;
   width: 16px;
   background: url(./images/star.svg) center no-repeat;
--- a/src/index.html
+++ b/src/index.html
@@ -8,56 +8,54 @@
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <div class="stars">
 
-      <div class="stars--0">
-        <div class="stars__star"></div>
-        <div class="stars__star"></div>
-        <div class="stars__star"></div>
-        <div class="stars__star"></div>
-        <div class="stars__star"></div>
-      </div>
-
-      <div class="stars--1">
-        <div class="stars__star"></div>
-        <div class="stars__star"></div>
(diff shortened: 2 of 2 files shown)


src/style.css
```diff
@@ -0,0 +1,40 @@
+body {
+  margin: unset;
+}
+
+.stars {
+  box-sizing: border-box;
```
Viktor-Kost, line 6: move `box-sizing` to `*` selector

src/style.css
```diff
@@ -0,0 +1,40 @@
...
+}
+
+.stars__star {
+  display: flex;
+  content: "";
+  height: 16px;
+  width: 16px;
+  background-image: url(./images/star.svg);
+  background-repeat: no-repeat;
+  background-size: 13px;
```
Viktor-Kost, line 22: redundant 
```suggestion
```

Viktor-Kost (changes requested):

src/style.css
levchenko-dmytro: Please check, changes made as per your advise.

src/style.css
```diff
@@ -0,0 +1,34 @@
...
+
+body {
+  margin: unset;
+}
+
+.stars {
+  background-color: #fff;
+}
+
+.stars > [class] {
```
vlad-demchuk, line 13: do not increase css specificity, use just classes

src/style.css
```diff
@@ -0,0 +1,34 @@
+* {
+  box-sizing: border-box;
+}
+
+body {
+  margin: unset;
```
vlad-demchuk, line 6: why do we need unset?

vlad-demchuk (changes requested):

levchenko-dmytro:
Please check, corrections done as per your request.

src/style.css
```diff
@@ -0,0 +1,30 @@
...
+  height: 16px;
+  width: 16px;
+  background: url(./images/star.svg) center no-repeat;
+}
+
+.stars__star:not(:last-child) {
+  margin-right: 4px;
+}
+
+.stars--1 > .stars__star:first-child,
```
OleziO, line 24: Make all code in the same style. Replace with `-n + 1`

OleziO (approved):

PR #2754, opened 2023-04-04
add task solution

VolodymyrPst:


src/style.css
```diff
@@ -0,0 +1,35 @@
+body {
+  margin: 0;
+}
+
+.start__block {
+  /* width: 96px; */
+  /* display: flex;
+  flex-direction: column; */
+}
```
anstsot, line 9: remove the styles if you're not using

anstsot (changes requested):
Good job, but there are few fixes to improve your code. Please check the tests because they're failed

PR #2783, opened 2023-04-11
add task solution

shatikhin:


src/style.css
```diff
@@ -0,0 +1,49 @@
...
+.stars--4 .stars__star:nth-child(4) {
+  background-image: url(images/star-active.svg);
+  background-repeat: no-repeat;
+}
+
+.stars--5 .stars__star:nth-child(1),
+.stars--5 .stars__star:nth-child(2),
+.stars--5 .stars__star:nth-child(3),
+.stars--5 .stars__star:nth-child(4),
+.stars--5 .stars__star:nth-child(5N) {
```
anstsot, line 46: you can use a simpler selector. please fix it for all classes and merge similar styles
```suggestion
.stars--5 .stars__star:nth-child(-n+5) {
```

anstsot (changes requested):
It’s almost done! Let’s improve a little. Please check your report link because it seems to be broken

PR #2784, opened 2023-04-11
add task solution

maksym-nemera:
- [DEMO LINK](https://maksym-nemera.github.io/layout_stars/)
- [TEST REPORT LINK](https://maksym-nemera.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -7,56 +7,56 @@
     <title>Stars</title>
     <link rel="stylesheet" href="./style.css">
   </head>
-  
+
   <body>
-    <div class="all-stars">
-      <div class="stars stars--0">
-        <div class="stars__star"></div>
-        <div class="stars__star"></div>
-        <div class="stars__star"></div>
-        <div class="stars__star"></div>
-        <div class="stars__star"></div>
-      </div>
+    <div class="stars__container">
+      <ul class="stars stars--0">
+        <li class="stars__star"></li>
+        <li class="stars__star"></li>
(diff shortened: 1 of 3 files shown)


src/index.html
```diff
@@ -7,7 +7,56 @@
...
   <body>
-    <h1>Stars</h1>
+    <div class="all-stars">
+      <div class="stars stars--0">
+        <div class="stars__star"></div>
+        <div class="stars__star"></div>
+        <div class="stars__star"></div>
+        <div class="stars__star"></div>
+        <div class="stars__star"></div>
+      </div>
```
naziamoff, line 19: would be great to use semantic tags here. You got a list with list items here, right?

src/index.html
```diff
@@ -7,7 +7,56 @@
     <title>Stars</title>
     <link rel="stylesheet" href="./style.css">
   </head>
+  
   <body>
-    <h1>Stars</h1>
+    <div class="all-stars">
```
naziamoff, line 12: image

If that's not what you mean - it is better to rename this class for sth like stars__container

naziamoff (approved):
Awesome 👍 

See some minor suggestions below

src/index.html
NickMitchell61: You've recommended change <div> on ul>li*5, Am I clearly understanding you?

src/index.html
NickMitchell61: Got it!

PR #2786, opened 2023-04-11
add stars layout

cephy77:
[DEMO LINK](https://cephy77.github.io/layout_stars/)
[TEST REPORT LINK](https://cephy77.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -8,18 +8,12 @@
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1 class="heading">Stars</h1>
-
     <div class="stars__container">
       <ul class="stars stars--0">
         <li class="stars__star"></li>
-
         <li class="stars__star"></li>
-
         <li class="stars__star"></li>
-
         <li class="stars__star"></li>
-
         <li class="stars__star"></li>
       </ul>
     </div>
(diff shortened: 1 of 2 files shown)


src/index.html
```diff
@@ -8,6 +8,90 @@
...
+      <ul class="stars stars--0">
+        <li class="stars__star"></li>
+
+        <li class="stars__star"></li>
+
+        <li class="stars__star"></li>
+
+        <li class="stars__star"></li>
+
+        <li class="stars__star"></li>
```
naziamoff, line 23: They are pretty much the same, so I would say that spaces are redundant here. It does not improve readability, however, makes your code less ergonomic
```suggestion
        <li class="stars__star"></li>
        <li class="stars__star"></li>
        <li class="stars__star"></li>
        <li class="stars__star"></li>
        <li class="stars__star"></li>
```

src/style.css
```diff
@@ -0,0 +1,45 @@
+body {
+  margin: 0;
+}
+
+.heading {
+  position: absolute;
+  width: 1px;
+  height: 1px;
+  margin: -1px;
```
naziamoff, line 9: that seems wrong to me adding size/positioning to title. Do you really need it? What is it for?

naziamoff (changes requested):
Looks good 👍 
But what are those css rules for heading? I bet you don't need all of them, probably just some. Let's fix this and you're good

src/style.css
cephy77: I'm just making h1 visually hidden using [this](https://gist.github.com/luxplanjay/6f6936b1bba7afe006fa9317b07b6045) code

naziamoff (approved):
🔥

PR #2789, opened 2023-04-11
add task solution

prostonadobobb:
[DEMO LINK](https://prostonadobobb.github.io/layout_stars/)
[TEST REPORT LINK](https://prostonadobobb.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -8,6 +8,49 @@
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Stars</h1>
+
+    <div class="star stars--0">
+      <div class="stars__star"></div>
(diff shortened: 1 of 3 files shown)


src/index.html
```diff
@@ -8,6 +8,49 @@
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Stars</h1>
+
+    <div class="star stars--0">
```
vlad-demchuk, line 12:
```suggestion
    <div class="stars stars--0">
```
bem block should have the same name as beginning of bem elements

vlad-demchuk (changes requested):

IvanFesenko (approved):

PR #2797, opened 2023-04-13
add task solution

SerhiiVizir:
- [DEMO LINK](https://SerhiiVizir.github.io/layout_stars/)
- [TEST REPORT LINK](https://SerhiiVizir.github.io/layout_stars/report/html_report/)
- I'm sorry, but I don't understand what to change in the code

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -8,81 +8,81 @@
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <div class="stars">
-      <div class="stars__star stars--0">
+    <div class="stars stars--0">
+      <div class="stars__star">
       </div>
-      <div class="stars__star stars--0">
+      <div class="stars__star">
       </div>
-      <div class="stars__star stars--0">
+      <div class="stars__star">
       </div>
-      <div class="stars__star stars--0">
+      <div class="stars__star">
       </div>
-      <div class="stars__star stars--0">
-      </div>
-    </div>
-
-    <div class="stars">
-      <div class="stars__star stars--1">
-      </div>
-      <div class="stars__star stars--1">
-      </div>
-      <div class="stars__star stars--1">
-      </div>
-      <div class="stars__star stars--1">
-      </div>
-      <div class="stars__star stars--1">
+      <div class="stars__star">
       </div>
     </div>
 
-    <div class="stars">
-      <div class="stars__star stars--2">
+    <div class="stars stars--1">
+      <div class="stars__star">
       </div>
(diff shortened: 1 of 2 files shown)


src/style.css
```diff
@@ -0,0 +1,38 @@
...
+  margin: 0;
+}
+
+.stars {
+  display: flex;
+  align-items: center;
+}
+
+.stars__star {
+  background-image: url(/images/star.svg);
```
mykhalenych, line 11:
```
  background-image: url(/images/star.svg);
  background-position: center;
  background-repeat: no-repeat;
```

you need only this values for background

src/style.css
```diff
@@ -0,0 +1,38 @@
...
+  background-repeat: no-repeat;
+  background-size: contain;
+  width: 16px;
+  height: 16px;
+  margin-right: 4px;
+  border-radius: 0.5px;
+}
+
+.stars--1:nth-child(-n+1) {
```
mykhalenych, line 20:
```suggestion
.stars--1:nth-child(-n+1),
.stars--2:nth-child(-n+2),
... {
  background-image: url(/images/star-active.svg);
}
```

src/style.css
```diff
@@ -0,0 +1,38 @@
...
+  align-items: center;
+}
+
+.stars__star {
+  background-image: url(/images/star.svg);
+  background-repeat: no-repeat;
+  background-size: contain;
+  width: 16px;
+  height: 16px;
+  margin-right: 4px;
```
mykhalenych, line 16: don't add for last start in row

mykhalenych (changes requested):

src/index.html
```diff
@@ -8,6 +8,82 @@
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Stars</h1>
+    <div class="stars">
+      <div class="stars__star stars--0">
```
TarasHoliuk, line 12: modificator should be in `stars` BEM-block instead of each `stars__star` element. It should look like this:

```suggestion
    <div class="stars stars--0">
      <div class="stars__star">
```

TarasHoliuk: Then you should add changes to CSS (use chained selector - [CSS combinator](https://www.w3schools.com/css/css_combinators.asp))

TarasHoliuk (changes requested):
I've added comments how to fix task
If you have some question or need help - feel free to ask in the chat

IvanFesenko (approved):

PR #2803, opened 2023-04-15
add task solution

MaksymSikora:


src/style.css
```diff
@@ -0,0 +1,72 @@
+* {
+  margin: 0;
+  padding: 0;
+}
```
polosanya, line 4: Don't reset margins and paddings using * selector. Better specify the exact tags you want to reset

src/style.css
```diff
@@ -0,0 +1,72 @@
...
+
+.container {
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  flex-direction: column;
+  /* margin-top: 1px; */
+  /* margin-left: 2px; */
```
polosanya, line 12: Remove the useless comments

src/style.css
```diff
@@ -0,0 +1,72 @@
...
+  background-image: url(./images/star-active.svg);
+}
+
+.stars--3 div:nth-child(n + 4) {
+  background-image: url(./images/star.svg);
+}
+
+.stars--4 .stars__star {
+  background-image: url(./images/star-active.svg);
+}
```
polosanya, line 64: image

polosanya (changes requested):
Please don't forget to add demo and test links to the PR description
image

src/style.css
Volodymyr-Melnyk-Fron10der: not fixed
image

Volodymyr-Melnyk-Fron10der (changes requested):

polosanya (approved):
Well done!

PR #2809, opened 2023-04-18
add task solution

AlexKherson:


src/index.html
```diff
@@ -8,6 +8,47 @@
...
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+    </div>
+    <div class="stars stars--2">
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+    </div>
```
naziamoff, line 31: as those are on the same level of nesting and there are quite a lot of them - you can add some empty lines between them for better readability

```suggestion
    <div class="stars stars--0">
      <div class="stars__star"></div>
      <div class="stars__star"></div>
      <div class="stars__star"></div>
      <div class="stars__star"></div>
      <div class="stars__star"></div>
    </div>
    
    <div class="stars stars--1">
      <div class="stars__star"></div>
      <div class="stars__star"></div>
      <div class="stars__star"></div>
      <div class="stars__star"></div>
      <div class="stars__star"></div>
    </div>
    
    <div class="stars stars--2">
      <div class="stars__star"></div>
      <div class="stars__star"></div>
      <div class="stars__star"></div>
      <div class="stars__star"></div>
      <div class="stars__star"></div>
    </div>
```

naziamoff (approved):
everything seems great to me 🔥 👍 
There are some small suggestions for keeping them in mind in your following Pull Requests

But I don't see Demo Link and Test Report links in your description. For this reason - I reject work. Let's add it - and we're good.

naziamoff (changes requested):
oops, I was planning to reject

src/index.html
naziamoff: Also, it is better to use semantic tags here. Afterall, you got a list here, right?

naziamoff (approved):
🔥 🔥 🔥

PR #2810, opened 2023-04-18
add task solution

IhorVerba:
https://IhorVerba.github.io/layout_stars/
https://IhorVerba.github.io/layout_stars/report/html_report/

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -8,6 +8,49 @@
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Stars</h1>
+    <div class="box">
+      <div class="stars stars--0">
+        <div class="stars__star"></div>
+        <div class="stars__star"></div>
+        <div class="stars__star"></div>
+        <div class="stars__star"></div>
+        <div class="stars__star"></div>
+      </div>
+      <div class="stars stars--1">
+        <div class="stars__star"></div>
+        <div class="stars__star"></div>
+        <div class="stars__star"></div>
+        <div class="stars__star"></div>
+        <div class="stars__star"></div>
+      </div>
+      <div class="stars stars--2">
+        <div class="stars__star"></div>
+        <div class="stars__star"></div>
+        <div class="stars__star"></div>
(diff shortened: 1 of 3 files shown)


src/index.html
```diff
@@ -8,6 +8,54 @@
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Stars</h1>
+    <div class="block">
```
Viktor-Kost, line 11: The 'block' class for the parent div seems unnecessary. Consider removing it or giving it a more descriptive name like 'stars-container'.

src/style.css
```diff
@@ -0,0 +1,24 @@
...
+}
+
+.stars__star {
+  background-image: url(images/star.svg);
+  background-repeat: no-repeat;
+  background-position: 0;
+  width: 16px;
+  height: 16px;
+  margin: 0 2px;
+}
```
Viktor-Kost, line 16: we need to clear extra margin to avoid margin collapse. So you can just do like this
```suggestion
}

.stars__star:not(:last-child) {
  margin-right: 4px
}
```

src/style.css
```diff
@@ -0,0 +1,24 @@
...
+
+.stars {
+  display: flex;
+}
+
+.stars__star {
+  background-image: url(images/star.svg);
+  background-repeat: no-repeat;
+  background-position: 0;
```
Viktor-Kost, line 12: apply background `shorthand`

src/style.css
```diff
@@ -0,0 +1,24 @@
+* {
+  margin: 0;
+}
```
Viktor-Kost, line 3: it's allowed to use only box-sizing: border-box for * selector
If you need to reset default margin you can apply it to the tag selectors

src/style.css
Viktor-Kost, line 12: it will be clearer to use `center`
```suggestion
  background-position: center;
```

Viktor-Kost (changes requested):

vlad-demchuk (approved):

PR #2822, opened 2023-04-19
add task solution

Mykola-Maik:
- [DEMO LINK](https://Mykola-Maik.github.io/layout_stars/)
- [TEST REPORT LINK](https://Mykola-Maik.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -2,12 +2,85 @@
 <html lang="en">
   <head>
     <meta charset="UTF-8">
-    <meta name="viewport"
-          content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0">
+    <meta
+      name="viewport"
+      content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0"
+    >
     <title>Stars</title>
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Stars</h1>
+    <div class="stars stars--0">
+      <div class="stars__star">
+      </div>
+      <div class="stars__star">
+      </div>
+      <div class="stars__star">
+      </div>
+      <div class="stars__star">
+      </div>
+      <div class="stars__star">
+      </div>
+    </div>
+    <div class="stars stars--1">
+      <div class="stars__star">
+      </div>
+      <div class="stars__star">
+      </div>
+      <div class="stars__star">
+      </div>
+      <div class="stars__star">
+      </div>
+      <div class="stars__star">
+      </div>
+    </div>
+    <div class="stars stars--2">
+      <div class="stars__star">
+      </div>
+      <div class="stars__star">
+      </div>
+      <div class="stars__star">
+      </div>
+      <div class="stars__star">
+      </div>
+      <div class="stars__star">
+      </div>
+    </div>
+    <div class="stars stars--3">
+      <div class="stars__star">
+      </div>
+      <div class="stars__star">
+      </div>
+      <div class="stars__star">
+      </div>
+      <div class="stars__star">
+      </div>
+      <div class="stars__star">
+      </div>
+    </div>
+    <div class="stars stars--4">
+      <div class="stars__star">
+      </div>
+      <div class="stars__star">
+      </div>
+      <div class="stars__star">
+      </div>
+      <div class="stars__star">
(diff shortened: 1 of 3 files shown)


Mykola-Maik:
Is it a good practice "make some describing comments"?

src/style.css
```diff
@@ -0,0 +1,109 @@
...
+}
+
+/* STARS--0 */
+
+.stars--0 .stars__star:nth-of-type(1),
+.stars--0 .stars__star:nth-of-type(2),
+.stars--0 .stars__star:nth-of-type(3),
+.stars--0 .stars__star:nth-of-type(4),
+.stars--0 .stars__star:nth-of-type(5) {
+  background-image: url("./images/star.svg");
```
naziamoff, line 31: Instead of repeating the background styles for each modifier class. You can set the default background-image, background-repeat, and background-position styles once for the .stars__star class and only update the background-image for active stars in the modifier classes. 💡

src/index.html
```diff
@@ -2,12 +2,85 @@
...
+      </div>
+      <div class="stars__star">
+      </div>
+      <div class="stars__star">
+      </div>
+      <div class="stars__star">
+      </div>
+      <div class="stars__star">
+      </div>
+    </div>
```
naziamoff, line 24: to me it seems that you should have a list with list items here. `div` makes no sense, so it's better to add some semantics here

naziamoff: and also, as those stars blocks are on the same level of nesting and there are quite a lot of them - you can add some empty lines between them for better readability

naziamoff (changes requested):
you can create a single 5-lines selector instead of creating 5 of them for each of the star blocks. Try to figure out how

naziamoff:
> Is it a good practice "make some describing comments"?

nope, not really. Good code is usually self-documenting. You should not leave comments unless you really need to

Mykola-Maik:
If I fixed mistakes should I write the name 'add task solution' or another text like 'fixed mistakes'?

src/index.html
```diff
@@ -2,12 +2,79 @@
...
+    <ul class="stars stars--0">
+      <li class="stars__star"></li>
+
+      <li class="stars__star"></li>
+
+      <li class="stars__star"></li>
+
+      <li class="stars__star"></li>
+
+      <li class="stars__star"></li>
```
naziamoff, line 22:
```suggestion
      <li class="stars__star"></li>
      <li class="stars__star"></li>
      <li class="stars__star"></li>
      <li class="stars__star"></li>
      <li class="stars__star"></li>
```

src/style.css
```diff
@@ -0,0 +1,61 @@
...
+  background-image: url("./images/star-active.svg");
+}
+
+.stars--5 .stars__star:nth-of-type(1),
+.stars--5 .stars__star:nth-of-type(2),
+.stars--5 .stars__star:nth-of-type(3),
+.stars--5 .stars__star:nth-of-type(4),
+.stars--5 .stars__star:nth-of-type(5) {
+  background-image: url("./images/star-active.svg");
+}
```
naziamoff, line 61: seems it is not yet fixed
Screenshot 2023-04-20 at 16 38 37

naziamoff (changes requested):
> If I fixed mistakes should I write the name 'add task solution' or another text like 'fixed mistakes'?

As a commit name you should write what exactly you did in this commit. e.g. `fixed indents` or `slight code style changes` or sth like that

src/style.css
```diff
@@ -0,0 +1,51 @@
...
+
+.stars--4 .stars__star:nth-of-type(-n + 4) {
+  background-image: url("./images/star-active.svg");
+}
+
+.stars--5 .stars__star:nth-of-type(-n + 5) {
+  background-image: url("./images/star-active.svg");
+}
```
naziamoff, line 51: [checklist p.2](https://github.com/mate-academy/layout_stars/blob/master/checklist.md)

naziamoff (changes requested):

src/style.css
Mykola-Maik: Ohh, now I get it))) Sorry...

naziamoff (approved):
great 🔥

src/style.css
naziamoff: next time - feel free to ask if you're hesitant about the comment

PR #2826, opened 2023-04-19
add task solution layout-stars

annholovko:
[DEMO LINK](https://annholovko.github.io/layout_stars/)
[TEST REPORT LINK](https://annholovko.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,30 @@
+* {
+  box-sizing: border-box;
+}
+
+body {
+  margin: 0;
+}
+
+.stars {
+  display: flex;
+}
+
+.stars__star {
+  height: 16px;
+  width: 16px;
+  margin-right: 4px;
+  background-image: url(images/star.svg);
+  background-repeat: no-repeat;
+  background-position: center;
+}
+
+.stars--1 .stars__star:first-child,
+.stars--2 .stars__star:nth-child(-n+2),
+.stars--3 .stars__star:nth-child(-n+3),
+.stars--4 .stars__star:nth-child(-n+4),
+.stars--5 .stars__star:nth-child(-n+5) {
+  background-image: url(images/star-active.svg);
+  background-repeat: no-repeat;
+  background-position: center;
+}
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,30 @@
...
+}
+
+.stars__star {
+  height: 16px;
+  width: 16px;
+  margin-right: 4px;
+  background-image: url(images/star.svg);
+  background-repeat: no-repeat;
+  background-position: center;
+}
```
Ihor-Karpyn, line 20:
```suggestion
.stars__star {
  height: 16px;
  width: 16px;
  margin-right: 4px;
  
  background-image: url(images/star.svg);
  background-repeat: no-repeat;
  background-position: center;
}
```

src/style.css
```diff
@@ -0,0 +1,30 @@
...
+
+.stars--1 .stars__star:first-child,
+.stars--2 .stars__star:nth-child(-n+2),
+.stars--3 .stars__star:nth-child(-n+3),
+.stars--4 .stars__star:nth-child(-n+4),
+.stars--5 .stars__star:nth-child(-n+5) {
+  background-image: url(images/star-active.svg);
+  background-repeat: no-repeat;
+  background-position: center;
+}
```
Ihor-Karpyn, line 30: This stile doesn't change depending on the stars-n modifier, let's move it to `stars__star` selector

```
  background-repeat: no-repeat;
  background-position: center;
```

src/style.css
```diff
@@ -0,0 +1,30 @@
...
+
+.stars {
+  display: flex;
+}
+
+.stars__star {
+  height: 16px;
+  width: 16px;
+  margin-right: 4px;
```
Ihor-Karpyn, line 16: Removing margin-right from the last element and margin-left from the first element of a series of elements that are displayed inline or as a horizontal list is important because it helps to prevent unwanted spacing or gaps between the elements.

Ihor-Karpyn (changes requested):
GJ, almost done

Ihor-Karpyn (approved):
🔥

PR #2828, opened 2023-04-20
add task solution

ulositska:
- [DEMO LINK](https://ulositska.github.io/layout_stars/)
- [TEST REPORT LINK](https://ulositska.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -8,6 +8,47 @@
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Stars</h1>
+    <div class="stars stars--0">
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+    </div>
+    <div class="stars stars--1">
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+    </div>
+    <div class="stars stars--2">
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+    </div>
+    <div class="stars stars--3">
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+    </div>
+    <div class="stars stars--4">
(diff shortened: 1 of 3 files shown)


src/index.html
```diff
@@ -8,6 +8,47 @@
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Stars</h1>
+    <div class="stars stars--0">
```
naziamoff, line 11: take a look at the [checklist](https://github.com/mate-academy/layout_stars/blob/master/checklist.md) pls and fix your layout accordingly

naziamoff (changes requested):

src/style.css
```diff
@@ -0,0 +1,42 @@
...
+  background-image: url(\images\star-active.svg);
+}
+
+.stars--4 > :nth-child(-n+4) {
+  background-image: url(\images\star-active.svg);
+}
+
+.stars--5 > :nth-child(-n+5) {
+  background-image: url(\images\star-active.svg);
+}
```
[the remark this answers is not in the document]
ulositska: I changed it, well sort of. 
Is there any way we can reuse the '--n' part of the classname in :nth-child?

src/index.html
ulositska: I don't know you, Nazar, but I already worship your patience. 
Thanks!

src/index.html
```diff
@@ -8,6 +8,76 @@
...
+    <div class="stars stars--1">
+      <div class="stars__star"></div>
+
+      <div class="stars__star"></div>
+
+      <div class="stars__star"></div>
+
+      <div class="stars__star"></div>
+
+      <div class="stars__star"></div>
```
naziamoff, line 32: Actually it is not necessary to make empty lines between all of the elements of the same nesting level. It is not a mistake, but I would say if you leave this part without empty lines - it will become even more readable, as they're pretty much the same and I don't need to distinguish between them
```suggestion
      <div class="stars__star"></div>
      <div class="stars__star"></div>
      <div class="stars__star"></div>
      <div class="stars__star"></div>
      <div class="stars__star"></div>
```

src/style.css
```diff
@@ -0,0 +1,30 @@
...
+  margin-right: 0;
+}
+
+.stars--1 > :nth-child(-n+1),
+.stars--2 > :nth-child(-n+2),
+.stars--3 > :nth-child(-n+3),
+.stars--4 > :nth-child(-n+4),
+.stars--5 > :nth-child(-n+5) {
+  background-image: url(\images\star-active.svg);
+}
```
naziamoff, line 30: yeah, that's it 🔥

naziamoff (approved):
well done!

PR #2829, opened 2023-04-20
Add task solution

amashutikov:
- [DEMO LINK](https://amashutikov.github.io/layout_stars/)
- [TEST REPORT LINK](https://amashutikov.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -10,6 +10,7 @@ body {
 }
 
 .stars__star {
+  background-image: url(/images/star.svg);
   background-repeat: no-repeat;
   background-position: center;
   margin-left: 4px;
@@ -17,7 +18,7 @@ body {
   width: 16px;
 }
 
-.stars--0 > * {
+.stars--0 .stars__star {
   background-image: url(/images/star.svg);
 }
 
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,41 @@
...
+.stars--5 .stars__star:nth-child(-n+5) {
+  background-image: url(/images/star-active.svg);
+}
+
+.stars--1 .stars__star:nth-child(n+2),
+.stars--2 .stars__star:nth-child(n+3),
+.stars--3 .stars__star:nth-child(n+4),
+.stars--4 .stars__star:last-child {
+  background-image: url(/images/star.svg);
+}
```
naziamoff, line 41: let's add this to `stars__star` selector, for all of the stars. And after that reassing it for those that need it

naziamoff (changes requested):
not bad, but there are some things to fix

src/style.css
```diff
@@ -0,0 +1,35 @@
...
+  background-repeat: no-repeat;
+  background-position: center;
+  margin-left: 4px;
+  height: 16px;
+  width: 16px;
+}
+
+.stars--0 .stars__star {
+  background-image: url(/images/star.svg);
+}
```
naziamoff, line 23: I don't think you need this one. In the selector above you are already doing that, so it does not seem to affect your stars behavior at all. Check this out and let me know

naziamoff (changes requested):
almost there, just one thing left

naziamoff (approved):
great 👍

PR #2831, opened 2023-04-20
task solution

okosohor:
- [DEMO LINK](https://okosohor.github.io/layout_stars/)
- [TEST REPORT LINK](https://okosohor.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -7,7 +7,50 @@
     <title>Stars</title>
     <link rel="stylesheet" href="./style.css">
   </head>
-  <body>
-    <h1>Stars</h1>
+  <body class="body">
+    <div class="stars">
+      <div class="stars--0">
+        <div class="stars__star"></div>
+        <div class="stars__star"></div>
+        <div class="stars__star"></div>
+        <div class="stars__star"></div>
+        <div class="stars__star"></div>
+      </div>
+      <div class="stars--1">
+        <div class="active stars__star"></div>
+        <div class="stars__star"></div>
+        <div class="stars__star"></div>
+        <div class="stars__star"></div>
+        <div class="stars__star"></div>
+      </div>
+      <div class="stars--2">
+        <div class="active stars__star"></div>
+        <div class="active stars__star"></div>
+        <div class="stars__star"></div>
+        <div class="stars__star"></div>
+        <div class="stars__star"></div>
+      </div>
+      <div class="stars--3">
+        <div class="active stars__star"></div>
+        <div class="active stars__star"></div>
+        <div class="active stars__star"></div>
+        <div class="stars__star"></div>
+        <div class="stars__star"></div>
+      </div>
+      <div class="stars--4">
+        <div class="active stars__star"></div>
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,41 @@
...
+  margin: 0;
+}
+
+.stars__star {
+  background-image: url(images/star.svg);
+  background-position: center center;
+  background-repeat: no-repeat;
+  width: 16px;
+  height: 16px;
+  margin-right: 4px;
```
Ihor-Karpyn, line 15: Removing margin-right from the last element and margin-left from the first element of a series of elements that are displayed inline or as a horizontal list is important because it helps to prevent unwanted spacing or gaps between the elements.

src/index.html
```diff
@@ -7,7 +7,55 @@
...
+      <div class="stars--0">
+        <div class="stars__star"></div>
+        <div class="stars__star"></div>
+        <div class="stars__star"></div>
+        <div class="stars__star"></div>
+        <div class="stars__star"></div>
+      </div>
+
+      <div class="stars--1">
+        <div class="active stars__star"></div>
```
Ihor-Karpyn, line 21: You shouldn't add class active to active stars, start should become active depending on modifier stars--n that you add to starts' wrapper

src/style.css
```diff
@@ -0,0 +1,41 @@
...
+  flex-direction: row;
+  display: flex;
+}
+
+.stars--0,
+.stars--1,
+.stars--2,
+.stars--3,
+.stars--4,
+.stars--5 {
```
Ihor-Karpyn, line 28: Add class starts to all wrappers

src/style.css
```diff
@@ -0,0 +1,41 @@
...
+  flex-direction: row;
+  display: flex;
+
+}
+
+.active {
+
+  background-image: url(images/star-active.svg);
```
Ihor-Karpyn, line 36: Redundant break line

Ihor-Karpyn (changes requested):

src/style.css
```diff
@@ -0,0 +1,44 @@
...
+.stars--0,
+.stars--1,
+.stars--2,
+.stars--3,
+.stars--4,
+.stars--5 {
+  flex-direction: row;
+  display: flex;
+
+}
```
Ihor-Karpyn, line 36:
```suggestion
stars {
  flex-direction: row;
  display: flex;
}
```

Ihor-Karpyn (changes requested):
image

src/style.css
```diff
@@ -0,0 +1,39 @@
...
+  flex-direction: row;
+  display: flex;
+}
+
+.stars {
+  flex-direction: row;
+  display: flex;
+
+}
+
```
Ihor-Karpyn, line 32:
```suggestion
.stars {
  flex-direction: row;
  display: flex;
}

```

Ihor-Karpyn (approved):

PR #2833, opened 2023-04-20
add task solution

MaksimKad:
- [DEMO LINK](https://MaksimKad.github.io/layout_stars/)
- [TEST REPORT LINK](https://MaksimKad.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,38 @@
+* {
+  box-sizing: border-box;
+}
+
+body {
+  margin: 0;
+}
+
+.stars {
+  display: flex;
+  list-style: none;
+  margin: 0;
+  padding: 0;
+}
+
+.stars__star {
+  background-image: url(images\star.svg);
+  background-repeat: no-repeat;
+  background-position: center;
+  height: 16px;
+  width: 16px;
+  margin-right: 4px;
+  position: relative;
+}
+
+.stars--1 .stars__star:nth-child(-n+1),
+.stars--2 .stars__star:nth-child(-n+2),
+.stars--3 .stars__star:nth-child(-n+3),
+.stars--4 .stars__star:nth-child(-n+4),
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,38 @@
...
+  padding: 0;
+}
+
+.stars__star {
+  background-image: url(images\star.svg);
+  background-repeat: no-repeat;
+  background-position: center;
+  height: 16px;
+  width: 16px;
+  margin-right: 4px;
```
Ihor-Karpyn, line 22: Removing margin-right from the last element and margin-left from the first element of a series of elements that are displayed inline or as a horizontal list is important because it helps to prevent unwanted spacing or gaps between the elements.

src/style.css
```diff
@@ -0,0 +1,38 @@
...
+.stars--4 .stars__star:nth-child(-n+4),
+.stars--5 .stars__star:nth-child(-n+5) {
+  background-image: url(images\star-active.svg);
+  background-repeat: no-repeat;
+  background-position: center;
+  height: 16px;
+  width: 16px;
+  position: relative;
+  margin-right: 4px;
+}
```
Ihor-Karpyn, line 38: This modifier should change only background-image, remove redundant rules or move them to `.stars__star` selector

Ihor-Karpyn (changes requested):
GJ, almost done

src/style.css
```diff
@@ -0,0 +1,36 @@
...
+
+.stars {
+  display: flex;
+  list-style: none;
+  margin: 0;
+  padding: 0;
+}
+
+ul :last-child {
```
Ihor-Karpyn, line 16: Do not use tag as selector

Ihor-Karpyn (changes requested):

Ihor-Karpyn (approved):
🔥

PR #2837, opened 2023-04-20
add task solution

PanyaPrimal:
- [DEMO LINK](https://PanyaPrimal.github.io/layout_stars/)
- [TEST REPORT LINK](https://PanyaPrimal.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -2,12 +2,6 @@
   margin: 0;
 }
 
-.stars-container {
-  width: 140px;
-  height: 160px;
-  margin-top: 2px;
-}
-
 .stars {
   display: flex;
   justify-content: center;
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,34 @@
+* {
+  margin: 0;
+}
+
+.stars-container {
+  width: 140px;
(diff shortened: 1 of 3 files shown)


naziamoff (changes requested):
not bad, but there are a couple of mistakes, that are described in the [checklist](https://github.com/mate-academy/layout_stars/blob/master/checklist.md)

Please take a look again and fix accordingly

src/style.css
```diff
@@ -0,0 +1,34 @@
+* {
+  margin: 0;
+}
+
+.stars-container {
```
naziamoff, line 5: container should not have any sizes set, let the content make its size

PanyaPrimal: i did it, but now i have 0 test completed caused by some invisible margin/top.
Check it out  😄

naziamoff: I'm not really sure, but I bet that's because you added `justify-content`, so the tests now see it differently. I don't see anything critical here though. Looks good 👍

naziamoff (approved):
nice 👍

PR #2843, opened 2023-04-20
add task solution

hekaq:
- [DEMO LINK](https://hekaq.github.io/layout_stars/)
- [TEST REPORT LINK](https://hekaq.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -4,6 +4,7 @@ body {
 
 .stars {
   display: flex;
+  list-style: none;
 
   margin: 0;
   padding: 0;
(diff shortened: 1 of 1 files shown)


src/style.css
```diff
@@ -0,0 +1,43 @@
...
+  padding: 0;
+}
+
+.stars__star {
+  width: 16px;
+  height: 16px;
+
+  margin-right: 4px;
+
+  list-style: none;
```
naziamoff, line 18: seems like a rule for the list, not for the star

src/style.css
```diff
@@ -0,0 +1,43 @@
...
+  background-image: url("./images/star-active.svg");
+}
+
+.stars--4 .stars__star:nth-child(-n+4) {
+  background-image: url("./images/star-active.svg");
+}
+
+.stars--5 .stars__star:nth-child(-n+5) {
+  background-image: url("./images/star-active.svg");
+}
```
naziamoff, line 43: those are the same, let's combine it

src/style.css
```diff
@@ -0,0 +1,43 @@
...
+
+  margin: 0;
+  padding: 0;
+}
+
+.stars__star {
+  width: 16px;
+  height: 16px;
+
+  margin-right: 4px;
```
naziamoff, line 16: should the last star have margin-right?

naziamoff (changes requested):
not bad, but there are suggestions

naziamoff (approved):
nice!

PR #2844, opened 2023-04-20
Develop

thienlaoo:
- [DEMO LINK](https://thienlaoo.github.io/layout_stars/)
- [TEST REPORT LINK](https://thienlaoo.github.io/

I have some problems with position between stars, the star size looks like in the referense, but maybe because of margins position is wrong


src/index.html
```diff
@@ -8,6 +8,170 @@
...
   </head>
   <body>
-    <h1>Stars</h1>
+    <div class="con2">
+      <div class="stars--0 stars">
+        <img
+          class="stars__star"
+          src="./images/star.svg"
+          alt="star"
+        >
```
Ihor-Karpyn, line 17: For decorative icons we use background-image

src/style.css
```diff
@@ -0,0 +1,27 @@
+body {
+  position: relative;
+  display: block;
+  margin: 0;
+  width: 140px;
+  height: 160px;
```
Ihor-Karpyn, line 6: All these rules except resetting margin redundant

src/style.css
```diff
@@ -0,0 +1,27 @@
...
+}
+
+.con2 {
```
Ihor-Karpyn, line 9: Not clear what this class name mean, maybe wrapper

src/style.css
```diff
@@ -0,0 +1,27 @@
...
+  position: absolute;
+  top: 50%;
+  left: 50%;
+  transform: translate(-50%, -50%);
+  display: inline-block;
+  height: auto;
+  width: auto;
+}
```
Ihor-Karpyn, line 17: All these styles redundant

src/index.html
```diff
@@ -8,6 +8,170 @@
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Stars</h1>
+    <div class="con2">
```
Ihor-Karpyn, line 11: Do we need this container?

src/style.css
```diff
@@ -0,0 +1,27 @@
...
+
+.stars {
+  display: flex;
+  justify-content: center;
+  height: auto;
+}
+
+.stars__star {
+  margin: 3px 4px;
+}
```
Ihor-Karpyn, line 27: Not sure that we need this margin. On the markup we have only margin-right 4px and size of stars 16px

Ihor-Karpyn (changes requested):

Anna-Tkachenko (changes requested):
Fix test report link in PR description

thienlaoo:
- [DEMO LINK](https://thienlaoo.github.io/layout_stars/)
- [TEST REPORT LINK](https://thienlaoo.github.io/layout_stars/report/html_report/)

src/style.css
```diff
@@ -0,0 +1,23 @@
...
+
+.container {
+  display: flex;
+}
+
+.star {
+  width: 16px;
+  height: 16px;
+  margin-right: 4px;
+  background: url("./images/star.svg") no-repeat;
```
Anna-Tkachenko, line 14: try to set the background image in center of the block

thienlaoo:
it looks like I need to change directions of my stars to rows, should I rework it? Because Entire_document test looks like the reference

src/index.html
```diff
@@ -7,7 +7,50 @@
     <title>Stars</title>
     <link rel="stylesheet" href="./style.css">
   </head>
-  <body>
-    <h1>Stars</h1>
+  <body class="body">
+    <div class="container">
```
Anna-Tkachenko, line 11: do you need this `container` with one style `display: block`? it's block by default, so you don't need to specify it one more time
and this container is redundant, you don't use it

Anna-Tkachenko (approved):

PR #2846, opened 2023-04-20
task solution

julshvets:
- [DEMO LINK](https://julshvets.github.io/layout_stars/)
- [TEST REPORT LINK](https://julshvets.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -10,73 +10,49 @@
   <body>
     <div class="stars stars--0">
       <span class="stars__star"></span>
-
       <span class="stars__star"></span>
-
       <span class="stars__star"></span>
-
       <span class="stars__star"></span>
-
       <span class="stars__star"></span>
     </div>
 
     <div class="stars stars--1">
       <span class="stars__star"></span>
-
       <span class="stars__star"></span>
-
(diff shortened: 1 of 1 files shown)


src/index.html
```diff
@@ -8,6 +8,76 @@
...
+    <div class="stars stars--0">
+      <span class="stars__star"></span>
+
+      <span class="stars__star"></span>
+
+      <span class="stars__star"></span>
+
+      <span class="stars__star"></span>
+
+      <span class="stars__star"></span>
```
naziamoff, line 20: That's not always when you should add empty lines between elements of the same nesting level, sometimes it is even better for readability to keep it as it is. In this case they are the same, so we don't lose anything if leave them as it is
```suggestion
      <span class="stars__star"></span>
      <span class="stars__star"></span>
      <span class="stars__star"></span>
      <span class="stars__star"></span>
      <span class="stars__star"></span>
```

naziamoff (approved):
well done! Just one small suggestion

src/index.html
julshvets: Fixed.
Thank you.

PR #2849, opened 2023-04-20
add task solution

skinnyjazz:


src/style.css
```diff
@@ -0,0 +1,54 @@
...
+
+.stars--4 .stars__star:nth-child(-n+4)::after {
+  display: block;
+  background-image: url("./images/star-active.svg");
+}
+
+.stars--5 .stars__star:nth-child(-n+5)::after {
+  display: block;
+  background-image: url("./images/star-active.svg");
+} */
```
naziamoff, line 54: all the redundant comments should be removed

src/style.css
```diff
@@ -0,0 +1,54 @@
...
+}
+
+.stars__star::after {
+  content: "";
+  display: block;
+  width: 12px;
+  height: 12px;
+  background-image: url("./images/star.svg");
+  background-size: cover;
+}
```
naziamoff, line 25: you can set background-image just for the star, that is above

naziamoff (changes requested):
Also, see this one:
https://mateacademy.slack.com/archives/C0506RS8F62/p1681906726598509

silchyn (approved):

naziamoff (approved):

PR #2855, opened 2023-04-21
add stars rate

natalia-serogina:
https://natalia-serogina.github.io/layout_stars/
https://natalia-serogina.github.io/layout_stars/report/html_report/

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,46 @@
+html body {
+  margin: 0;
+}
+
+* {
+  box-sizing: border-box;
+}
+
+.stars {
+  display: flex;
+}
+
+.star__star {
+  margin: 0 2px;
+  width: 16px;
+  height: 16px;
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,46 @@
...
+  background-image: url(./images/star-active.svg);
+}
+
+.stars--4 :nth-child(-n+4) {
+  background-image: url(./images/star-active.svg);
+}
+
+.stars--5 :nth-child(-n+5) {
+  background-image: url(./images/star-active.svg);
+}
```
OleziO, line 46: Combine it with one selector:

```
.stars--1 :nth-child(-n+1),
.stars--2 :nth-child(-n+2),
.stars--3 :nth-child(-n+3),
.stars--4 :nth-child(-n+4),
.stars--5 :nth-child(-n+5) {
}
```

OleziO (changes requested):

PR #2858, opened 2023-04-21
add task solution

LeonIllya:
- [DEMO LINK](https://LeonIllya.github.io/layout_stars/)
- [TEST REPORT LINK](https://LeonIllya.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -8,6 +8,54 @@
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Stars</h1>
+    <div class="container">
+      <ul class="stars stars--0">
+        <li class="stars__star"></li>
+        <li class="stars__star"></li>
+        <li class="stars__star"></li>
+        <li class="stars__star"></li>
+        <li class="stars__star"></li>
+      </ul>
+
+      <ul class="stars stars--1">
+        <li class="stars__star"></li>
+        <li class="stars__star"></li>
+        <li class="stars__star"></li>
+        <li class="stars__star"></li>
+        <li class="stars__star"></li>
+      </ul>
+
+      <ul class="stars stars--2">
+        <li class="stars__star"></li>
+        <li class="stars__star"></li>
+        <li class="stars__star"></li>
+        <li class="stars__star"></li>
+        <li class="stars__star"></li>
+      </ul>
+
+      <ul class="stars stars--3">
+        <li class="stars__star"></li>
+        <li class="stars__star"></li>
+        <li class="stars__star"></li>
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,43 @@
...
+    
+.stars {
+  display: flex;
+  list-style: none;
+}
+    
+.stars__star {
+  width: 16px;
+  height: 16px;
+  margin-right: 4px;
```
Ihor-Karpyn, line 18: Removing margin-right from the last element and margin-left from the first element of a series of elements that are displayed inline or as a horizontal list is important because it helps to prevent unwanted spacing or gaps between the elements.

src/style.css
```diff
@@ -0,0 +1,43 @@
...
+  background-image: url(images/star-active.svg);
+}
+        
+.stars--4 > .stars__star:nth-child(-n+4) {
+  background-image: url(images/star-active.svg);
+}
+  
+.stars--5 > .stars__star:nth-child(-n+5) {
+  background-image: url(images/star-active.svg);
+}
```
Ihor-Karpyn, line 43: Group rules using `,`

Ihor-Karpyn (changes requested):
GJ, almost done

src/index.html
```diff
@@ -8,6 +8,54 @@
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Stars</h1>
+    <div class="container">
+      <ul class="stars stars--0">
```
Anna-Tkachenko, line 12: Good job with using semantic tags 🔥

src/style.css
```diff
@@ -0,0 +1,40 @@
...
+  margin-right: 4px;
+  
+  background-image: url(images/star.svg);
+  background-repeat: no-repeat;
+  background-position: center;
+}
+
+.stars__star:first-child {
+  margin-left: 0;
+}
```
Anna-Tkachenko, line 28: you don't set any margin-left for the `stars__star` elements, do you need to reset it here?

src/style.css
```diff
@@ -0,0 +1,40 @@
...
+  list-style: none;
+  margin-right: 0;
+}
+    
+.stars__star {
+  width: 16px;
+  height: 16px;
+  margin-right: 4px;
+  
+  background-image: url(images/star.svg);
```
Anna-Tkachenko, line 21: will be better to wrap with quoter the url for the image 
```suggestion
  background-image: url("images/star.svg");
```

Anna-Tkachenko (approved):
Good job!

PR #2861, opened 2023-04-21
add task solution

Huracann:
- [DEMO LINK](https://Huracann.github.io/layout_stars/)
- [TEST REPORT LINK](https://Huracann.github.io/layout_stars/report/html_report/)


silchyn (changes requested):
There is right margin for last star

33Orange (approved):
Well done!

PR #2863, opened 2023-04-22
add task solution

VVBanas:
- [DEMO LINK](https://VVBanas.github.io/layout_stars/)
- [TEST REPORT LINK](https://VVBanas.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -8,6 +8,54 @@
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Stars</h1>
+    <ul class="stars stars--0">
+      <li class="stars__star"></li>
+      <li class="stars__star"></li>
+      <li class="stars__star"></li>
+      <li class="stars__star"></li>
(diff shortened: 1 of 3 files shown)


src/index.html
```diff
@@ -8,6 +8,54 @@
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Stars</h1>
+    <ul class="stars stars--0">
```
Anna-Tkachenko, line 11: Good job with using semantic tags 🔥

src/style.css
```diff
@@ -0,0 +1,32 @@
+body {
+  margin: 0;
+}
+
+.stars {
+  display: flex;
+  list-style: none;
+  margin: 0;
+  padding: 0;
```
Anna-Tkachenko, line 9: You can also reset these styles for the `ul` tag directly as you did id for `body`

Anna-Tkachenko (approved):
Good job! ⭐️

PR #2891, opened 2023-05-01
add task layout_stars

P-Nazar:
https://P-Nazar.github.io/layout_stars/
https://P-Nazar.github.io/layout_stars/report/html_report/

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -8,6 +8,52 @@
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Stars</h1>
+    <ul class="stars stars--0">
+      <li class="stars__star"></li>
+      <li class="stars__star"></li>
+      <li class="stars__star"></li>
+      <li class="stars__star"></li>
+      <li class="stars__star"></li>
+    </ul>
+
+    <ul class="stars stars--1">
+      <li class="stars__star"></li>
+      <li class="stars__star"></li>
+      <li class="stars__star"></li>
+      <li class="stars__star"></li>
+      <li class="stars__star"></li>
+    </ul>
+
+    <ul class="stars stars--2">
+      <li class="stars__star"></li>
+      <li class="stars__star"></li>
+      <li class="stars__star"></li>
+      <li class="stars__star"></li>
+      <li class="stars__star"></li>
+    </ul>
+
+    <ul class="stars stars--3">
+      <li class="stars__star"></li>
+      <li class="stars__star"></li>
+      <li class="stars__star"></li>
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,66 @@
+* {
+  margin: 0;
+  padding: 0;
+  text-decoration: none;
+  list-style: none;
+  box-sizing: border-box;
```
VitaliyBondarenko1982, line 6:
```suggestion
  box-sizing: border-box;
```
* selector use only box-sizing.

reset default styles only for tags which necessary

src/style.css
```diff
@@ -0,0 +1,66 @@
...
+  background: url(./images/star-active.svg)  no-repeat;
+  background-position: center;
+}
+
+.stars--5 .stars__star:nth-child(5) {
+  width: 16px;
+  height: 16px;
+  background: url(./images/star-active.svg)  no-repeat;
+  background-position: center;
+}
```
VitaliyBondarenko1982, line 66:
```suggestion
.stars--1 .stars__star:nth-child(-n+1),
.stars--2 .stars__star:nth-child(-n+2),
.stars--3 .stars__star:nth-child(-n+3),
.stars--4 .stars__star:nth-child(-n+4),
.stars--5 .stars__star:nth-child(-n+5) {
  width: 16px;
  height: 16px;
  background: url(./images/star-active.svg)  no-repeat;
  background-position: center;
}
```

no need such hardcode and tone of code,
check another way...

src/index.html
```diff
@@ -8,6 +8,52 @@
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Stars</h1>
+    <ul class="stars stars--0">
+      <li class="stars__star"></li>
```
VitaliyBondarenko1982, line 12:
```suggestion
     <div class="stars stars--0">
      <div class="stars__star"></div>
```
it is not list
in this case better to make div.
also will be no need to reset list styles

src/style.css
```diff
@@ -0,0 +1,66 @@
...
+  display: flex;
+  justify-content: center;
+  align-items: center;
+}
+
+.stars__star {
+  width: 16px;
+  height: 16px;
+  background: url(./images/star.svg) no-repeat;
+  margin-right: 4px;
```
VitaliyBondarenko1982, line 19: do not set margin right for last star in row

VitaliyBondarenko1982 (changes requested):

IvanFesenko (approved):
Check comment to avoid same mistakes in the future

PR #2895, opened 2023-05-02
Developing...

timohahaaa:
https://timohahaaa.github.io/layout_stars/report/html_report/

https://timohahaaa.github.io/layout_stars/

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -15,38 +15,10 @@ body {
   background-position: center;
 }
 
-.stars--1 span:first-child {
-  width: 16px;
-  height: 16px;
-  background-image: url(images/star-active.svg);
-  background-repeat: no-repeat;
-  background-position: center;
-}
-
-.stars--2 span:nth-child(-n+2) {
-  width: 16px;
-  height: 16px;
-  background-image: url(images/star-active.svg);
-  background-repeat: no-repeat;
-  background-position: center;
-}
-
-.stars--3 span:nth-child(-n+3) {
-  width: 16px;
-  height: 16px;
-  background-image: url(images/star-active.svg);
-  background-repeat: no-repeat;
-  background-position: center;
-}
-
-.stars--4 span:nth-child(-n+4) {
-  width: 16px;
-  height: 16px;
-  background-image: url(images/star-active.svg);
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,56 @@
...
+  background-position: center;
+}
+
+.stars--5 span:nth-child(-n+5) {
+  width: 16px;
+  height: 16px;
+  background-image: url(images/star-active.svg);
+  background-repeat: no-repeat;
+  background-position: center;
+}
```
qoa11a, line 56: You should move it under one selector like this:
`.stars--1 span:first-child, .stars--2 span:nth-child(-n+2), ... { ...styles go here }`

qoa11a (changes requested):
Almost perfect! One small fix.

src/style.css
```diff
@@ -0,0 +1,28 @@
...
+  background-image: url(images/star.svg);
+  background-repeat: no-repeat;
+  background-position: center;
+}
+
+.stars--1 span:first-child,
+.stars--2 span:nth-child(-n+2),
+.stars--3 span:nth-child(-n+3),
+.stars--4 span:nth-child(-n+4),
+.stars--5 span:nth-child(-n+5) {
```
polosanya, line 22:
```suggestion
.stars--1 :first-child,
.stars--2 :nth-child(-n+2),
.stars--3 :nth-child(-n+3),
.stars--4 :nth-child(-n+4),
.stars--5 :nth-child(-n+5) {
```
There is no need to increase the specificity

src/style.css
```diff
@@ -0,0 +1,28 @@
...
+  width: 16px;
+  height: 16px;
+  background-image: url(images/star-active.svg);
+  background-repeat: no-repeat;
+  background-position: center;
```
polosanya, line 27: Do you really need to duplicate all styles? Only background image changes, all other properties are the same

polosanya (changes requested):
A few small, but important changes left 🙂

polosanya (approved):
Well done! 🔥

polosanya (approved):
Super!

qoa11a (approved):

PR #2898, opened 2023-05-03
add task solution

Vladyslav-amp:


src/index.html
```diff
@@ -8,6 +8,52 @@
...
+      <div class="starsArr">
+        <div class="stars--0">
+          <div class="stars__star"></div>
+          <div class="stars__star"></div>
+          <div class="stars__star"></div>
+          <div class="stars__star"></div>
+          <div class="stars__star"></div>
+        </div>
+        <div class="stars--1">
+          <div class="stars--N"></div>
```
slavanosatskyi, line 21: it should be just starts__star, colored stars should be controlled by parent class stars--N, where N is a number of colored stars 
Screenshot 2023-05-03 at 17 15 03

slavanosatskyi (changes requested):

src/style.css
```diff
@@ -0,0 +1,31 @@
+* {
+  margin: 0;
+}
+
+.stars--0,
```
OleziO, line 5: Create a common starts class, where there will be common styles, and attach it to each block

OleziO (changes requested):

PR #2927, opened 2023-05-10
add task solution

dimkamg21:
https://dimkamg21.github.io/layout_stars/
https://dimkamg21.github.io/layout_stars/report/html_report/

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -8,6 +8,72 @@
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Stars</h1>
+    <div class="stars stars--0">
+      <div class="stars__star"></div>
+
+      <div class="stars__star"></div>
+
+      <div class="stars__star"></div>
+
+      <div class="stars__star"></div>
+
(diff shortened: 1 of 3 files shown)


src/index.html
```diff
@@ -8,6 +8,72 @@
...
+      
+      <div class="stars__star"></div>
+      
+      <div class="stars__star"></div>
+      
+      <div class="stars__star"></div>
+      
+      <div class="stars__star"></div>
+    </div>
+    </div>
```
etojeDenys, line 77: There is an extra closing 'div' tag. Please remove it.

src/style.css
```diff
@@ -0,0 +1,32 @@
+html {
```
etojeDenys, line 1: It's better to apply the margin reset to the 'body' element, not the 'html' element.

etojeDenys (changes requested):

etojeDenys:
> https://dimkamg21.github.io/layout_stars/ https://dimkamg21.github.io/layout_stars/report/html_report/

your links do not work

PR #2928, opened 2023-05-10
layout stars

oltarasov:
- [DEMO LINK](https://oltarasov.github.io/layout_stars/)
- [TEST REPORT LINK](https://oltarasov.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,42 @@
+body {
+  margin: 0;
+}
+
+.stars {
+  display: flex;
+  align-items: center;
+}
+
+.stars_star {
+  background-image: url(images/star.svg);
+  background-repeat: no-repeat;
+  background-position: center;
+  width: 16px;
+  height: 16px;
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,42 @@
...
+  background-repeat: no-repeat;
+  background-position: center;
+  width: 16px;
+  height: 16px;
+  margin-right: 4px;
+}
+
+.stars--1 .stars_star:first-child {
+  background-image: url(images/star-active.svg);
+  background-repeat: no-repeat;
```
vlad-demchuk, line 21: 1. you already have `background-repeat: no-repeat;` in `stars_star`
2. combine all selectors using coma
```
.stars--1 .stars_star:first-child,
.stars--2 .stars_star:nth-child(-n + 2) {
}

vlad-demchuk (changes requested):

PR #2945, opened 2023-05-14
add layout_stars solution

Tvinki228:
https://Tvinki228.github.io/layout_stars/
https://Tvinki228.github.io/layout_stars/report/html_report/

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -16,13 +16,11 @@ body {
   background-position: center center;
 }
 
+.stars--0 .stars__star:nth-child(-n+0),
 .stars--1 .stars__star:nth-child(-n+1),
 .stars--2 .stars__star:nth-child(-n+2),
 .stars--3 .stars__star:nth-child(-n+3),
 .stars--4 .stars__star:nth-child(-n+4),
 .stars--5 .stars__star:nth-child(-n+5) {
   background-image: url(images/star-active.svg);
-  width: 16px;
-  background-repeat: no-repeat;
-  height: 16px;
 }


src/style.css
```diff
@@ -0,0 +1,24 @@
+* {
+  margin: 0;
```
vlad-demchuk, line 2: do not reset margin for `*` selector

src/index.html
```diff
@@ -8,6 +8,84 @@
...
+        <div class="stars__empty">
+        </div>
+        <div class="stars__empty">
+        </div>
+        <div class="stars__empty">
+        </div>
+        <div class="stars__empty">
+        </div>
+        <div class="stars__empty">
+        </div>
```
vlad-demchuk, line 22: image
you should implement count of active stars with css styles

vlad-demchuk (changes requested):

src/style.css
```diff
@@ -0,0 +1,28 @@
...
+  background-position: center center;
+}
+
+.stars--1 .stars__star:nth-child(-n+1),
+.stars--2 .stars__star:nth-child(-n+2),
+.stars--3 .stars__star:nth-child(-n+3),
+.stars--4 .stars__star:nth-child(-n+4),
+.stars--5 .stars__star:nth-child(-n+5) {
+  background-image: url(images/star-active.svg);
+  width: 16px;
```
volodymyr-soltys97, line 25: There is no need to set the width and height properties again in this block since they are already set in the .stars__star class. You can remove lines 25 and 27.

src/style.css
```diff
@@ -0,0 +1,28 @@
...
+.stars__star {
+  background-image: url(images/star.svg);
+  width: 16px;
+  height: 16px;
+  margin-right: 4px;
+  background-repeat: no-repeat;
+  background-position: center center;
+}
+
+.stars--1 .stars__star:nth-child(-n+1),
```
volodymyr-soltys97, line 19: You missed the .stars--0 selector. Add .stars--0 .stars__star:nth-child(-n+0), to the list of selectors to cover all the cases from 0 to 5 active stars.

volodymyr-soltys97 (changes requested):
Good job 👍 
But let's make your code more sophisticated

vlad-demchuk (approved):

PR #2953, opened 2023-05-16
add task solution

ihodyk:
[DEMO LINK](https://ihodyk.github.io/layout_stars/)
[TEST REPORT LINK](https://ihodyk.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,4 +1,3 @@
-html,
 body {
   margin: 0;
 }


src/style.css
```diff
@@ -0,0 +1,37 @@
+html {
```
volodymyr-soltys97, line 1: The 'html' selector should be replaced with 'body' as the margin property does not apply to the 'html' element.

volodymyr-soltys97 (changes requested):
Good job
But let's solve the problems with the tests

src/style.css
```diff
@@ -0,0 +1,36 @@
...
+}
+
+.stars.stars--1 .stars__star:nth-child(-n+1),
+.stars.stars--2 .stars__star:nth-child(-n+2),
+.stars.stars--3 .stars__star:nth-child(-n+3),
+.stars.stars--4 .stars__star:nth-child(-n+4),
+.stars.stars--5 .stars__star:nth-child(-n+5) {
+  background: url(./images/star-active.svg);
+  background-repeat: no-repeat;
+  background-position: center;
```
volodymyr-soltys97, line 35: There is no need to duplicate the code here, if these properties already exist above
```suggestion
```

volodymyr-soltys97 (approved):
Great 🔥

PR #2954, opened 2023-05-16
add task solution

MakaronX:
https://MakaronX.github.io/layout_stars/
https://MakaronX.github.io/layout_stars/report/html_report/

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,25 @@
+body {
+  margin: 0;
+  --background-image: url(./images/star-active.svg);
+}
+
+.stars {
+  display: flex;
+}
+
+.stars__star {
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,25 @@
...
+  display: flex;
+}
+
+.stars__star {
+  background-image: url(./images/star.svg);
+  background-repeat: no-repeat;
+  background-position: center;
+  height: 16px;
+  width: 16px;
+  margin-right: 4px;
```
mykhalenych, line 16: the last star shouldn't have margin-right

mykhalenych (approved):

PR #2958, opened 2023-05-17
Solution

offsset:
https://offsset.github.io/layout_stars/


mykhalenych (changes requested):
Your test link isn't working

AlisaVlasova (approved):

PR #2962, opened 2023-05-18
Make a stars

Zhekich123:
DEMO LINK: https://zhekich123.github.io/layout_stars/
REPORT LINK: https://Zhekich123.github.io/layout_stars/report/html_report/

I faled 6 tests, but I have no idea why this blocks with stars have a difference with a reference.. 
I tried to move all blocks in the left side for 1-3px with margin, padding and just position: relative; right: 2px; and blocks is move successfully but difference is the same..


src/style.css
```diff
@@ -0,0 +1,67 @@
...
+
+.stars-frame {
+  width: 96px;
+  height: 96px;
+}
+
+.stars {
+  display: flex;
+  flex-wrap: wrap;
+  justify-content: space-between;
```
IvanFesenko, line 25: Add right margin for stars except last one

src/style.css
```diff
@@ -0,0 +1,67 @@
...
+  background-image: url(./images/star-active.svg);
+  background-position: center;
+  background-repeat: no-repeat;
+}
+
+.stars--5 .stars__star:nth-child(-n+5) {
+  background-image: url(./images/star-active.svg);
+  background-position: center;
+  background-repeat: no-repeat;
+}
```
IvanFesenko, line 67: Do not repeat your self!
```suggestion
.stars--1 .stars__star:first-child,
.stars--2 .stars__star:nth-child(-n+2),
.stars--3 .stars__star:nth-child(-n+3),
.stars--4 .stars__star:nth-child(-n+4),
.stars--5 .stars__star:nth-child(-n+5) {
  background-image: url(./images/star-active.svg);
}
```

IvanFesenko (changes requested):
Check my comments

src/style.css
Zhekich123: > Add right margin for stars except last one

I don't understand, I trying to add but it's still wrong..

PR #2963, opened 2023-05-18
add star solution

Medaed:
- [DEMO LINK](https://Medaed.github.io/layout_stars/)
- [TEST REPORT LINK](https://Medaed.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -55,6 +55,5 @@
       <span class="stars__star"></span>
       <span class="stars__star"></span>
     </div>
-
   </body>
 </html>
--- a/src/style.css
+++ b/src/style.css
@@ -1,4 +1,3 @@
-
 body {
   margin: 0;
 }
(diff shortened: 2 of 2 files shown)


src/style.css
```diff
@@ -0,0 +1,32 @@
...
+  display: flex;
+}
+
+.stars--0 .stars__star,
+.stars--1 .stars__star,
+.stars--2 .stars__star,
+.stars--3 .stars__star,
+.stars--4 .stars__star,
+.stars--5 .stars__star {
+  display: flex;
```
polosanya, line 16:
```suggestion
.stars__star {
  display: flex;
```
There is no need to list all modifiers, you can set all default styles just for `stars__star` class and only change background image for modifiers

src/index.html
```diff
@@ -8,6 +8,53 @@
...
+
+    <div class="stars stars--5">
+      <span class="stars__star"></span>
+      <span class="stars__star"></span>
+      <span class="stars__star"></span>
+      <span class="stars__star"></span>
+      <span class="stars__star"></span>
+    </div>
+
   </body>
```
polosanya, line 59:
```suggestion
    </div>
  </body>
```

src/style.css
```diff
@@ -0,0 +1,32 @@
+
+body {
```
polosanya, line 2:
```suggestion
body {
```

polosanya (changes requested):
Almost perfect!

anstsot (approved):
Good job, approved!

Medaed:
as i correctly understand i should re-request review then it would updated on site

SanyaBratashchuk (approved):

PR #2967, opened 2023-05-18
task 2: layout stars

TVChulkova:
[DEMO LINK]   https://tvchulkova.github.io/layout_stars/
[TEST REPORT LINK]  https://tvchulkova.github.io/layout_stars/report/html_report/

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -3,21 +3,11 @@
 }
 
 body {
-  background-color: #e5e5e5;
   margin: 0;
-}
-
-.container {
-  width: 100px;
-  height: 106px;
-  display: flex;
-  flex-direction: column;
-  background-color: #fff;
 }
 
 .stars {
   display: flex;
-  justify-content: space-between;
 }
 
 .stars__star {
--- a/src/index.html
+++ b/src/index.html
@@ -8,54 +8,52 @@
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <div class="container">
-      <div class="stars stars--0">
-        <div class="stars__star"></div>
-        <div class="stars__star"></div>
-        <div class="stars__star"></div>
-        <div class="stars__star"></div>
-        <div class="stars__star"></div>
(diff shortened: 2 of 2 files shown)


src/style.css
```diff
@@ -0,0 +1,56 @@
...
+  background-color: #e5e5e5;
+  margin: 0;
+}
+
+.container {
+  width: 100px;
+  height: 106px;
+  display: flex;
+  flex-direction: column;
+  background-color: #fff;
```
vlad-demchuk, line 15: no need to create this container

src/style.css
```diff
@@ -0,0 +1,56 @@
...
+  width: 16px;
+  height: 0;
+  padding-bottom: 14px;
+  margin: 2px;
+  margin-bottom: 0;
+  background-repeat: no-repeat;
+}
+
+.stars:nth-child(-n + 3) .stars__star {
+  background-image: url(./images/star.svg);
```
vlad-demchuk, line 33: make all stars with `background-image: url(./images/star.svg);`

and just changs active stars with separate selectors

vlad-demchuk (changes requested):

src/style.css
```diff
@@ -0,0 +1,28 @@
...
+}
+
+.stars {
+  display: flex;
+}
+
+.stars__star {
+  width: 16px;
+  height: 0;
+  padding-bottom: 14px;
```
nazarmatsevych, line 16: Redundant padding, we can use background-position: center instead

src/style.css
```diff
@@ -0,0 +1,28 @@
...
+  margin: 2px 2px 0;
```
nazarmatsevych, line 17: You can only add margin-right: 4px like in mockup and remove margin for the last element in a row

nazarmatsevych (changes requested):

nazarmatsevych (approved):

PR #3011, opened 2023-05-24
add task solution

unholysearcher:
https://unholysearcher.github.io/layout_stars/
https://unholysearcher.github.io/layout_stars/report/html_report/

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,40 @@
+* {
+  margin: 0;
+}
+
+.stars {
+  display: flex;
+}
+
+.stars__star {
+  width: 16px;
+  height: 16px;
+  margin-right: 4px;
+  background-image: url("../src/images/star.svg");
+  background-repeat: no-repeat;
+  background-position: center;
+}
+
+.stars__star:last-child {
+  margin-right: 0;
+}
+
+.stars--1 .stars__star:first-child {
+  background-image: url("../src/images/star-active.svg");
+}
+
+.stars--2 .stars__star:nth-child(-n+2) {
+  background-image: url("../src/images/star-active.svg");
+}
+
+.stars--3 .stars__star:nth-child(-n+3) {
+  background-image: url("../src/images/star-active.svg");
+}
+
+.stars--4 .stars__star:nth-child(-n+4) {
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,40 @@
...
+  margin-right: 0;
+}
+
+.stars--1 .stars__star:first-child {
+  background-image: url("../src/images/star-active.svg");
+}
+
+.stars--2 .stars__star:nth-child(-n+2) {
+  background-image: url("../src/images/star-active.svg");
+}
```
TarasHoliuk, line 28:
```suggestion
.stars--1 .stars__star:first-child,
.stars--2 .stars__star:nth-child(-n+2) {
  background-image: url("../src/images/star-active.svg");
}
```
you may list all selectors (if you add the same style for each one) to make it look more laconic

TarasHoliuk: combine all such selectors

TarasHoliuk (changes requested):

src/style.css
```diff
@@ -0,0 +1,28 @@
...
+  background-image: url("../src/images/star.svg");
+  background-repeat: no-repeat;
+  background-position: center;
+}
+
+.stars__star:last-child {
+  margin-right: 0;
+}
+
+.stars--1 .stars__star:first-child,
```
OleziO, line 22: Replace with `.stars--1 .stars__star:nth-child(-n+1)` to make code at the same style

OleziO (changes requested):

src/style.css
```diff
@@ -0,0 +1,28 @@
+* {
+  margin: 0;
+}
```
DarkMistyRoom, line 3: Don't use * selector for zeroing out your margins or paddings. It's still inefficient for browser to read your web document

DarkMistyRoom (changes requested):
Good job! Only one comment, but it is important

SerhiiKirik (approved):
Great 🔥

SanyaBratashchuk (approved):

loralevitska (approved):

PR #3017, opened 2023-05-25
add task solution

Teerveet:


src/style.css
```diff
@@ -0,0 +1,30 @@
+body {
+  margin: 0;
+  /* box-sizing: border-box; */
+}
+
+.stars {
+  display: flex;
+  flex-direction: row;
```
mykhalenych, line 8: when you use display: flex flex-direction: row will be by default

mykhalenych (approved):
don't forget add demo and test links to description

PR #3018, opened 2023-05-25
Develop

IgorOksentyuk:
- [DEMO LINK](https://igoroksentyuk.github.io/layout_stars/)
- [TEST REPORT LINK](https://igoroksentyuk.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -8,6 +8,51 @@
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Stars</h1>
+    <div class="stars stars--0">
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+    </div>
+
+    <div class="stars stars--1">
+      <div class="stars__star star--active"></div>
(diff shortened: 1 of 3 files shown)


src/index.html
```diff
@@ -8,6 +8,51 @@
...
+      <div class="stars__star star--active"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+    </div>
+
+    <div class="stars stars--3">
+      <div class="stars__star star--active"></div>
+      <div class="stars__star star--active"></div>
+      <div class="stars__star star--active"></div>
```
ManoylovAC, line 38: It is cheating)
You do not need to create an active class for specific `stars__star` items. In the conditions of this task, they should be highlighted under the influence of the parent modifier such as `stars--3`

ManoylovAC (changes requested):
Hey 👋 
The stars look as they should, but they should be made in a different way.
See my comment below.
Good luck!

image

src/index.html
IgorOksentyuk: Is this solution better?)

tetiana-ivanchenko (approved):

PR #3034, opened 2023-05-29
add task solution

pavlokuraksin:
- [DEMO LINK](https://CasioPayYeah.github.io/layout_stars/)
- [TEST REPORT LINK](https://CasioPayYeah.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,29 @@
+html,
+body {
+  margin: 0;
+}
+
+.stars {
+  display: flex;
+}
+
+.stars__star {
+  height: 16px;
+  width: 16px;
+  background-image: url(images/star.svg);
+  background-repeat: no-repeat;
+  background-position: center;
+}
+
+.stars__star:not(:last-child) {
+  margin-right: 4px;
+}
+
+.stars--1 .stars__star:nth-child(-n + 1),
+.stars--2 .stars__star:nth-child(-n + 2),
+.stars--3 .stars__star:nth-child(-n + 3),
+.stars--4 .stars__star:nth-child(-n + 4),
+.stars--5 .stars__star:nth-child(-n + 5),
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,29 @@
...
+  background-image: url(images/star.svg);
+  background-repeat: no-repeat;
+  background-position: center;
+}
+
+.stars__star:not(:last-child) {
+  margin-right: 4px;
+}
+
+.stars--1 .stars__star:nth-child(-n + 1),
```
ManoylovAC, line 22: hint: in this case you also can use direct number or just `:first-child`
for the symmetry it looks cool but every formula with `n` adds some small but additional calculation to you page

src/style.css
```diff
@@ -0,0 +1,29 @@
...
+.stars--2 .stars__star:nth-child(-n + 2),
+.stars--3 .stars__star:nth-child(-n + 3),
+.stars--4 .stars__star:nth-child(-n + 4),
+.stars--5 .stars__star:nth-child(-n + 5),
+.stars--6 .stars__star:nth-child(-n + 6) {
```
ManoylovAC, line 27: I can't see any block with `.stars--6` in html
if you no need you can delete this

ManoylovAC (changes requested):
Very good and almost done 
Please remove extra rule you don't really used

src/style.css
CasioPayYeah: What about :last-child, it can be used to?

CasioPayYeah: So, i fixed it and remove last pseudo-element to. Now it looks like just '.stars--1 .stars__star'.

ManoylovAC (approved):
Good work and improvement for selectors 👍

PR #3036, opened 2023-05-29
add task solution

AnastasiiaKhemych:
- [DEMO LINK](https://AnastasiiaKhemych.github.io/layout_stars/)
- [TEST REPORT LINK](https://AnastasiiaKhemych.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -8,6 +8,52 @@
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Stars</h1>
+    <div class="stars stars--0">
+      <div class="stars__star stars__passive"></div>
+      <div class="stars__star stars__passive"></div>
+      <div class="stars__star stars__passive"></div>
+      <div class="stars__star stars__passive"></div>
+      <div class="stars__star stars__passive"></div>
+    </div>
+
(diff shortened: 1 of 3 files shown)


src/index.html
```diff
@@ -8,6 +8,52 @@
...
+      <div class="stars__star stars__passive"></div>
+      <div class="stars__star stars__passive"></div>
+      <div class="stars__star stars__passive"></div>
+      <div class="stars__star stars__passive"></div>
+      <div class="stars__star stars__passive"></div>
+    </div>
+
+    <div class="stars stars--1">
+      <div class="stars__star stars__active"></div>
+      <div class="stars__star stars__passive"></div>
```
anstsot, line 21: do not `active` or `passive` modifier class for star. rely only on parent `.stars--N` class

anstsot (changes requested):
Good work, but that `active` os `passive` class should not be used.
Instead, you should rely only on `stars--N` class. You can use `:nth-child` selector - [docs](https://developer.mozilla.org/en-US/docs/Web/CSS/:nth-child).

YegorKorenkov (approved):
Looks good!

PR #3040, opened 2023-05-29
Add Stars solution

NataliGru:
- [DEMO LINK](https://NataliGru.github.io/layout_stars/)
- [TEST REPORT LINK](https://NataliGru.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -17,7 +17,7 @@
     </div>
 
     <div class="stars stars--1">
-      <div class="stars__star star--active"></div>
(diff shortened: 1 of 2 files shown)


anstsot (changes requested):
Instead, you should rely only on `stars--N` class. Uou can use `:nth-child` selector - [docs](https://developer.mozilla.org/en-US/docs/Web/CSS/:nth-child).

anstsot (approved):

PR #3046, opened 2023-05-29
add task solution

dianadiv:
- [DEMO LINK](https://dianadiv.github.io/layout_stars/)
- [TEST REPORT LINK](https://dianadiv.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -8,6 +8,49 @@
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Stars</h1>
+    <div class="container">
+      <div class="stars stars--0">
+        <div class="stars__star"></div>
+        <div class="stars__star"></div>
+        <div class="stars__star"></div>
+        <div class="stars__star"></div>
+        <div class="stars__star"></div>
+      </div>
+      <div class="stars stars--1">
+        <div class="stars__star colored"></div>
+        <div class="stars__star"></div>
+        <div class="stars__star"></div>
+        <div class="stars__star"></div>
+        <div class="stars__star"></div>
+      </div>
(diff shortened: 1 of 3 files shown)


And678 (changes requested):
Looks good, but that `colored` class should not be used.
Instead you should rely only on `stars--N` classes. For that you can use `:nth-child` selector - [docs](https://developer.mozilla.org/en-US/docs/Web/CSS/:nth-child).

anstsot (changes requested):
Good work:four_leaf_clover:
Let’s fix the code according to the checklist

src/index.html
```diff
@@ -8,6 +8,78 @@
...
+        <div class="stars__star"></div>
+
+        <div class="stars__star"></div>
+
+        <div class="stars__star"></div>
+        
+        <div class="stars__star"></div>
+      </div>
+
+      <div class="stars stars--1">
```
tetiana-ivanchenko, line 24:
```suggestion
      <div class="stars stars--0">
        <div class="stars__star"></div>
        <div class="stars__star"></div>
        <div class="stars__star"></div>
        <div class="stars__star"></div>
        <div class="stars__star"></div>
      </div>

      <div class="stars stars--1">
```

src/style.css
```diff
@@ -0,0 +1,32 @@
+body {
+  margin: 0;
+}
+
+.container {
+  margin: 0;
+}
```
tetiana-ivanchenko, line 7: redundant, div hasn't default margins

tetiana-ivanchenko (changes requested):

anstsot (approved):

PR #3047, opened 2023-05-29
add task solution

MariiaMonakhova:
- [DEMO LINK](https://MariiaMonakhova.github.io/layout_stars/)
- [TEST REPORT LINK](https://MariiaMonakhova.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -8,6 +8,47 @@
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Stars</h1>
+    <div class="stars stars--0">
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+    </div>
+    <div class="stars stars--1">
+      <div class="stars__star star--active"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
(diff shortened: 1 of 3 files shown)


src/index.html
```diff
@@ -8,6 +8,47 @@
...
+      <div class="stars__star star--active"></div>
+      <div class="stars__star"></div>
+    </div>
+    <div class="stars stars--5">
+      <div class="stars__star star--active"></div>
+      <div class="stars__star star--active"></div>
+      <div class="stars__star star--active"></div>
+      <div class="stars__star star--active"></div>
+      <div class="stars__star star--active"></div>
+    </div>
```
And678, line 52: Using `stars--active` is not a correct solution. You should rely only on `stars--5` in this case. For that you can use `:nth-child` selector.
```suggestion
    <div class="stars stars--5">
      <div class="stars__star"></div>
      <div class="stars__star"></div>
      <div class="stars__star"></div>
      <div class="stars__star"></div>
      <div class="stars__star"></div>
    </div>
```

And678 (changes requested):
Looks nice, but check out my comment below, you shouldn't use this stars--active. Also, check the [checklist #1](https://github.com/mate-academy/layout_stars/blob/master/checklist.md).

src/index.html
MariiaMonakhova: Done! Thank you for your feedback

tetiana-ivanchenko (approved):
GJ 👍

PR #3057, opened 2023-05-30
add task solution

KandaskalovOleg:
- [DEMO LINK](https://kandaskalovoleg.github.io/layout_stars/)
- [TEST REPORT LINK](https://kandaskalovoleg.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,28 @@
+body {
+  margin: 0;
+  padding: 0;
+}
+
+.stars {
+  display: flex;
+}
+
+.stars__star {
+  background-image: url("./images/star.svg");
+  background-repeat: no-repeat;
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,28 @@
...
+  width: 16px;
+  height: 16px;
+  background-position: center;
+}
+
+.stars__star:not(:last-child) {
+  margin-right: 4px;
+}
+
+.stars--5 .stars__star,
```
anstsot, line 22: it will be better to add proper `:nth-child` because what if you’ll have more than 5 stars children but the number of active stars will be `5`?

anstsot (approved):

PR #3059, opened 2023-05-30
STARS

FlameInBrains:
- [DEMO LINK](https://FlameInBrains.github.io/layout_stars/)
- [TEST REPORT LINK](https://FlameInBrains.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,27 @@
+body {
+  margin: 0;
+}
+
+.stars__star {
+  background-image: url(images/star.svg);
+  background-repeat: no-repeat;
+  height: 16px;
+  width: 16px;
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,27 @@
...
+
+.stars__star:not(:last-child) {
+  margin-right: 4px;
+}
+
+.stars {
+  display: flex;
+}
+
+.stars--5 .stars__star,
```
anstsot, line 21: add proper `:nth-child` because what if you’ll have more than 5 stars children but the number of active stars will be `5`?

anstsot (approved):
But pay attention to my comment, please

PR #3067, opened 2023-05-30
add task solution

fawnique:
https://fawnique.github.io/layout_stars/
https://fawnique.github.io/layout_stars/report/html_report/


And678 (changes requested):
Check the first point of the [checklist](https://github.com/mate-academy/layout_stars/blob/master/checklist.md) for this task, make sure the indentations are correct, otherwise good work, fix this and this is done :100:

tetiana-ivanchenko (approved):
GJ 👍

PR #3071, opened 2023-05-30
Add task solution

vladmakohon:
- [DEMO LINK](https://vladmakohon.github.io/layout_stars/)
- [TEST REPORT LINK](https://vladmakohon.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,30 @@
+body {
+  margin: 0;
+}
+
+.stars {
+  display: flex;
+}
+
+.stars__star {
+  height: 16px;
+  width: 16px;
+  background-image: url(images/star.svg);
+  background-repeat: no-repeat;
+  background-position: center;
+}
+
+.stars__star:not(:last-child) {
+  margin-right: 4px;
+}
+
+.stars--5 .stars__star,
+.stars--1 .stars__star:nth-child(1),
+.stars--2 .stars__star:nth-child(1),
+.stars--2 .stars__star:nth-child(2),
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,30 @@
...
+.stars--5 .stars__star,
+.stars--1 .stars__star:nth-child(1),
+.stars--2 .stars__star:nth-child(1),
+.stars--2 .stars__star:nth-child(2),
+.stars--3 .stars__star:nth-child(1),
+.stars--3 .stars__star:nth-child(2),
+.stars--3 .stars__star:nth-child(3),
+.stars--4 .stars__star:not(:nth-last-child(1)) {
+  background-image: url(images/star-active.svg);
+}
```
And678, line 30: Good work using nth-child, but it can be improved:
Instead of having:
```css
.stars--3 .stars__star:nth-child(1),
.stars--3 .stars__star:nth-child(2),
.stars--3 .stars__star:nth-child(3),
```
you can have this, using only one selector.
```css
.stars--3 .stars__star:nth-child(-n + 3)
```

Same for other number of stars, to see how it works, check out [the docs](https://developer.mozilla.org/en-US/docs/Web/CSS/:nth-child)

And678 (changes requested):
Good work, only one thing to improve below, and this is done :100:

vladmakohon:
Didn't notice at first that there is no sense in using `:nth-child(-n + 1)`

tetiana-ivanchenko (approved):
GJ 👍

PR #3073, opened 2023-05-30
add task solution

ArtemAndrushchenko:
[DEMO LINK](https://ArtemAndrushchenko.github.io/layout_stars/)
- [TEST REPORT LINK](https://ArtemAndrushchenko.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -8,6 +8,47 @@
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Stars</h1>
+    <div class="stars stars--0">
+      <div class="star"></div>
+      <div class="star"></div>
+      <div class="star"></div>
+      <div class="star"></div>
+      <div class="star"></div>
+    </div>
+    <div class="stars stars--1">
+      <div class="star stars--N"></div>
+      <div class="star"></div>
+      <div class="star"></div>
+      <div class="star"></div>
+      <div class="star"></div>
+    </div>
+    <div class="stars stars--2">
+      <div class="star stars--N"></div>
+      <div class="star stars--N"></div>
+      <div class="star"></div>
+      <div class="star"></div>
+      <div class="star"></div>
+    </div>
+    <div class="stars stars--3">
+      <div class="star stars--N"></div>
+      <div class="star stars--N"></div>
+      <div class="star stars--N"></div>
+      <div class="star"></div>
+      <div class="star"></div>
+    </div>
+    <div class="stars stars--4">
+      <div class="star stars--N"></div>
+      <div class="star stars--N"></div>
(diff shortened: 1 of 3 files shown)


src/index.html
```diff
@@ -8,6 +8,47 @@
...
   <body>
-    <h1>Stars</h1>
+    <div class="stars stars--0">
+      <div class="star"></div>
+      <div class="star"></div>
+      <div class="star"></div>
+      <div class="star"></div>
+      <div class="star"></div>
+    </div>
+    <div class="stars stars--1">
```
anstsot, line 18: do not forget to add empty lines between multiline sibling blocks of HTML (checklist - point 1)

src/index.html
```diff
@@ -8,6 +8,47 @@
...
+      <div class="star stars--N"></div>
```
anstsot, line 19: there should be no additional class for active star. you have to rely only on parent `.stars--N` class

src/style.css
```diff
@@ -0,0 +1,30 @@
...
+  width: 16px;
+  margin-left: 4px;
+  border-radius: 0.5px;
+}
+
+.stars {
+  display: flex;
+}
+
+.stars--5 .star,
```
anstsot, line 24: please add proper `:nth-child` because what if you’ll have more than 5 stars children but the number of active stars will be `5`?

src/style.css
```diff
@@ -0,0 +1,30 @@
...
+
+.star {
+  background: url("images/star.svg");
+  background-repeat: no-repeat;
+  background-size: contain;
+  background-position: center;
+  height: 16px;
+  width: 16px;
+  margin-left: 4px;
+  border-radius: 0.5px;
```
anstsot, line 17: is `border-radius` needed here?

anstsot (changes requested):
Good start, but you have several fixes and `stars--N` class for active star should not be used.
Instead, you should rely only on **parent** `stars--N` class. Try to use `:nth-child` selector - [docs](https://developer.mozilla.org/en-US/docs/Web/CSS/:nth-child). 

Don't forget to fix all your tests

src/style.css
```diff
@@ -0,0 +1,29 @@
...
+}
+
+.star {
+  background: url("images/star.svg");
+  background-repeat: no-repeat;
+  background-size: contain;
+  background-position: center;
+  height: 16px;
+  width: 16px;
+  margin-left: 4px;
```
YegorKorenkov, line 16: Please remove margin-left for each first star
image

YegorKorenkov (changes requested):
You didn't pass any tests, try changing the value of the `background-size` rule to pass the tests, also please pay sttention on comment

PR #3075, opened 2023-05-30
task stars solution

outofs:
https://outofs.github.io/layout_stars/
https://outofs.github.io/layout_stars/report/html_report/

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -10,7 +10,6 @@
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <!-- <h1>Stars</h1> -->
     <div class="stars stars--0">
       <div class="stars__star"></div>
 
@@ -24,7 +23,7 @@
     </div>
 
     <div class="stars stars--1">
-      <div class="stars__star active"></div>
+      <div class="stars__star"></div>
 
       <div class="stars__star"></div>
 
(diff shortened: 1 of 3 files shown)


src/index.html
```diff
@@ -1,13 +1,86 @@
...
+    <meta
+      name="viewport"
+      content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0"
+    >
     <title>Stars</title>
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Stars</h1>
+    <!-- <h1>Stars</h1> -->
```
anstsot, line 13: please don't push comments
```suggestion
```

src/index.html
```diff
@@ -1,13 +1,86 @@
...
+
+      <div class="stars__star"></div>
+
+      <div class="stars__star"></div>
+
+      <div class="stars__star"></div>
+    </div>
+
+    <div class="stars stars--1">
+      <div class="stars__star active"></div>
```
anstsot, line 27: don't add `active` modifier class for star. rely only on parent `stars--N` class.

anstsot (changes requested):
Instead, you should rely only on parent `stars--N` class. You can use `:nth-child` selector - [docs](https://developer.mozilla.org/en-US/docs/Web/CSS/:nth-child).

anstsot (approved):

PR #3076, opened 2023-05-30
add task solution

ivankovbohdan:
- [DEMO LINK](https://ivankovbohdan.github.io/layout_stars/)
- [TEST REPORT LINK](https://ivankovbohdan.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -3,11 +3,58 @@
   <head>
     <meta charset="UTF-8">
     <meta name="viewport"
-          content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0">
+          content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0"
+    >
     <title>Stars</title>
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Stars</h1>
+    <div class="stars stars--0">
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+    </div>
+
+    <div class="stars stars--1">
+      <div class="stars__star stars__star--active"></div>
+      <div class="stars__star"></div>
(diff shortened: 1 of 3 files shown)


src/index.html
```diff
@@ -3,11 +3,58 @@
...
+    <div class="stars stars--0">
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+    </div>
+
+    <div class="stars stars--1">
+      <div class="stars__star stars__star--active"></div>
```
anstsot, line 21: do not use `active` modifier class for star. rely only on parent `stars--N` class

anstsot (changes requested):
Instead, you should rely only on `stars--N` class. You can use `:nth-child` selector - [docs](https://developer.mozilla.org/en-US/docs/Web/CSS/:nth-child).

YegorKorenkov (changes requested):
Hi, you probably didn’t really understand what your mistake is, you don’t need to set any additional classes for stars, except for stars__star. You have to make the stars active depending on the parent container's modifier class stars--1(should be 1 gold star), stars--2(should be 2 active stars)

src/style.css
```diff
@@ -0,0 +1,43 @@
...
+
+.stars--3 .stars__star:nth-child(n+4) {
+  background-image: url("images/star.svg");
+}
+
+.stars--4 .stars__star:nth-child(n+5) {
+  background-image: url("images/star.svg");
+}
+
+.stars--5 .stars__star:nth-child(n+6) {
```
anstsot, line 41: you can merge the styles
```suggestion
.stars--0 .stars__star:nth-child(n+1),
.stars--1 .stars__star:nth-child(n+2),
.stars--2 .stars__star:nth-child(n+3),
.stars--3 .stars__star:nth-child(n+4),
.stars--4 .stars__star:nth-child(n+5),
.stars--5 .stars__star:nth-child(n+6) {
```

anstsot (changes requested):
Almost done, let's add the last fix

anstsot (approved):

PR #3077, opened 2023-05-30
stars

Niko932205:
https://Niko932205.github.io/layout_stars
https://Niko932205.github.io/layout_stars/report/html_report

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -10,13 +10,13 @@ body {
   height: 16px;
   width: 16px;
   background-image: url(images/star.svg);
-  background-size: contain;
   background-repeat: no-repeat;
   background-position: center;
+  margin-left: 4px;
 }
 
-.stars_star:not(:last-child) {
-  margin-left: 4px;
+.stars_star:first-child {
+  margin-left: 0;
 }
 
 .stars--1 > :nth-child(1),


src/style.css
```diff
@@ -0,0 +1,55 @@
...
+}
+
+.stars--1 .stars_star:first-child {
+  background-image: url(images/star-active.svg);
+  background-size: contain;
+  background-repeat: no-repeat;
+  background-position: center;
+  height: 16px;
+  width: 16px;
+  margin-left: 4px;
```
anstsot, line 31: these styles are already set for `.stars_star` class so there is no need to duplicate

src/style.css
```diff
@@ -0,0 +1,55 @@
...
+.stars--3 .stars_star:nth-child(3),
+.stars--4 .stars_star:first-child,
+.stars--4 .stars_star:nth-child(2),
+.stars--4 .stars_star:nth-child(3),
+.stars--4 .stars_star:nth-child(4),
+.stars--5 .stars_star:first-child,
+.stars--5 .stars_star:nth-child(2),
+.stars--5 .stars_star:nth-child(3),
+.stars--5 .stars_star:nth-child(4),
+.stars--5 .stars_star:nth-child(5) {
```
anstsot, line 47: try to check the `:nth-child` pseudo-class to reduce the code. here can be used next construction as `:nth-child(-n + 3)` etc.
here is the [doc](https://developer.mozilla.org/en-US/docs/Web/CSS/:nth-child)

src/style.css
```diff
@@ -0,0 +1,55 @@
+body {
+  margin: 0;
+}
+
+.container {
+  height: 96px;
+  width: 96px;
```
anstsot, line 7: do not hardcode the `weight` and `height`

anstsot (changes requested):
Good start:four_leaf_clover:
Let’s add a few fixes to improve your code. and please fix all tests because they're failed

Niko932205:
> Good work from a code perspective but the tests failed. please try to fix them

check it psl

PR #3078, opened 2023-05-30
add task solution Stars

Max-Komov:
(https://Max-Komov.github.io/layout_stars/)
(https://Max-Komov.github.io/layout_stars/report/html_report/)


slavanosatskyi (approved):
👍 nice work

PR #3093, opened 2023-05-31
develop stars

TarasB01:
[DEMO LINK](https://TarasB01.github.io/layout_stars/)
[TEST REPORT LINK](https://TarasB01.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -8,6 +8,47 @@
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Stars</h1>
+    <div class="stars stars--0">
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+    </div>
+    <div class="stars stars--1">
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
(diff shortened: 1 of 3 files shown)


src/index.html
```diff
@@ -8,6 +8,47 @@
...
   <body>
-    <h1>Stars</h1>
+    <div class="stars stars--0">
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+    </div>
+    <div class="stars stars--1">
```
anstsot, line 18: do not forget to add empty lines between multiline sibling blocks of HTML (checklist - point 1)

src/style.css
```diff
@@ -0,0 +1,27 @@
...
+  height: 16px;
+  width: 16px;
+  background-repeat: no-repeat;
+}
+
+.stars__star:not(:last-child) {
+  margin-right: 4px;
+}
+
+.stars--5 .stars__star,
```
anstsot, line 21: it will be better to add proper `:nth-child` because what if you’ll have more than 5 stars children but the number of active stars will be `5`?

anstsot (changes requested):
Good start:four_leaf_clover:
Let’s fix the code according to the checklist

PR #3123, opened 2023-06-08
add task solution

s-huralskyi:


src/style.css
```diff
@@ -0,0 +1,25 @@
...
+}
+
+.stars {
+  display: flex;
+}
+
+.stars__star {
+  width: 16px;
+  height: 16px;
+  margin-right: 4px;
```
mykhalenych, line 12: the last star shouldn't have margin-right

mykhalenych (changes requested):
don't forget add demo and test links to description

src/style.css
s-huralskyi: done

etojeDenys (approved):
lgtm

PR #3138, opened 2023-06-12
add stars block

OPTIMISTIXX:
https://OPTIMISTIXX.github.io/layout_stars/
https://OPTIMISTIXX.github.io/layout_stars/report/html_report/

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,42 @@
+body {
+  margin: 0;
+  padding: 0;
+}
+
+.stars {
+  display: flex;
+}
+
+.stars__star {
+  background-image: url(images/star.svg);
+  background-position: center;
+  background-repeat: no-repeat;
+  width: 16px;
+  height: 16px;
+  margin: 0 2px;
+  border-radius: 0.5px;
+}
+
+.stars__star:first-child {
+  margin-left: 0;
+}
+
+.stars--5 > .stars__star {
+  background-image: url(images/star-active.svg);
+}
+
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,42 @@
+body {
+  margin: 0;
+  padding: 0;
```
vlad-demchuk, line 3: body doesn't have paddings

src/style.css
```diff
@@ -0,0 +1,42 @@
...
+  background-image: url(images/star.svg);
+  background-position: center;
+  background-repeat: no-repeat;
+  width: 16px;
+  height: 16px;
+  margin: 0 2px;
+  border-radius: 0.5px;
+}
+
+.stars__star:first-child {
```
vlad-demchuk, line 20: you can use :not pseudo-class instead

src/style.css
```diff
@@ -0,0 +1,42 @@
...
+  margin-left: 0;
+}
+
+.stars--5 > .stars__star {
+  background-image: url(images/star-active.svg);
+}
+
+.stars--1 > .stars__star:first-child {
```
vlad-demchuk, line 28: combine selectors with comma
```suggestion
.stars--5 > .stars__star,
.stars--1 > .stars__star:first-child {
  background-image: url(images/star-active.svg);
}
```

vlad-demchuk (changes requested):

volodymyr-soltys97 (approved):
Nice work 👍

PR #3144, opened 2023-06-13
add task solution

mshmigelska:
- [DEMO LINK](https://mshmigelska.github.io/layout_stars/)
- [TEST REPORT LINK](https://mshmigelska.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,44 @@
+body {
+  margin: 0;
+}
+
+.stars {
+  display: flex;
+}
+
+.stars__star {
+  background-image: url(images/star.svg);
+  width: 16px;
+  height: 16px;
+  background-repeat: no-repeat;
+  background-position: center;
+  margin-right: 4px;
+}
+
+.stars :last-child {
+  margin: 0;
+}
+
+.stars--1 :nth-child(-n+1) {
+  background-image: url(images/star-active.svg);
+}
+
+.stars--2 :nth-child(-n+2) {
+  background-image: url(images/star-active.svg);
+}
+
+.stars--3 :nth-child(-n+3) {
+  background-image: url(images/star-active.svg);
+}
+
+.stars--4 :nth-child(-n+4) {
+  background-image: url(images/star-active.svg);
+}
+
+.stars--5 :nth-child(-n+5) {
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,44 @@
...
+  background-image: url(images/star-active.svg);
+}
+
+.stars--5 :nth-child(-n+5) {
+  background-image: url(images/star-active.svg);
+}
+
+.stars--6 :nth-child(-n+6) {
+  background-image: url(images/star-active.svg);
+}
```
OleziO, line 44: Optimize code. Replace with:

```
.stars--1 :nth-child(-n+1),
.stars--2 :nth-child(-n+2),
.stars--3 :nth-child(-n+3),
.stars--4 :nth-child(-n+4),
.stars--5 :nth-child(-n+5) {
}
```

OleziO (changes requested):

src/style.css
```diff
@@ -0,0 +1,29 @@
...
+.stars :last-child {
+  margin: 0;
+}
+
+.stars--1 :nth-child(-n+1),
+.stars--2 :nth-child(-n+2),
+.stars--3 :nth-child(-n+3),
+.stars--4 :nth-child(-n+4),
+.stars--5 :nth-child(-n+5),
+.stars--6 :nth-child(-n+6) {
```
volodymyr-soltys97, line 27: There should be only 5 star elements inside the .stars block, so the .stars--6 modifier class is not needed. Remove the last line in the CSS rules for active stars.

src/style.css
```diff
@@ -0,0 +1,29 @@
...
+.stars__star {
+  background-image: url(images/star.svg);
+  width: 16px;
+  height: 16px;
+  background-repeat: no-repeat;
+  background-position: center;
+  margin-right: 4px;
+}
+
+.stars :last-child {
```
volodymyr-soltys97, line 18: You should use the .stars__star:last-child selector instead of .stars :last-child to specifically target the last star element in the .stars block.

volodymyr-soltys97 (changes requested):
Good job 👍

volodymyr-soltys97 (approved):

PR #3145, opened 2023-06-13
Stars layout

oshapkun:
- [DEMO LINK](https://oshapkun.github.io/layout_stars/)
- [TEST REPORT LINK](https://oshapkun.github.io/layout_stars/report/html_report/)

I have an unexpected error but I don't know what I can do with it for now


vlad-demchuk (changes requested):
In this case, you should ask for help in fe_chat

PR #3153, opened 2023-06-15
Develop

Znamm9:
- [DEMO LINK](https://znamm9.github.io/layout_stars/)
- [TEST REPORT LINK](https://znamm9.github.io/layout_stars/


src/styles.css
```diff
@@ -0,0 +1,27 @@
...
+}
+
+.stars__star.active {
+  background: url("images/star-active.svg");
+  background-repeat: no-repeat;
+  background-size: auto;
+  background-position: center;
+  width: 16px;
+  height: 16px;
+}
```
IvanFesenko, line 27: Remove active class number of active stars should enebled by `stars--n` modifier and `:nth-child`
Do not repeat your self too. 
```suggestion
.stars--n :nth-child {
  background: url("images/star-active.svg");
}
```

IvanFesenko (changes requested):
Your demo link isn't working

src/styles.css
```diff
@@ -0,0 +1,44 @@
...
+  background-image: url(images/star-active.svg);
+}
+
+.stars--4 .stars__star:nth-child(-n+4) {
+  background-image: url(images/star-active.svg);
+}
+
+.stars--5 .stars__star:nth-child(-n+5) {
+  background-image: url(images/star-active.svg);
+}
```
maxim2310, line 44: combine these rules using ","

maxim2310 (changes requested):
Great job overall! Just one tiny thing needs to be fixed before approval)

src/styles.css
Znamm9: @maxim2310 sorry, I didn't get you. Should be something like .stars--1 .stars__star:nth-child(-n+1), .stars--2 .stars__star:nth-child(-n+2), ... etc.?

etojeDenys: yeah, it will be much better :D

etojeDenys (changes requested):

src/styles.css
Znamm9: done

src/styles.css
```diff
@@ -0,0 +1,32 @@
...
+.stars {
+  display: flex;
+}
+
+.stars__star {
+  background-image: url(images/star.svg);
+  height: 16px;
+  width: 16px;
+  background-repeat: no-repeat;
+  background-position: center;
```
sTorba24, line 19: background shorthand property can be used

sTorba24 (approved):

PR #3159, opened 2023-06-18
layout_stars

Nadinnnka:


src/style.css
```diff
@@ -0,0 +1,31 @@
...
+  margin-right: 4px;
+}
+
+.stars__star--active {
+  height: 16px;
+  width: 16px;
+  background-image: url(images/star-active.svg);
+  background-repeat: no-repeat;
+  background-position: 50%;
+}
```
anastasiiavorobiova, line 27: You don't need the active modifier. The block with stars--N modifier should have exactly N first stars active, so use this logic

anastasiiavorobiova (changes requested):
Great start! Added a tiny comment below.

src/style.css
Nadinnnka: corrected

PR #3172, opened 2023-06-21
add task solution

liiasav:


src/style.css
```diff
@@ -0,0 +1,27 @@
...
+.stars {
+  display: flex;
+}
+
+.stars__star {
+  width: 16px;
+  height: 16px;
+  background-image: url(images/star.svg);
+  background-repeat: no-repeat;
+  background-position: 50%;
```
Viktor-Kost, line 14: you can apply background `shorthand`

Viktor-Kost (approved):
GJ!

src/style.css
liiasav: Thanks

PR #3177, opened 2023-06-22
tried to add the task solution

may-xxx:
https://may-xxx.github.io/layout_stars/
https://may-xxx.github.io/layout_stars/report/html_report/

some mistakes appeared duaring the tests but mentor said that it's probably problems with tests :/


src/style.css
```diff
@@ -0,0 +1,38 @@
...
+  background-image: url(images/star-active.png);
+}
+
+.stars--4 .stars__star:nth-child(-n + 4) {
+  background-image: url(images/star-active.png);
+}
+
+.stars--5 .stars__star:nth-child(-n + 5) {
+  background-image: url(images/star-active.png);
+}
```
volodymyr-soltys97, line 38: Try to rewrite this
image

volodymyr-soltys97 (changes requested):
Good job 👍 
But let's make your code more perfect

volodymyr-soltys97 (approved):
Nice work!

PR #3190, opened 2023-06-26
add task solution

Vayts:


roman-mirzoian (approved):
The task is done well, the code looks good, but be careful about the condition of the task, because it says that "Each block should have 5 elements inside with the class `stars__star`", and now you have the class `stars__item` instead.

PR #3192, opened 2023-06-26
add task solution

thaiduk:


Anna-Tkachenko (commented):
help me with tests

src/index.html
```diff
@@ -8,6 +8,49 @@
...
+    <div class="wrapper">
+      <div class="stars stars--0">
+        <div class="stars__star star__noActive star_noMargin"></div>
+        <div class="stars__star star__noActive"></div>
+        <div class="stars__star star__noActive"></div>
+        <div class="stars__star star__noActive"></div>
+        <div class="stars__star star__noActive"></div>
+      </div>
+      <div class="stars stars--1">
+        <div class="stars__star star__active star_noMargin"></div>
```
Anna-Tkachenko, line 20: Use semantic tags instead of `div` ones

src/index.html
```diff
@@ -8,6 +8,49 @@
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Stars</h1>
+    <div class="wrapper">
+      <div class="stars stars--0">
+        <div class="stars__star star__noActive star_noMargin"></div>
```
Anna-Tkachenko, line 13: You should manage active stars with modifier for `stars` BEM block
You should not add any additional classes to the stars element

Anna-Tkachenko (changes requested):

src/style.css
```diff
@@ -0,0 +1,33 @@
...
+  height: 16px;
+  background-position: center;
+  background-repeat: no-repeat;
+}
+
+.star--noMargin {
+  margin-left: 0;
+}
+
+.star--noActive {
```
dffUqp, line 27: better to use nth-of-type / nth-child pseudo-classes instead of class `star--noActive`/`star--active`:smiley:

src/style.css
```diff
@@ -0,0 +1,33 @@
...
+
+.stars__star {
+  margin-left: 4px;
+  width: 16px;
+  height: 16px;
+  background-position: center;
+  background-repeat: no-repeat;
+}
+
+.star--noMargin {
```
dffUqp, line 23: can we replace it with some pseudo-class?🤔

dffUqp (changes requested):

alexander-ignatow (approved):
Looks good 👍 

Just one point - @Anna-Tkachenko comment on semantic tags was probably not on making a stars row `<li>` but rather making each row individual list, and each star `<li>`

But your solution is valid in the context of the task!

PR #3209, opened 2023-06-29
add task solution

1umamaster:
- [DEMO LINK](https://1umamaster.github.io/layout_stars/)
- [TEST REPORT LINK](https://1umamaster.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -8,6 +8,78 @@
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Stars</h1>
+    <main class="main">
+      <div class="stars stars--0">
+        <div class="stars__star"></div>
+
+        <div class="stars__star"></div>
+
+        <div class="stars__star"></div>
+
+        <div class="stars__star"></div>
+
+        <div class="stars__star"></div>
+      </div>
+
+      <div class="stars stars--1">
+        <div class="stars__star"></div>
+
+        <div class="stars__star"></div>
+
+        <div class="stars__star"></div>
+
+        <div class="stars__star"></div>
+
+        <div class="stars__star"></div>
+      </div>
+
+      <div class="stars stars--2">
+        <div class="stars__star"></div>
+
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,27 @@
+body {
+  margin: 0;
+}
+
+.stars {
+  display: flex;
+  width: 96px;
```
Andriy-Shymkiv, line 7: why you need to set `width: 96px` ? you have the size for each star (`16px`) so let them stretch the parent container

src/style.css
```diff
@@ -0,0 +1,27 @@
...
+  justify-content: space-between;
+  padding-right: 4px;
```
Andriy-Shymkiv, line 9: you really don't need this padding, better give `margin-right` for each star except the last one in `stars__star` block. for ex. you can go something like this:
```
  margin-right: 4px;
}
```

Andriy-Shymkiv (changes requested):
good job! left comments

src/style.css
1umamaster: Thx. My main problem was in using 'space-between'. Just updated task solution.

src/index.html
```diff
@@ -8,6 +8,78 @@
...
+      <div class="stars stars--0">
+        <div class="stars__star"></div>
+
+        <div class="stars__star"></div>
+
+        <div class="stars__star"></div>
+
+        <div class="stars__star"></div>
+
+        <div class="stars__star"></div>
```
DarkMistyRoom, line 21:
```suggestion
        <div class="stars__star"></div>
        <div class="stars__star"></div>
        <div class="stars__star"></div>
        <div class="stars__star"></div>
        <div class="stars__star"></div>
```

DarkMistyRoom: no need to add empty lines between one-line blocks

DarkMistyRoom (approved):
Good job!

PR #3221, opened 2023-07-02
add task solution

kitpes:
- [DEMO LINK](https://kitpes.github.io/layout_stars/)
- [TEST REPORT LINK](https://kitpes.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,36 @@
+.page_body {
+  box-sizing: border-box;
+  margin: 0;
+}
+
+.stars-package {
+  display: flex;
+  flex-direction: column;
+  justify-content: center;
+  align-items: center;
+
+  width: fit-content;
+}
+
+.stars {
+  display: flex;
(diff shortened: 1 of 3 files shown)


volodymyr-soltys97 (changes requested):
No need to add margin-right for the last child.
If you remove it, you task will be better
image

mykhalenych (changes requested):
check comments, also your links aren't working

src/style.css
```diff
@@ -0,0 +1,57 @@
...
+  flex-direction: column;
+  justify-content: center;
+  align-items: center;
+
+  width: fit-content;
+}
+
+.stars {
+  display: flex;
+  flex-direction: row;
```
anastasiiavorobiova, line 17: Consider deleting this value, it's redundant and settled by default

anastasiiavorobiova (approved):
Great solution!

PR #3223, opened 2023-07-02
Develop

robsza1122:


.bemlintrc.json
```diff
@@ -0,0 +1,16 @@
+{
```
Radoslaw-Czerniawski, line 1: Don't add irrelevant changes to the pr, this file has nothing to do with the request star implementation.

src/style.css
```diff
@@ -0,0 +1,31 @@
+body {
+  margin: 0;
```
Radoslaw-Czerniawski, line 2: This only resets margin for the body, not for all elements.

src/style.css
```diff
@@ -0,0 +1,31 @@
...
+}
+
+div {
```
Radoslaw-Czerniawski, line 5: - Don't use tags for styling elements
- div are block by default so this rule is redundant

src/index.html
```diff
@@ -9,5 +9,48 @@
...
+      <div class="stars__star"></div>
+    </div>
+    <div class="stars stars--5">
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+    </div>
+
```
Radoslaw-Czerniawski, line 54: Remove this empty line

src/index.html
```diff
@@ -9,5 +9,48 @@
   </head>
   <body>
     <h1>Stars</h1>
+    <div class="stars stars--0">
```
Radoslaw-Czerniawski, line 12: It makes sense for these stars to be an `ul` with `li`'s rather than just divs, this block will be reused in your future prs so it will future-proof it. 
Generally it's a bad practice to use divs for everything, avoid it as much as possible.

Radoslaw-Czerniawski (changes requested):
- Demo is not available, most likely you need to run `npm run deploy`
image
adjust according to the comments and deploy this pr so it may be viewed

robsza1122:
I have some problem with tests. References are this same but it can not pass.

Radoslaw-Czerniawski (changes requested):
Again requesting changes
- correct reset
- run `npm run deploy` locally to enable the demo -> if problem occurs consult during qna / on chat

src/style.css
DorotaLeniecDev: As Radek mention reset margin only for the body and we are good to go

DorotaLeniecDev (approved):
One more change needed, I will approve

PR #3226, opened 2023-07-03
Task solution

domalewski:
https://domalewski.github.io/layout_stars/
https://domalewski.github.io/layout_stars/report/html_report/

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,24 @@
+body {
+  margin: 0;
+}
+
+.stars {
+  display: flex;
+}
+
+.stars__star {
+  background-image: url(./images/star.svg);
+  background-repeat: no-repeat;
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,24 @@
...
+  display: flex;
+}
+
+.stars__star {
+  background-image: url(./images/star.svg);
+  background-repeat: no-repeat;
+  background-position: center;
+  height: 16px;
+  width: 16px;
+  margin-right: 4px;
```
Oskinavara, line 15: It would be good to not add margin-right to the last star

Oskinavara (changes requested):

choeqq (approved):
All good! Test are passing - Approved ✅

PR #3229, opened 2023-07-03
layout_stars

emil-owczarek:
[DEMO LINK](https://emil-owczarek.github.io/layout_stars/)
[TEST REPORT LINK](https://emil-owczarek.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -22,3 +22,7 @@
 .stars__star_active {
   background-image: url("images/star-active.svg");
 }
+
+.stars__star:last-child {
+  margin: 0;
+}
--- a/src/style.css
+++ b/src/style.css
@@ -1,5 +1,9 @@
-body {
+*,
+*::after,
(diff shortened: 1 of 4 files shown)


src/style.css
```diff
@@ -0,0 +1,24 @@
...
+}
+
+.stars {
+  display: flex;
+}
+
+.stars__star {
+  height: 16px;
+  width: 16px;
+  margin-right: 4px;
```
Oskinavara, line 16: It would be good to not have the margin for the last star.

Oskinavara (changes requested):

src/style.css
emil-owczarek: Sure, I have just changed the margin for each "last child" in a row.

asakevych (approved):
Great job! 🚀 👏 👏

PR #3231, opened 2023-07-04
add task solution

ogospadzipalina:
- [DEMO LINK](https://ogospadzipalina.github.io/layout_stars/)
- [TEST REPORT LINK](https://ogospadzipalina.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -19,6 +19,14 @@
   height: 16px;
 }
 
-.stars__star--active {
-  background-image: url("images/star-active.svg");
+.stars__star:last-child {
+  margin-right: 0;
+}
+
+.stars--1 .stars__star:nth-child(-n+1),
+.stars--2 .stars__star:nth-child(-n+2),
+.stars--3 .stars__star:nth-child(-n+3),
+.stars--4 .stars__star:nth-child(-n+4),
+.stars--5 .stars__star:nth-child(-n+5) {
+  background-image: url(./images/star-active.svg);
 }
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,24 @@
+*,
+*::after,
+*::before {
+  box-sizing: border-box;
+  margin: 0;
+  padding: 0;
+}
+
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,24 @@
...
+
+.stars {
+  display: flex;
+}
+
+.stars__star {
+  background-image: url("images/star.svg");
+  background-position: center;
+  background-repeat: no-repeat;
+  margin-right: 4px;
```
asakevych, line 17: last star shouldn't have margin-right

src/style.css
```diff
@@ -0,0 +1,24 @@
...
+  width: 16px;
+  height: 16px;
+}
+
+.stars__star--active {
+  background-image: url("images/star-active.svg");
+}
```
asakevych, line 24: please do not use modificator here
due to checklist you need to use stars--N to dedicate active stars
try to use `:nth-child`
you can red about it here: https://developer.mozilla.org/en-US/docs/Web/CSS/:nth-child

asakevych (changes requested):
Good job! but you can do it even better, try another solution😉

src/style.css
ogospadzipalina: Thank you! All done

mvjl000 (approved):
Great job!

choeqq (changes requested):
Although your solution looks really nice, it would be perfect if you've completed it using SCSS

Sorry for misunderstanding  - we know that in readme there is option to go with both CSS and SCSS, but we think that SCSS is way more common in larger projects and we want you to know it better!

Please provide solution with SCSS, following [checklist](https://github.com/mate-academy/layout_product-cards/blob/master/checklist.md)

ogospadzipalina:
@choeqq but this task was even before theme scss and checklist you sent us for product card and not for stars 🤔

choeqq (approved):

PR #3233, opened 2023-07-04
add task solution

szymonpachucki:
https://szymonpachucki.github.io/layout_stars/
https://szymonpachucki.github.io/layout_stars/report/html_report/

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -7,7 +7,38 @@
     <title>Stars</title>
     <link rel="stylesheet" href="./style.css">
   </head>
-  <body>
-    <h1>Stars</h1>
+  <body class="body">
+    <div class="Stars"> <img src="images/star.svg" alt="activ">
+      <img src="images/star.svg" alt="activ">
+      <img src="images/star.svg" alt="activ">
+      <img src="images/star.svg" alt="activ">
+      <img src="images/star.svg" alt="activ">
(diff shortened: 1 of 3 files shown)


src/index.html
```diff
@@ -7,7 +7,38 @@
...
+      <img src="images/star-active.svg" alt="not_activ">
+      <img src="images/star-active.svg" alt="not_activ">
+      <img src="images/star-active.svg" alt="not_activ">
+      <img src="images/star.svg" alt="activ">
+      <img src="images/star-active.svg" alt="not_activ">
+      <img src="images/star-active.svg" alt="not_activ">
+      <img src="images/star-active.svg" alt="not_activ">
+      <img src="images/star-active.svg" alt="not_activ">
+      <img src="images/star-active.svg" alt="not_activ">
+    </div>
```
asakevych, line 41: that's not the best solution - please read checklist once more
1. try to use `<div>` elements for each row with class star
2. `<div class="stars stars--1">` can have children divs - example:  `<div class="stars__star">` 
3. try to make those divs appropriate size and give them background image
4. to make some stars yellow you should use `:nth-child` - read here: https://developer.mozilla.org/en-US/docs/Web/CSS/:nth-child

choeqq (approved):
All good, nice job 🚀

PR #3234, opened 2023-07-04
add task solution

HaniaNassalska:
https://hanianassalska.github.io/layout_stars/
https://hanianassalska.github.io/layout_stars/report/html_report/

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -8,49 +8,53 @@
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <div class="stars">
-      <div class="stars--0">
-        <div class="stars__star"></div>
-        <div class="stars__star"></div>
-        <div class="stars__star"></div>
-        <div class="stars__star"></div>
-        <div class="stars__star"></div>
-      </div>
-      <div class="stars--1">
-        <div class="stars__star stars__star--active"></div>
-        <div class="stars__star"></div>
-        <div class="stars__star"></div>
-        <div class="stars__star"></div>
-        <div class="stars__star"></div>
-      </div>
-      <div class="stars--2">
-        <div class="stars__star stars__star--active"></div>
-        <div class="stars__star stars__star--active"></div>
-        <div class="stars__star"></div>
-        <div class="stars__star"></div>
-        <div class="stars__star"></div>
-      </div>
-      <div class="stars--3">
-        <div class="stars__star stars__star--active"></div>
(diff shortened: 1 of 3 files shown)


src/index.html
```diff
@@ -3,11 +3,54 @@
...
     <meta name="viewport"
-          content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0">
+      content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0">
     <title>Stars</title>
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Stars</h1>
+    <div class="stars">
+      <div class="stars--0">
```
asakevych, line 12: 1. stars--0 is modificator, so in one element you should have `class=stars stars--0` 
2. 1 div, which now has class 'stars' is not necessary, but if you want to have some container it can be just div without any class (or with class 'wrapper'  for example)
3. make empty lines beetween blocks

src/style.css
```diff
@@ -0,0 +1,25 @@
...
+  display: flex;
+}
+
+.stars__star {
+  background-image: url("./images/star.svg");
+  background-repeat: no-repeat;
+  background-position: center;
+  height: 16px;
+  width: 16px;
+  margin-right: 4px;
```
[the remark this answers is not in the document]
HaniaNassalska: Thanks for advice! :)

HaniaNassalska:
1. div with 'stars' class removed;
2. 'stars' class moved to lower blocks; 
3. blank lines between blocks

src/style.css
```diff
@@ -0,0 +1,24 @@
...
+}
+
+.stars__star:last-child {
+  margin-right: 0;
+}
+
+.stars__star--active {
+  background-image: url("./images/star-active.svg");
+}
```
choeqq, line 24: The block with stars--N modifier should have exactly N first stars active, according to [requirements](https://github.com/mate-academy/layout_stars#the-task)

```suggestion
.stars--1 .stars__star:nth-child(-n+1), {
  ... more selectors to target every line of stars ... 
  background-image: url("./images/star-active.svg");
}
```

choeqq (changes requested):
You solution is good, but we want to depend on N, not concrete numbers 🔢

asakevych (approved):
Great job!! 🚀 👏 👏

PR #3235, opened 2023-07-04
add task solution

marekder:
- [DEMO LINK](https://marekder.github.io/layout_stars/)
- [TEST REPORT LINK](https://marekder.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,20 @@
+body {
+  margin: 0;
+}
+
+.stars {
+  display: flex;
+}
+
+.stars__star {
+  height: 16px;
+  width: 16px;
+  margin-right: 4px;
+  background-image: url("images/star.svg");
+  background-position: center;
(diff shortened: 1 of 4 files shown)


src/style.css
```diff
@@ -0,0 +1,20 @@
...
+}
+
+.stars {
+  display: flex;
+}
+
+.stars__star {
+  height: 16px;
+  width: 16px;
+  margin-right: 4px;
```
Oskinavara, line 12: It would be good to add :not(:last-of-type) for this margin so the last star doesn't have the margin

Oskinavara (changes requested):

src/style.css
marekder: Done.

Oskinavara (approved):
Good job

PR #3236, opened 2023-07-04
task solution

Sancz0pansa:
[DEMO LINK](https://Sancz0pansa.github.io/layout_stars/)
[TEST REPORT LINK](https://Sancz0pansa.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -8,6 +8,76 @@
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Stars</h1>
+    <div class="stars stars--0">
+      <div class="stars__star"></div>
+
+      <div class="stars__star"></div>
+
+      <div class="stars__star"></div>
+
+      <div class="stars__star"></div>
+
+      <div class="stars__star"></div>
+    </div>
+
+    <div class="stars stars--1">
+      <div class="stars__star--active"></div>
+
+      <div class="stars__star"></div>
+
+      <div class="stars__star"></div>
+
+      <div class="stars__star"></div>
+
+      <div class="stars__star"></div>
+    </div>
+
+    <div class="stars stars--2">
+      <div class="stars__star--active"></div>
+
+      <div class="stars__star--active"></div>
+
+      <div class="stars__star"></div>
+
+      <div class="stars__star"></div>
+
+      <div class="stars__star"></div>
+    </div>
+
+    <div class="stars stars--3">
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,29 @@
...
+}
+
+.stars__star--active {
+  background-image: url("/images/star-active.svg");
+  background-position: center;
+  background-repeat: no-repeat;
+  width: 16px;
+  height: 16px;
+  margin-left: 4px;
+}
```
mvjl000, line 29: Two things about this piece of code:

1. You are setting all of these styles in a selector above. Get rid of `.stars_star--active`(all stars should have one class - "stars__star")  class and create a different selector which changes only `background-image: url("/images/star-active.svg");`

2. Instead of using class with modificator, try to determine which star needs to have different background image with help of :nth-child and stars--N
You can check the [MDN docs](https://developer.mozilla.org/en-US/docs/Web/CSS/:nth-child)

src/index.html
```diff
@@ -8,6 +8,76 @@
...
+
+      <div class="stars__star"></div>
+
+      <div class="stars__star"></div>
+
+      <div class="stars__star"></div>
+    </div>
+
+    <div class="stars stars--1">
+      <div class="stars__star--active"></div>
```
mvjl000, line 24: Let's keep just one class for start - "stars__star". I have explained it more deeply in your css section ;)

mvjl000 (changes requested):

src/style.css
```diff
@@ -11,19 +11,40 @@ body {
...
+
+.stars--4 > .stars__star:nth-child(n+5) {
+  background-image: url("/images/star.svg");
   background-position: center;
   background-repeat: no-repeat;
-  width: 16px;
-  height: 16px;
-  margin-left: 4px;
 }
```
choeqq: that's good, but you can make it even better 😄 
it seems like you're doing a lot of very repetetive work with those styles, try to stack them inside one block with `,` operator

and you don't really need to change anything about background - it's already set in `.stars__star` class

choeqq (changes requested):
Tests are passing - just small improvements needed

src/style.css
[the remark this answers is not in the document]
Sancz0pansa: I'm using margin-left.

choeqq (approved):
Nicely done, approved! 🚀

PR #3237, opened 2023-07-04
add task solution

cardinal1991:
- [DEMO LINK](https://cardinal1991.github.io/layout_stars/)
- [TEST REPORT LINK](https://cardinal1991.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -8,6 +8,48 @@
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Stars</h1>
+    <div class="stars stars--0">
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+    </div>
+    <div class="stars stars--1">
+      <div class="stars__star modifier"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+    </div>
+    <div class="stars stars--2">
+      <div class="stars__star modifier"></div>
+      <div class="stars__star modifier"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+    </div>
+    <div class="stars stars--3">
+      <div class="stars__star modifier"></div>
+      <div class="stars__star modifier"></div>
+      <div class="stars__star modifier"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+    </div>
+    <div class="stars stars--4">
+      <div class="stars__star modifier"></div>
+      <div class="stars__star modifier"></div>
+      <div class="stars__star modifier"></div>
+      <div class="stars__star modifier"></div>
+      <div class="stars__star"></div>
+    </div>
+    <div class="stars stars--5">
+      <div class="stars__star modifier"></div>
+      <div class="stars__star modifier"></div>
+      <div class="stars__star modifier"></div>
+      <div class="stars__star modifier"></div>
+      <div class="stars__star modifier"></div>
+    </div>
+
   </body>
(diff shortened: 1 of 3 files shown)


src/index.html
```diff
@@ -8,6 +8,48 @@
...
-    <h1>Stars</h1>
+    <div class="stars stars--0">
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+    </div>
+    <div class="stars stars--1">
+      <div class="stars__star modifier"></div>
```
asakevych, line 19: 1. add empty lines between stars blocks
2. if you want to implement it with modifier, you should use for it class with block__element--modifier, so for example it could be `stars__star--active`

src/style.css
```diff
@@ -0,0 +1,24 @@
...
+
+.stars {
+  display: flex;
+}
+
+.stars__star {
+  background-image: url(/images/star.svg);
+  height: 16px;
+  width: 16px;
+  margin-right: 4px;
```
asakevych, line 16: last element shouldn't have margin-right

src/style.css
```diff
@@ -0,0 +1,24 @@
...
+  background-repeat: no-repeat;
+
```
asakevych, line 18: unnecessary empty line

asakevych (changes requested):
Good job! just some small changes needed 😉

src/index.html
```diff
@@ -8,6 +8,53 @@
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Stars</h1>
+    <div class="stars stars--0">
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
```

src/style.css
```diff
@@ -0,0 +1,27 @@
...
+  background-position: center;
+}
+
+.stars:nth-child(5) {
+  margin-right: 0;
+}
+
+.stars__star.stars__star--active {
+  background-image: url(/images/star-active.svg);
+}
```
choeqq, line 27: The block with stars--N modifier should have exactly N first stars active, according to [requirements](https://github.com/mate-academy/layout_stars#the-task)

```suggestion
.stars--1 .stars__star:nth-child(-n+1), {
  ... more selectors to target every line of stars ... 
  background-image: url("./images/star-active.svg");
}
```

choeqq (changes requested):

src/index.html
cardinal1991: like that its an error

src/index.html
```diff
@@ -8,6 +8,53 @@
...
+      <div class="stars__star stars__star--active"></div>
+      <div class="stars__star"></div>
+    </div>
+
+    <div class="stars stars--5">
+      <div class="stars__star stars__star--active"></div>
+      <div class="stars__star stars__star--active"></div>
+      <div class="stars__star stars__star--active"></div>
+      <div class="stars__star stars__star--active"></div>
+      <div class="stars__star stars__star--active"></div>
```
choeqq, line 56: please remove all of those classes - they aren't used right now

src/style.css
```diff
@@ -0,0 +1,43 @@
...
+  background-image: url("./images/star-active.svg");
+}
+
+.stars--4 .stars__star:nth-child(-n+4) {
+  background-image: url("./images/star-active.svg");
+}
+
+.stars--5 .stars__star:nth-child(-n+5) {
+  background-image: url("./images/star-active.svg");
+}
```
choeqq, line 43: looks good, but it's seems like you're doing a lot of repetetive work, let's try to stack every selector inside one block with `,` operator 😄

choeqq (changes requested):
Small adjustments needed

choeqq (approved):
Looks good, approved 🚀

PR #3239, opened 2023-07-04
add task solution

klaudia-klis:
- [DEMO LINK](https://klaudia-klis.github.io/layout_stars/)
- [TEST REPORT LINK](https://klaudia-klis.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -16,6 +16,10 @@ body {
   margin-right: 4px;
 }
 
+.stars__star:last-child {
+  margin-right: 0;
+}
+
 .stars--1 .stars__star:first-child,
 .stars--2 .stars__star:nth-child(-n + 2),
 .stars--3 .stars__star:nth-child(-n + 3),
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,26 @@
...
+  margin: 0;
+  box-sizing: border-box;
+}
+
+.stars {
+  display: flex;
+}
+
+.stars__star {
+  display: block;
```
mvjl000, line 11: There is no need to set `display: block` in here. Divs are block by default ;)

mvjl000 (changes requested):
Just one small adjustment but besides of that, it is great!
Nicely done!

src/style.css
choeqq, line 1: please remove `margin-right` from last star ⭐

choeqq (changes requested):
Only one small change needed and you're good to go! All tests are passing ✅

choeqq (approved):
Well done! ⭐

PR #3244, opened 2023-07-04
add task solution

Vira-V:
[DEMO LINK](https://Vira-V.github.io/layout_stars/)
[TEST REPORT LINK](https://Vira-V.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -8,15 +8,18 @@
 }
 
 .stars__star {
-  background-image: url(./images/star.svg);
-  background-repeat: no-repeat;
-  background-position: 50%;
   width: 16px;
   height: 16px;
+  background-image: url(./images/star.svg);
+  background-repeat: no-repeat;
+  background-position: center;
   margin-right: 4px;
 }
 
-.stars__star--active {
+.stars--1 .stars__star:nth-child(-n+1),
+.stars--2 .stars__star:nth-child(-n+2),
+.stars--3 .stars__star:nth-child(-n+3),
+.stars--4 .stars__star:nth-child(-n+4),
+.stars--5 .stars__star:nth-child(-n+5) {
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,22 @@
...
+  justify-content: center;
+}
+
+.stars__star {
+  background-image: url(./images/star.svg);
+  background-repeat: no-repeat;
+  background-position: 50%;
+  width: 16px;
+  height: 16px;
+  margin-right: 4px;
```
asakevych, line 16: last star shouldn't have margin-right

src/style.css
```diff
@@ -0,0 +1,22 @@
...
+}
+
+.stars__star--active {
+  background-image: url(./images/star-active.svg);
+  background-repeat: no-repeat;
+}
```
asakevych, line 22: please do not use modificator here
due to checklist you need to use stars--N to dedicate active stars
try to use `:nth-child` 
you can red about it here: https://developer.mozilla.org/en-US/docs/Web/CSS/:nth-child

asakevych (changes requested):
Good job! but you can do it even better, try another solution😉

Vira-V:
Changes are uploaded

src/style.css
choeqq, line 1: please remove margin-right from last star ⭐

choeqq (changes requested):
One little change and you're good to go! 🚀

Vira-V:
margin updated 

- [DEMO LINK](https://Vira-V.github.io/layout_stars/)
- [TEST REPORT LINK](https://Vira-V.github.io/layout_stars/report/html_report/)

mvjl000 (approved):
Correct, nicely done!

PR #3247, opened 2023-07-04
add task solution

Capoverde:
[DEMO LINK](https://Capoverde.github.io/layout_stars/)
 [TEST REPORT LINK](https://Capoverde.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -9,6 +9,8 @@
   </head>
   <body>
     <!-- <h1>Stars</h1> -->
+
+    <!-- do nth-childow skasoac klase --active -->
     <div class="stars stars--0">
       <div class="stars__star"></div>
       <div class="stars__star"></div>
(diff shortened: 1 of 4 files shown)


src/style.css
```diff
@@ -0,0 +1,36 @@
...
+  margin-left: 0.25rem;
+}
+
+.stars__star {
+  background-image: url("/images/star.svg");
+}
+
+.stars__star--active {
+  background-image: url("/images/star-active.svg");
+}
```
mvjl000, line 36: You need to make some changes in here.

1.  lets set `background-image: url("/images/star.svg");` in `.stars__star` selector.

2. Get rid of class with modificator both from html and css, use only `stars__star` class.

3.  Determine which star is supposed to have `background-image: url("/images/star-active.svg");` with help of :nth-child and stars--N - here are [MDN docs](https://developer.mozilla.org/en-US/docs/Web/CSS/:nth-child)

src/style.css
```diff
@@ -0,0 +1,47 @@
...
+*::before,
+*::after {
+  margin: 0;
+  padding: 0;
+  box-sizing: border-box;
+}
+
+:root {
+  --star-size: 1rem;
+}
```
choeqq, line 11: please remove unecessary varibales

src/style.css
```diff
@@ -0,0 +1,47 @@
...
+  background-image: url(images/star-active.svg);
+}
+
+.stars--4 :nth-child(-n + 4) {
+  background-image: url(images/star-active.svg);
+}
+
+.stars--5 :nth-child(-n + 5) {
+  background-image: url(images/star-active.svg);
+}
```
choeqq, line 47: that's good, but you can make it even better - it seems like you're doing a lot of repetetive work, let's try to use `,` operator to stack those selectors in one block

src/index.html
```diff
@@ -8,6 +8,50 @@
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Stars</h1>
+    <!-- <h1>Stars</h1> -->
+
+    <!-- do nth-childow skasoac klase --active -->
```
choeqq, line 13: please remove every comment before pushing commit - unless it's really necessary 😉

choeqq (changes requested):
Only small improvements needed

Capoverde:
Corrections are made, pls rview the code. :)

choeqq (approved):
Nice, approved and I would probably use CSS variables only when we're using it more than once 🚀

PR #3251, opened 2023-07-04
add task solution

oliaalia:
- [DEMO LINK](https://oliaalia.github.io/layout_stars/)
- [TEST REPORT LINK](https://oliaalia.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -8,6 +8,53 @@
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Stars</h1>
+    <div class="stars  stars--0">
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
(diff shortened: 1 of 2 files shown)


volodymyr-soltys97 (changes requested):
Demo link. You need to fix it
image

src/index.html
```diff
@@ -8,6 +8,53 @@
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Stars</h1>
+    <div class="stars  stars--0">
```
loralevitska, line 11: you don't need to add more than one space between words, fix all cases
```suggestion
    <div class="stars stars--0">
```

loralevitska (approved):
Well done)

PR #3256, opened 2023-07-04
Develop

DanielToporek:
https://DanielToporek.github.io/layout_stars/
https://DanielToporek.github.io/layout_stars/report/html_report/


choeqq (changes requested):
You solution is good, but we want to depend on N, not concrete numbers 🔢

mvjl000 (approved):
All correct!

PR #3258, opened 2023-07-05
add task solution

SviatoslavVasyletskyi:


src/style.css
choeqq, line 1: please remove `margin-right` from last star ⭐

choeqq (changes requested):
One little change needed and you're good to go! 🚀

src/style.css
SviatoslavVasyletskyi: done!

choeqq: I can't see any updates 🤔

mvjl000 (approved):
Good job!

PR #3260, opened 2023-07-05
Develop

Noini111:
- [DEMO LINK](https://Noini111.github.io/layout_stars/)
- [TEST REPORT LINK](https://Noini111.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -12,80 +12,50 @@
     <div class="block_stars">
       <div class="stars stars--0">
         <div class="stars__star"></div>
-
         <div class="stars__star"></div>
-
         <div class="stars__star"></div>
-
         <div class="stars__star"></div>
-
         <div class="stars__star"></div>
-
       </div>
 
       <div class="stars stars--1">
-        <div class="stars__star stars__star--active"></div>
-
         <div class="stars__star"></div>
-
         <div class="stars__star"></div>
-
         <div class="stars__star"></div>
-
         <div class="stars__star"></div>
-
+        <div class="stars__star"></div>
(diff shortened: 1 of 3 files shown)


src/index.html
```diff
@@ -8,6 +8,86 @@
...
+
+        <div class="stars__star stars__star--active"></div>
+
+        <div class="stars__star stars__star--active"></div>
+
+        <div class="stars__star stars__star--active"></div>
+
+      </div>
+
+    </div>
```
asakevych, line 91: 1. you can have empty lines only between stars blocks, it semms to be easier to read 
2. you can change `:nth-child` for stars to implement different stars color, you do not need to give every stars_star a modifier 😉 
read here:
 
 example:
 `.stars--1 > .stars__star:nth-child(-n + 1)`

asakevych (changes requested):
Great job! but you can make it even better 😉

src/index.html
Noini111: Ok, I will change that. :)

choeqq (approved):
To address your previous comment: 

I think both solutions are good, but at the beginning of your journey you should learn new things - you won't be as shocked when you see it in the future 😄 

All test are passing ✅

src/style.css
```diff
@@ -0,0 +1,49 @@
+/* #region Reset */
+
+:root {
+  margin: 0;
+  padding: 0;
+}
```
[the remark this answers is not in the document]
Noini111: ok!

PR #3264, opened 2023-07-05
add task solution

arturwisnia:


src/style.css
```diff
@@ -0,0 +1,20 @@
...
+.stars__star {
+  height: 16px;
+  width: 16px;
+  background-image: url(images/star.svg);
+  background-position: center;
+  background-repeat: no-repeat;
+  margin-right: 4px;
+}
+
+.stars__star--isactive {
```
mvjl000, line 18: Instead of using class with modificator, try to determine which star needs to have different background image with help of :nth-child and stars--N
You can check the [MDN docs](https://developer.mozilla.org/en-US/docs/Web/CSS/:nth-child)

So, get rid of `.stars__star--isactive` class, same in HTML, leave just "stars__star" class.

mvjl000 (changes requested):
One change to code and the task is done!

choeqq (commented):
Waiting for any updates @arturwisnia 👀

PR #3266, opened 2023-07-05
Develop

vensecue:
- [DEMO LINK](https://vensecue.github.io/layout_stars/)
- [TEST REPORT LINK](https://vensecue.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -9,7 +9,7 @@
   </head>
   <body>
     <div class="stars--0 starline">
-      <div class="stars__sta`r-inactive"></div>
+      <div class="stars__star-inactive"></div>
       <div class="stars__star-inactive"></div>
       <div class="stars__star-inactive"></div>
       <div class="stars__star-inactive"></div>
--- a/src/index.html
+++ b/src/index.html
@@ -8,6 +8,47 @@
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Stars</h1>
+    <div class="stars--0 starline">
+      <div class="stars__sta`r-inactive"></div>
+      <div class="stars__star-inactive"></div>
+      <div class="stars__star-inactive"></div>
+      <div class="stars__star-inactive"></div>
+      <div class="stars__star-inactive"></div>
+    </div>
+    <div class="stars--1 starline">
+      <div class="stars__star-active"></div>
+      <div class="stars__star-inactive"></div>
+      <div class="stars__star-inactive"></div>
+      <div class="stars__star-inactive"></div>
+      <div class="stars__star-inactive"></div>
+    </div>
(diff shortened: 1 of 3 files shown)


src/index.html
```diff
@@ -8,6 +8,47 @@
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Stars</h1>
+    <div class="stars--0 starline">
+      <div class="stars__star-inactive"></div>
+      <div class="stars__star-inactive"></div>
+      <div class="stars__star-inactive"></div>
+      <div class="stars__star-inactive"></div>
+      <div class="stars__star-inactive"></div>
```

src/index.html
```diff
@@ -8,6 +8,47 @@
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Stars</h1>
+    <div class="stars--0 starline">
+      <div class="stars__star-inactive"></div>
```
choeqq, line 12: by default every start is inactive, so we don't need another class to explicitly set that

src/style.css
```diff
@@ -0,0 +1,29 @@
...
+.stars__star-active {
+  background-image: url(/images/star-active.svg);
+  height: 16px;
+  width: 16px;
+  margin-right: 4px;
+  background-position: 50%;
+  background-repeat: no-repeat;
+}
+
+.stars__star-inactive {
```
choeqq, line 22: please provide default start class to style every star initially, and then modify active stars

src/style.css
```diff
@@ -0,0 +1,29 @@
...
+}
+
+.stars__star-active {
+  background-image: url(/images/star-active.svg);
+  height: 16px;
+  width: 16px;
+  margin-right: 4px;
+  background-position: 50%;
+  background-repeat: no-repeat;
+}
```
choeqq, line 20: The block with stars--N modifier should have exactly N first stars active, according to [requirements](https://github.com/mate-academy/layout_stars#the-task)

```suggestion
.stars--1 .stars__star:nth-child(-n+1), {
  ... more selectors to target every line of stars ... 
  background-image: url("./images/star-active.svg");
}
```

src/style.css
```diff
@@ -0,0 +1,29 @@
+*,
+:after,
+:before {
```
choeqq, line 3: we use `::` to access pseudoelements - `:` is for pseudoclasses

choeqq (changes requested):

Try not to repeat code as much as you can 😉

choeqq (commented):
Waiting for any updates @vensecue 👀

vensecue:
Updated now, should be fine!

choeqq (approved):
Nice! 🚀

PR #3269, opened 2023-07-05
task done

patryk177m:
[DEMO LINK](https://patryk177m.github.io/layout_stars/)
- [TEST REPORT LINK](https://patryk177m.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -9,39 +9,39 @@
   </head>
   <body>
     <div class="stars stars--0">
-      <div class="stars__star no_active"></div>
-      <div class="stars__star no_active"></div>
-      <div class="stars__star no_active"></div>
-      <div class="stars__star no_active"></div>
-      <div class="stars__star no_active"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
     </div>
     <div class="stars stars--1">
       <div class="stars__star active"></div>
-      <div class="stars__star no_active"></div>
-      <div class="stars__star no_active"></div>
-      <div class="stars__star no_active"></div>
-      <div class="stars__star no_active"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
     </div>
     <div class="stars stars--2">
       <div class="stars__star active"></div>
       <div class="stars__star active"></div>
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,30 @@
...
+
+.stars {
+  display: flex;
+  /* justify-content: center;
+  align-items: center; */
+}
+
+.no_active {
+  background-image: url("images/star.svg");
+}
```
choeqq, line 14: if star isn't active it's just a star, so you can add those styles to single star, you don't need to add another class

src/index.html
```diff
@@ -8,6 +8,47 @@
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Stars</h1>
+    <div class="stars stars--0">
+      <div class="stars__star no_active"></div>
```
choeqq, line 12: please remove no_active class

src/index.html
```diff
@@ -8,6 +8,47 @@
...
+      <div class="stars__star no_active"></div>
+      <div class="stars__star no_active"></div>
+      <div class="stars__star no_active"></div>
+      <div class="stars__star no_active"></div>
```

src/style.css
```diff
@@ -0,0 +1,30 @@
...
+
+.active {
+  background-image: url("images/star-active.svg");
+}
```
choeqq, line 18: The block with stars--N modifier should have exactly N first stars active, according to [requirements](https://github.com/mate-academy/layout_stars#the-task)

```suggestion
.stars--1 .stars__star:nth-child(-n+1), {
  ... more selectors to target every line of stars ... 
  background-image: url("./images/star-active.svg");
}
```

src/style.css
```diff
@@ -0,0 +1,30 @@
+body {
+  box-sizing: border-box;
+  margin: 0;
+}
+
+.stars {
+  display: flex;
+  /* justify-content: center;
+  align-items: center; */
+}
```
choeqq, line 10: please remove those comments if they're not used

src/style.css
```diff
@@ -0,0 +1,30 @@
...
+
+.no_active {
+  background-image: url("images/star.svg");
+}
+
+.active {
+  background-image: url("images/star-active.svg");
+}
+
+.stars__star {
```
choeqq, line 20: add default background to this class

choeqq (changes requested):

mvjl000 (changes requested):

mvjl000 (approved):
Looks good 👌

PR #3275, opened 2023-07-07
add task solution

ihor-pavlov:
[DEMO LINK](https://ihor-pavlov.github.io/layout_stars/)
[TEST REPORT LINK](https://ihor-pavlov.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,43 @@
+body {
+  margin: 0;
+  padding: 0;
+}
+
+.stars {
+  display: flex;
+  /* width: fit-content; */
+  align-items: center;
+}
+
+.stars__star {
+  height: 16px;
+  width: 16px;
+  background-image: url(images/star.svg);
+  background-repeat: no-repeat;
+  background-position: center;
+  margin-right: 4px;
+}
+
+.stars--0 :nth-child(-n+0) {
+  background-image: url(images/star-active.svg);
+}
+
+.stars--1 :nth-child(-n+1) {
+  background-image: url(images/star-active.svg);
+}
+
+.stars--2 :nth-child(-n+2) {
+  background-image: url(images/star-active.svg);
+}
+
+.stars--3 :nth-child(-n+3) {
+  background-image: url(images/star-active.svg);
+}
+
+.stars--4 :nth-child(-n+4) {
+  background-image: url(images/star-active.svg);
+}
+
+.stars--5 :nth-child(-n+5) {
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,43 @@
...
+  background-image: url(images/star-active.svg);
+}
+
+.stars--4 :nth-child(-n+4) {
+  background-image: url(images/star-active.svg);
+}
+
+.stars--5 :nth-child(-n+5) {
+  background-image: url(images/star-active.svg);
+}
```
Andriy-Shymkiv, line 43: use nested selectors. here you can go something like this:
```css
.stars--1 :nth-child(-n + 1),
.stars--2 :nth-child(-n + 2),
.stars--3 :nth-child(-n + 3),
.stars--4 :nth-child(-n + 4),
}
```
I didn't test it localy but hope it will work

src/style.css
```diff
@@ -0,0 +1,43 @@
+body {
+  margin: 0;
+  padding: 0;
+}
+
+.stars {
+  display: flex;
+  /* width: fit-content; */
```
Andriy-Shymkiv, line 8: remove comments

src/style.css
```diff
@@ -0,0 +1,43 @@
...
+  align-items: center;
```
Andriy-Shymkiv, line 9: you don't need to center the flex items if they are the same size

src/style.css
```diff
@@ -0,0 +1,43 @@
...
+}
+
+.stars__star {
+  height: 16px;
+  width: 16px;
+  background-image: url(images/star.svg);
+  background-repeat: no-repeat;
+  background-position: center;
+  margin-right: 4px;
```
Andriy-Shymkiv, line 18: please do not set the `margin-right`  for last child. here you can go like this:
```css
  margin-right: 4px;
}
```

src/style.css
```diff
@@ -0,0 +1,43 @@
+body {
+  margin: 0;
+  padding: 0;
```
Andriy-Shymkiv, line 3: no need to reset padding here, as it zero by default

Andriy-Shymkiv (changes requested):
good job! left comments

src/style.css
ihor-pavlov: If I remove the margin-right for last child, six autometed tests are failed. I cann't do it.

Andriy-Shymkiv (approved):
good job!

PR #3277, opened 2023-07-07
created star blocks

maksym-mishchanchuk:


src/style.css
```diff
@@ -0,0 +1,24 @@
...
+  background-position: center;
+}
+
+.stars--1 > :first-child,
+.stars--2 > :nth-child(-n + 2),
+.stars--3 > :nth-child(-n + 3),
+.stars--4 > :nth-child(-n + 4),
+.stars--5 > :nth-child(-n + 5) {
+  background-image: url(./images/star-active.svg);
+}
```
alexander-ignatow, line 24: This is correct nd valid code, however you can make it better and cleaner in two steps:

1. replace :first-child with :nth-child(...guess what you should put her...)
2. utilize https://sass-lang.com/documentation/at-rules/control/for/
3.

PR #3278, opened 2023-07-07
add task solution

cty2802:


src/style.css
```diff
@@ -0,0 +1,30 @@
...
+  width: fit-content;
+}
+
+.stars {
+  display: flex;
+  justify-content: space-between;
+}
+
+.stars__star {
+  padding-right: 4px;
```
alexander-ignatow, line 15: Don't use paddings to align items inside flex-box.

In your particular case `padding-right: 4px` has no effect as you set the `width` explicitly.
So as a result you get non-optimal item sizing (notice - items in figma are 16x16, not 16x20)

The solution here is:
1. set item witdth to 16px
2. use margins to align items as you need inside flex container

alexander-ignatow (changes requested):
Good - just one fila touch to make it perfect!

src/style.css
cty2802: Fixed

src/index.html
```diff
@@ -8,6 +8,52 @@
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Stars</h1>
+    <!-- <h1>Stars</h1> -->
```
BudnikOleksii, line 11: Remove comments.

src/index.html
```diff
@@ -8,6 +8,52 @@
...
+        <div class="stars__star"></div>
+      </div>
+      <div class="stars stars--5">
+        <div class="stars__star"></div>
+        <div class="stars__star"></div>
+        <div class="stars__star"></div>
+        <div class="stars__star"></div>
+        <div class="stars__star"></div>
+      </div>
+
```
BudnikOleksii, line 56: Remove unnecessary empty line.

BudnikOleksii (approved):
Nice job👍🏻
Don't forget to clean comments and unnecessary empty lines.

PR #3279, opened 2023-07-07
add task solution

GavriliukArtem:
done

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,29 @@
+body {
+  margin: 0;
+}
+
+.stars {
+  display: flex;
+}
+
+.stars__star {
+  width: 16px;
+  height: 16px;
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,29 @@
...
+  margin-right: 4px;
+
+  background-image: url(./images/star.svg);
+  background-repeat: no-repeat;
+  background-position: center;
+}
+
+.stars__star:last-child {
+  margin-right: 0;
+}
```
Andriy-Shymkiv, line 21: here you can just:
```css
  margin-right: 4px;
}
```

and remove `margin-right: 4px;` from `stars__star` class

Andriy-Shymkiv (approved):
good job! left comments and approved in advance

PR #3280, opened 2023-07-08
layout stars

IwonaIvona:
https://iwonaivona.github.io/layout_stars/
https://iwonaivona.github.io/layout_stars/report/html_report/


DorotaLeniecDev (approved):
Well done :)

PR #3289, opened 2023-07-10
add task solution

kabakTech:
[DEMO LINK](https://kabaktech.github.io/layout_stars/)
[TEST REPORT LINK](https://kabaktech.github.io/layout_stars/report/html_report/)

it fails the test but works fine. I managed to remove margin-right: 4px;


src/style.css
```diff
@@ -0,0 +1,35 @@
+* {
+  margin: 0;
+  padding: 0;
+}
+
+.stars {
+  display: flex;
+  justify-content: space-between;
+  width: 96px;
```
JoyCoffeeAddict, line 9: If you delete `width` and `justify-content`, then the effect will be almost the same, but this time correct. This is probably causing your problems with tests. Please read checklist again and apply point 3.

src/style.css
```diff
@@ -0,0 +1,35 @@
...
+}
+
+.star__wrapper {
```
JoyCoffeeAddict, line 12: You don't have such class in your html, so this selector is unnecessary.

src/style.css
```diff
@@ -0,0 +1,35 @@
...
+  display: flex;
+  width: 96px;
+}
+
+.stars__item {
```
JoyCoffeeAddict, line 17: You need to set fixed margin on each start to position them correctly.

JoyCoffeeAddict (changes requested):

kabakTech:
[DEMO LINK](https://kabaktech.github.io/layout_product-cards/)
[TEST REPORT LINK](https://kabaktech.github.io/layout_product-cards/report/html_report/)

PR #3295, opened 2023-07-11
add task solution

Andrii-OttO:


src/style.css
```diff
@@ -0,0 +1,31 @@
+
+* {
```
vlad-demchuk, line 2: do not use `*` to reset margins
it will reset margins for all elements, it is not good for performance

src/style.css
```diff
@@ -0,0 +1,31 @@
...
+  margin: 0;
+}
+
+body {
+  display: flex;
+  flex-direction: column;
```
vlad-demchuk, line 8: do not style tags, use classes

vlad-demchuk: it would be better to create a separate block for it

src/style.css
```diff
@@ -0,0 +1,31 @@
...
+  margin-right: 4px;
+  background-image: url(images/star-active.svg);
+}
+
+.stars--0 .stars__star:nth-child(n+1),
+.stars--1 .stars__star:nth-child(n+2),
+.stars--2 .stars__star:nth-child(n+3),
+.stars--3 .stars__star:nth-child(n+4),
+.stars--4 .stars__star:nth-child(n+5),
+.stars--5 .stars__star:nth-child(n+6) {
```
vlad-demchuk, line 29: do you need stars--6 modifier?

vlad-demchuk (changes requested):

Andrii-OttO:
fixed remarks, check please

OleziO (approved):

PR #3298, opened 2023-07-12
add stars

FrontKid:
- [DEMO LINK](https://FrontKid.github.io/layout_stars/)
- [TEST REPORT LINK](https://FrontKid.github.io/layout_stars/report/html_report/)


src/index.html
```diff
@@ -1,3 +1,4 @@
+<!-- #region head -->
```
BudnikOleksii, line 1: Remove unnecessary comments

src/style.css
```diff
@@ -0,0 +1,40 @@
...
+}
+
+.stars__star {
+  background-repeat: no-repeat;
+  background-position: center;
+  width: 16px;
+  height: 16px;
+  background-image: url("./images/star.svg");
+
+}
```
BudnikOleksii, line 16: Remove unnecessary space.

src/style.css
```diff
@@ -0,0 +1,40 @@
...
+  background-image: url("./images/star-active.svg");
+}
+
+.stars.stars--4 .stars__star:not(:last-child) {
+  background-image: url("./images/star-active.svg");
+}
+
+.stars.stars--5 .stars__star:nth-child(n) {
+  background-image: url("./images/star-active.svg");
+}
```
BudnikOleksii, line 40: Don't forger about checklist:
2. [CODE STYLE] - If several selectors MUST always have the same styles, group them using , to prevent accidental out ot sync in future

Also, you don't need to specify ".stars" selector every time and you can use Child combinator to avoid so many classes:
https://developer.mozilla.org/en-US/docs/Web/CSS/Child_combinator

BudnikOleksii (changes requested):
Nice start👍🏻
Let's improve your code slightly.

BudnikOleksii (changes requested):
Nice start👍🏻
Left some comments, let's improve your code slightly.

src/style.css
FrontKid: Thank you I made changes but I don't understand what difference btw ">" and "space"  for this problem
Why using > better than using space?

src/style.css
```diff
@@ -0,0 +1,27 @@
...
+
+.stars__star:not(:last-child) {
+  margin-right: 4px;
+}
+
+.stars--1 > :first-child,
+.stars--2 > :nth-child(-1n + 2),
+.stars--3 > :nth-child(-1n + 3),
+.stars--4 > :not(:last-child),
+.stars--5 > :nth-child(n) {
```
tiserett, line 25: looks complicated, we can simplify it ;)

tiserett (changes requested):
LGTM, but you can make it better ;)

Anna-Tkachenko (approved):
Well Done! ⭐️

PR #3299, opened 2023-07-12
add task solution

bkostanetskyi:


src/index.html
```diff
@@ -8,6 +8,52 @@
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Stars</h1>
+    <div class="stars stars--0">
+      <a class="stars__star"></a>
+      <a class="stars__star"></a>
+      <a class="stars__star"></a>
+      <a class="stars__star"></a>
+      <a class="stars__star"></a>
```
denys-danyliuk, line 16: why you have chosen to use `a`? Try to find a better tag for this

src/style.css
```diff
@@ -0,0 +1,41 @@
...
+  background-image: url(images/star-active.svg);
+}
+
+.stars--4 .stars__star:nth-child(-n+4) {
+  background-image: url(images/star-active.svg);
+}
+
+.stars--5 .stars__star:nth-child(-n+5) {
+  background-image: url(images/star-active.svg);
+}
```
denys-danyliuk, line 41: 2'nd point in checklist

src/style.css
```diff
@@ -0,0 +1,41 @@
...
+  display: flex;
+  justify-content: center;
+}
+
+.stars__star {
+  width: 16px;
+  height: 16px;
+  background-image: url(images/star.svg);
+  background-repeat: no-repeat;
+  margin-right: 4px;
```
denys-danyliuk, line 16: but what about the rightmost stars?

denys-danyliuk, line 16: also try to use background-position, maybe it will help you with tests :)

denys-danyliuk (changes requested):
Pay attention to the review, also your demo looks weird😵‍💫
Знімок екрана 2023-07-13 о 11 10 21

src/style.css
```diff
@@ -8,11 +8,16 @@ body {
   justify-content: center;
 }
 
+.stars:last-child {
+  margin-right: 0;
+}
```
Esceype, line 13: Stats this is your Block but you have margin-right inside element of this block so for remove it from last element you should use this pseudo class for element and not for block

Esceype (changes requested):
Perhaps you did not notice the last comment written by @denys-danyliuk , because you did not correct it, it is also indicated in the checklist point 2, we believe that you will easily complete this item from the checklist.

src/style.css
```diff
@@ -0,0 +1,46 @@
...
+  height: 16px;
+  background-image: url(images/star.svg);
+  background-repeat: no-repeat;
+  background-position: center;
+  margin-right: 4px;
+}
+
+.stars__star :last-child {
+  margin-right: 0;
+}
```
Andriy-Shymkiv, line 22: here you can just:
```css
  margin-right: 4px;
} 
```

Andriy-Shymkiv (changes requested):
please request review again after all fixes

src/style.css
bkostanetskyi: коли так пишу чомусь не проходить тести

Esceype (changes requested):
I see that you requested a review but did not make a new commit with the edits.
At the moment you have 3 commits, but it should be 4, since you already have 3 reviews

src/style.css
```diff
@@ -0,0 +1,31 @@
...
+
+.stars--0 .stars__star:nth-child(-n),
```
Andriy-Shymkiv, line 24: you can remove this line, it doesn't affect anything

src/style.css
```diff
@@ -0,0 +1,31 @@
+
+body {
+  margin: 0;
+}
+
+.stars {
+  display: flex;
+  justify-content: center;
```
Andriy-Shymkiv, line 8: you don't need to center stars for the tests

src/style.css
Andriy-Shymkiv: ok, let's try to fix everything and try to pass the tests again with `:not(:last-child)` usage

Andriy-Shymkiv (changes requested):
left comments

Andriy-Shymkiv (approved):
here you go! everything is good. well done!

PR #3301, opened 2023-07-13
add task solution

sergii-stognii:


src/style.css
```diff
@@ -0,0 +1,31 @@
+/* #region page */
+body {
+  margin: 0;
+}
+/* #endregion */
+/* #region stars */
```
witflash, line 6: You can remove any comments

witflash (approved):
Good job!

Pay attention to my comments for your future work.

PR #3302, opened 2023-07-14
Solved task

AndriiKuz:
- [DEMO LINK](https://AndriiKuz.github.io/layout_stars/)
- [TEST REPORT LINK](https://AndriiKuz.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -8,6 +8,52 @@
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Stars</h1>
+    <div class="stars stars--0">
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+    </div>
+
+    <div class="stars stars--1">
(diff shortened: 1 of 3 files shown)


src/index.html
```diff
@@ -8,6 +8,52 @@
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
```
alexander-ignatow, line 10: Better add an additional container here

It is not a universal rule (and even more - it does not affect the task), but as long as the task sounds like implementing some design library element, it is better to have it inside a separate isolated container (just a good habit).

src/style.css
```diff
@@ -0,0 +1,44 @@
...
+
+.stars__star {
+  width: 16px;
+  height: 16px;
+  background-image: url(./images/star.svg);
+  background-position: center;
+  background-repeat: no-repeat;
+}
+
+.stars__star:not(:last-child) {
```
alexander-ignatow, line 18: You can simplify this by 
- setting `margin-right: 4px` to all stars
- setting `margin-right: 0` to `:last-child` star

It is not critical, but in most cases simpler is better.

PR #3310, opened 2023-07-15
Add task solution

ghost:
- [DEMO LINK](https://maks-n.github.io/layout_stars/)
- [TEST REPORT LINK](https://maks-n.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -16,51 +16,51 @@
   <body class="page__body">
     <div class="page__block block">
       <div class="block__stars stars stars--0">
-        <div class="stars__star star--1"></div>
-        <div class="stars__star star--2"></div>
-        <div class="stars__star star--3"></div>
-        <div class="stars__star star--4"></div>
-        <div class="stars__star star--5"></div>
+        <div class="stars__star"></div>
+        <div class="stars__star"></div>
+        <div class="stars__star"></div>
+        <div class="stars__star"></div>
+        <div class="stars__star"></div>
       </div>
 
       <div class="block__stars stars stars--1">
-        <div class="stars__star star--1"></div>
-        <div class="stars__star star--2"></div>
-        <div class="stars__star star--3"></div>
-        <div class="stars__star star--4"></div>
-        <div class="stars__star star--5"></div>
+        <div class="stars__star"></div>
+        <div class="stars__star"></div>
+        <div class="stars__star"></div>
+        <div class="stars__star"></div>
+        <div class="stars__star"></div>
       </div>
 
       <div class="block__stars stars stars--2">
-        <div class="stars__star star--1"></div>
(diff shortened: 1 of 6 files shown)


src/style.css
```diff
@@ -0,0 +1,46 @@
...
+}
+
+/* Selects all childs from +x to last position */
+.stars--0 :nth-child(n + 1),
+.stars--1 :nth-child(n + 2),
+.stars--2 :nth-child(n + 3),
+.stars--3 :nth-child(n + 4),
+.stars--4 :nth-child(n + 5) {
+  background-image: url(images/star.svg);
+}
```
SanyaBratashchuk, line 37: u can apply this rule just for stars__star

src/style.css
```diff
@@ -0,0 +1,46 @@
...
+.block {
+  display: flex;
+  flex-direction: column;
+}
+
+.block__stars {
+  padding-right: 4px;
+}
+
+.stars {
```
SanyaBratashchuk, line 18: move each BEM block to the separate file

SanyaBratashchuk (changes requested):

src/style.css
maks-n: @SanyaBratashchuk, hi

Yes, but it will load star.svg for every star even behind star-active.svg

So, there are 2 options:
1. for sake of page load performance, or
2. to simplify code base

Every option has its own tradeoff, but with which one should I stay?

maks-n: Applied the rule for `.stars__star`

src/style.css
maks-n: Moved.

src/index.html
```diff
@@ -1,13 +1,67 @@
...
+        <div class="stars__star star--4"></div>
+        <div class="stars__star star--5"></div>
+      </div>
+
+      <div class="block__stars stars stars--1">
+        <div class="stars__star star--1"></div>
+        <div class="stars__star star--2"></div>
+        <div class="stars__star star--3"></div>
+        <div class="stars__star star--4"></div>
+        <div class="stars__star star--5"></div>
```
sTorba24, line 31:
```suggestion
        <div class="stars__star"></div>
        <div class="stars__star"></div>
        <div class="stars__star"></div>
        <div class="stars__star"></div>
        <div class="stars__star"></div>
```

sTorba24: not needed to numerate stars

sTorba24 (changes requested):
Almost good! Lets improve it

src/index.html
maks-n: Removed.

DarkMistyRoom (approved):
Good job!

PR #3311, opened 2023-07-15
Develop

ViraMMM:


nazarmatsevych (approved):
Good work but no need to add margin-right to the last star in a row
image

PR #3313, opened 2023-07-16
add task solution

AlyonaTsehelnyk:
https://AlyonaTsehelnyk.github.io/layout_stars/
https://AlyonaTsehelnyk.github.io/layout_stars/report/html_report/

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,37 @@
+body {
+  margin: 0;
+  box-sizing: border-box;
+}
+
+.stars {
+  display: flex;
+
+}
+
+.stars__star {
+  background-image: url("images/star.svg");
+  background-repeat: no-repeat;
+  background-position: center;
+  backface-visibility: visible;
+  margin-right: 4px;
+  margin-top: 0;
+}
+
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,37 @@
...
+.stars--5 .stars__star:nth-child(-n+5) {
+  background-image: url("images/star-active.svg");
+  background-repeat: no-repeat;
+  background-position: center;
+}
+
+div {
+  min-width: 16px;
+  min-height: 16px;
+}
```
nazarmatsevych, line 33: Move it to `stars__star` class and don't use tags for styling next time

src/style.css
```diff
@@ -0,0 +1,37 @@
...
+
+.container {
+  padding: 32px 22px;
+}
```
nazarmatsevych, line 37: remove it, you don't have `container `class in your html

nazarmatsevych (approved):
Good work,
left minor comments,
also no need to add margin-right to the last star in a row
image

PR #3315, opened 2023-07-16
add task solution

Koliras:
- [DEMO LINK](https://Koliras.github.io/layout_stars/)
- [TEST REPORT LINK](https://Koliras.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -8,6 +8,78 @@
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Stars</h1>
+    <div>
+      <ul class="stars stars--0">
+        <li class="stars__star"></li>
+
+        <li class="stars__star"></li>
+
+        <li class="stars__star"></li>
+
+        <li class="stars__star"></li>
+
+        <li class="stars__star"></li>
+      </ul>
+
+      <ul class="stars stars--1">
+        <li class="stars__star"></li>
+
+        <li class="stars__star"></li>
+
+        <li class="stars__star"></li>
+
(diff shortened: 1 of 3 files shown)


src/index.html
```diff
@@ -8,6 +8,78 @@
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Stars</h1>
+    <div>
```
Anna-Tkachenko, line 11: Do you need this wrapper here?

src/style.css
```diff
@@ -0,0 +1,47 @@
+
+body {
+  margin: 0;
+}
+
+.stars {
+  display: flex;
+  list-style: none;
+  padding: 0;
+  margin: 0;
```
Anna-Tkachenko, line 10: you can't set margins to BEM block
if you need to reset default styles for the tag elements, use reset.css file or reset it in the same way, as you did for the `body`

src/style.css
```diff
@@ -0,0 +1,47 @@
...
+  background-image: url(./images/star-active.svg);
+}
+
+.stars--4 > *:nth-child(-n + 4) {
+  background-image: url(./images/star-active.svg);
+}
+
+.stars--5 > *:nth-child(-n + 5) {
+  background-image: url(./images/star-active.svg);
+}
```
Anna-Tkachenko, line 47: Check checklist ⚠️

Anna-Tkachenko (changes requested):

Anna-Tkachenko (approved):
Good job! ⭐️

PR #3319, opened 2023-07-17
Add task solution

davinnchii:


tiserett (changes requested):
you shouldn't use these classes star--active / stars--inactive

add link on the page please

davinnchii:
> 

Sorry, what link should I add? Readme file is changed, link to my gh there is correct

dffUqp (approved):
LGTM:fire:

will be really cool if you add demo link and test report link in your PR like:
image
this will help mentors work faster

PR #3322, opened 2023-07-17
add task solution

polinavafik:
- [DEMO LINK](https://polinavafik.github.io/layout_stars/)
- [TEST REPORT LINK](https://polinavafik.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,40 @@
+body {
+  margin: 0;
+}
+
+.stars {
+  display: flex;
+  /* align-items: center; */
+}
+
+.stars__star {
+  width: 16px;
+  height: 16px;
+  background-image: url(./images/star.svg);
+  background-position: center;
+  background-repeat: no-repeat;
+}
+
+.stars__star:not(:last-child) {
+  margin-right: 4px;
+}
+
+.stars--1 :nth-child(-n+1) {
+  background-image: url(./images/star-active.svg);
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,40 @@
+body {
+  margin: 0;
+}
+
+.stars {
+  display: flex;
+  /* align-items: center; */
```
alexander-ignatow, line 7: Don't keep dead code :)

src/style.css
```diff
@@ -0,0 +1,40 @@
...
+  background-image: url(./images/star-active.svg);
+}
+
+.stars--4 :nth-child(-n+4) {
+  background-image: url(./images/star-active.svg);
+}
+
+.stars--5 :nth-child(-n+5) {
+  background-image: url(./images/star-active.svg);
+}
```
alexander-ignatow, line 40: you can unite these selectors, like so:

```
.stars--1 :nth-child(-n+1),
.stars--2 :nth-child(-n+2),
...,
.stars--5 :nth-child(-n+5) {
}
```

alexander-ignatow (changes requested):
Very good! Just minor polish needed.

src/style.css
polinavafik: that would be wise, thank you 😎

PR #3324, opened 2023-07-17
add task solution

Orpheus29:
[DEMO LINK](https://Orpheus29.github.io/layout_stars/)
[TEST REPORT LINK](https://Orpheus29.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -1,13 +1,88 @@
 <!doctype html>
 <html lang="en">
+
   <head>
     <meta charset="UTF-8">
     <meta name="viewport"
-          content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0">
+      content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0">
     <title>Stars</title>
     <link rel="stylesheet" href="./style.css">
   </head>
+
   <body>
-    <h1>Stars</h1>
+    <div class="stars-container">
+      <ul class="stars stars--0">
+        <li class="stars__star"></li>
+
+        <li class="stars__star"></li>
+
+        <li class="stars__star"></li>
+
+        <li class="stars__star"></li>
+
+        <li class="stars__star"></li>
+      </ul>
+
+      <ul class="stars stars--1">
+        <li class="stars__star star--active"></li>
+
+        <li class="stars__star"></li>
+
+        <li class="stars__star"></li>
+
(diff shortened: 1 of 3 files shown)


src/index.html
```diff
@@ -1,13 +1,88 @@
 <!doctype html>
 <html lang="en">
+
```
BudnikOleksii, line 3: Remove unnecessary empty spaces between parent and child elements.

src/index.html
```diff
@@ -1,13 +1,88 @@
...
+        <li class="stars__star"></li>
+
+        <li class="stars__star"></li>
+      </ul>
+
+      <ul class="stars stars--2">
+        <li class="stars__star star--active"></li>
+
+        <li class="stars__star star--active"></li>
```
BudnikOleksii, line 41: You don't need to specify "star--active" class, class "stars--2" will define how many stars should be active on it and you don't need to change it manually in the future.

src/style.css
```diff
@@ -0,0 +1,31 @@
+body {
+  margin: 0;
+}
+
+/* .stars-container {
+  display: flex;
+  flex-direction: column;
+} */
```
BudnikOleksii, line 8: Remove comments.

src/style.css
```diff
@@ -0,0 +1,31 @@
...
+  background-image: url(images/star.svg);
+}
+
+.stars__star:not(:last-child) {
+  margin-right: 4px;
+}
+
+.star--active {
+  background-image: url(images/star-active.svg);
+}
```
BudnikOleksii, line 31: Remove this class and let's specify active stars in another way.

BudnikOleksii (changes requested):
Nice start👍🏻
Left some comments, let's improve your code.

src/style.css
Orpheus29: Thank you for your attitude, it is very inspiring! ))

src/style.css
```diff
@@ -0,0 +1,30 @@
+body {
+  margin: 0;
+}
+
+.stars {
+  display: flex;
+  margin: 0;
```
Anna-Tkachenko, line 7: you can't set margins to BEM block

Anna-Tkachenko: if you need to reset default margins for the elements, you can create reset.css file for this purpose

Anna-Tkachenko (changes requested):

dffUqp (approved):
Perfect🚀

PR #3328, opened 2023-07-18
add solution

Yevheniia-Sid:


tiserett (changes requested):
you shouldn't use class active / noactive

add link for page and tests

BudnikOleksii (approved):
Great work🚀

PR #3329, opened 2023-07-18
add task solution

irahos:


Anna-Tkachenko (changes requested):
Add demo link and test report link to pull request description

src/style.css
```diff
@@ -0,0 +1,30 @@
...
+}
+
+.stars--active {
+  margin-right: 4px;
+  width: 16px;
+  height: 16px;
+  background-image: url(/images/star-active.svg);
+  background-repeat: no-repeat;
+  background-position: center;
+}
```
dffUqp, line 30: difference between `stars--00` and `stars--active` is only background-image

can we do something like: 
```css
.stars__star {
  margin-right: 4px;
  width: 16px;
  height: 16px;
  background-image: url(/images/star.svg);
  background-repeat: no-repeat;
  background-position: center;
}

.stars--active {
}
```

dffUqp (changes requested):

BudnikOleksii (changes requested):
Nice, but need some fixes.
Let's get rid of the "stars--active" class and unused classes in html.

tiserett (approved):
LGTM 🔥🚀

alexander-ignatow (approved):
👍 💪

PR #3331, opened 2023-07-18
add task solution

Oleh0293:


src/index.html
```diff
@@ -8,6 +8,47 @@
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Stars</h1>
+    <div class="stars stars--0">
+      <div class="stars__star"></div>
```

src/style.css
```diff
@@ -0,0 +1,25 @@
...
+}
+
+.stars {
+  display: flex;
+}
+
+.stars__star {
+  width: 16px;
+  height: 16px;
+  flex-shrink: 0;
```
Anna-Tkachenko, line 12: Do you need this property if you set the fixed sizes for the element?

Anna-Tkachenko (changes requested):

src/index.html
```diff
@@ -8,6 +8,47 @@
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Stars</h1>
+    <nav class="stars stars--0">
+      <section class="stars__star"></section>
```
dffUqp, line 12: [nav](https://www.geeksforgeeks.org/html-nav-tag/) - used for declaring the navigational section
[section](https://www.w3schools.com/tags/tag_section.asp) - used for defines a section in a document

I think in this case we should  `ul` and `li`

dffUqp (changes requested):
LGTM:fire: but needs a one minor fix:smiley:

alexander-ignatow (approved):
Good 👍

PR #3333, opened 2023-07-18
add task solution

ydratsevska:
- [DEMO LINK](https://ydratsevska.github.io/layout_stars/)
- [TEST REPORT LINK](https://ydratsevska.github.io/layout_stars/report/html_report/)


dffUqp (changes requested):
better to use nth-of-type / nth-child pseudo-classes instead of class `active`:smiley:

alexander-ignatow (approved):
well done! 👍

PR #3338, opened 2023-07-18
add task solution

Nikitavyskreb:
- [DEMO LINK](https://Nikitavyskreb.github.io/layout_stars/)
- [TEST REPORT LINK](https://Nikitavyskreb.github.io/layout_stars/report/html_report/)


tiserett (changes requested):
instead, use nth-of-type pseudoclass

roman-mirzoian (approved):
Well done 👍

PR #3339, opened 2023-07-18
add tack solution Layout Stars

ihorid:


src/style.css
```diff
@@ -0,0 +1,49 @@
+html {
+  --star_acive: url(images/star-active.svg);
+  --star: url(images/star.svg);
+}
```
BudnikOleksii, line 4: You don't actually need to specify them as variables, you can use them directly, but it's ok/

src/style.css
```diff
@@ -0,0 +1,49 @@
...
+
+div,
+body {
+  margin: 0;
+  padding: 0;
+}
+
+.stars {
+  display: flex;
+  /* justify-content: center; */
```
BudnikOleksii, line 14: Remove comments.

src/style.css
```diff
@@ -0,0 +1,49 @@
...
+  background-image: var(--star_acive);
+}
+
+.stars--4 > :nth-child(-n + 4) {
+  background-image: var(--star_acive);
+}
+
+.stars--5 > * {
+  background-image: var(--star_acive);
+}
```
BudnikOleksii, line 49: Checklist:
2. [CODE STYLE] - If several selectors MUST always have the same styles, group them using , to prevent accidental out ot sync in future
image

BudnikOleksii (changes requested):
Good job👍🏻
Just a few things need to be fixed.

src/style.css
```diff
@@ -0,0 +1,38 @@
+html {
+  --star_acive: url(images/star-active.svg);
```
tiserett, line 2: acive -> active

src/style.css
```diff
@@ -0,0 +1,38 @@
...
+
+.stars__star {
+  list-style: none;
+  background-position: center;
+  object-fit: cover;
+  display: flex;
+  background-repeat: no-repeat ;
+  width: 16px;
+  height: 16px;
+  background-image: var(--star);
```
tiserett, line 25: read about positioning of your styles inside block, it's preferably to set display, width, margin etc. on the top of block, and then other styles

tiserett (approved):
LGTM 🚀

PR #3341, opened 2023-07-18
need help 0 passed

KlymSudakov:
- [DEMO LINK](https://KlymSudakov.github.io/layout_stars/)
- [TEST REPORT LINK](https://KlymSudakov.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,52 @@
+* {
+  margin: 0;
+  padding: 0;
+  --active_star: url(images/star-active.svg);
+}
+
+.stars__block {
+  align-content: center;
+  justify-content: center;
+  display: flex;
+  flex-direction: row;
+  flex-wrap: wrap;
+  width: 140px;
+  height: 160px;
+}
+
+.stars {
+  display: flex;
+  flex-direction: row;
+  height: 16px;
+}
+
+.stars__star {
+  width: 16px;
+  background-image: url(images/star.svg);
+  background-position: center;
+  background-repeat: no-repeat;
+}
+
+.stars__star:not(:first-child) {
+  margin-left: 4px ;
+}
+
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,52 @@
...
+  background-image: var(--active_star);
+}
+
+.stars--2 :nth-child(-n+2) {
+  background-image: var(--active_star);
+}
+
+.stars--1 :first-child {
+  background-image: var(--active_star);
+}
```
BudnikOleksii, line 52: Pay attention to the [checklist](https://github.com/mate-academy/layout_stars/blob/master/checklist.md):
2) [CODE STYLE] - If several selectors MUST always have the same styles, group them using , to prevent accidental out ot sync in future

BudnikOleksii (changes requested):
Nice job!
Answered in the personal channel, please, check. Also, left a few comments, let's fix them.

src/style.css
```diff
@@ -0,0 +1,36 @@
+* {
+  margin: 0;
+  padding: 0;
+  --active_star: url(images/star-active.svg);
+}
```
BudnikOleksii, line 5: Avoid * selector, actually you don't need url as a variable, but if you want to specify variables use :root selector.
```css
:root {
  --main-color: blue;
}
```

src/style.css
```diff
@@ -0,0 +1,36 @@
...
+
+.stars__block {
+  align-content: center;
+  justify-content: center;
+  display: flex;
+  flex-direction: row;
+  flex-wrap: wrap;
+  height: 96px;
+  width: 100px;
```
BudnikOleksii, line 14: You don't need this to do it flex and specify flex-wrap property, as the block takes 100%.

PR #3342, opened 2023-07-18
add task solution

voronine:
- [DEMO LINK](https://voronine.github.io/layout_stars/)
- [TEST REPORT LINK](https://voronine.github.io/layout_stars/report/html_report/)


tiserett (changes requested):

dffUqp (approved):
LGTM:fire:

PR #3345, opened 2023-07-18
add task solution

okanivan:


src/style.css
```diff
@@ -0,0 +1,25 @@
...
+.stars__star {
+  width: 16px;
+  height: 16px;
+  margin-right: 4px;
+  background-image: url(./images/star.svg);
+  background-repeat: no-repeat;
+  background-position: center;
+}
+
+.stars--1 > :first-child,
```
tiserett, line 19: can we use same approach as below?

tiserett (approved):
LGTM 🔥🔥

PR #3346, opened 2023-07-18
add task solution

valet2102:


src/index.html
```diff
@@ -3,11 +3,57 @@
...
     <meta name="viewport"
-          content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0">
+      content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0">
     <title>Stars</title>
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Stars</h1>
+    <div class="stars stars--0">
+      <div class="stars__star"></div>
```

src/style.css
```diff
@@ -0,0 +1,29 @@
+body {
+  box-sizing: border-box;
```
Anna-Tkachenko, line 2: Do you need this property for the `body` tag? For what purpose?

src/style.css
```diff
@@ -0,0 +1,29 @@
...
+  margin: 0;
+  padding: 0;
```
Anna-Tkachenko, line 4: `body` tag doesn't have any default paddings

Anna-Tkachenko (changes requested):

tiserett (approved):
LGTM 🚀🔥

PR #3347, opened 2023-07-18
stars

vatatan:


tiserett (changes requested):
instead, use nth-of-type pseudoclass

src/index.html
```diff
@@ -8,6 +8,52 @@
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Stars</h1>
+    <div class="stars stars--0">
+      <div class="stars__star"></div>
```

src/style.css
```diff
@@ -0,0 +1,33 @@
...
+  margin: 0;
+}
+
+.stars--0,
+.stars--1,
+.stars--2,
+.stars--3,
+.stars--4,
+.stars--5 {
+  display: flex;
```
Anna-Tkachenko, line 11: can we move this style to the `stars__star` class? why do we need to add it separately?

Anna-Tkachenko (changes requested):

src/style.css
vatatan: we can't add it to stars_star class, but i added it to stars class

dffUqp (approved):
Perfect🚀

PR #3350, opened 2023-07-18
add task solution

kaavka:


src/index.html
```diff
@@ -3,11 +3,84 @@
...
-          content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0">
+    content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0">
     <title>Stars</title>
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Stars</h1>
+    <h1 class="visually-hidden">Stars</h1>
+
+    <div class="star_row stars--0">
```
Anna-Tkachenko, line 13: Use semantic tags instead of `div` ones

src/index.html
```diff
@@ -3,11 +3,84 @@
...
+      <div class="stars__star active"></div>
+
+      <div class="stars__star"></div>
+
+      <div class="stars__star"></div>
+
+      <div class="stars__star"></div>
+    </div>
+
+    <div class="star_row stars--3">
```
Anna-Tkachenko, line 50: `stars--3` modifier can be added to `stars` BEM block
I don't see any `stars` BEM block here

Anna-Tkachenko: Check class naming

src/index.html
```diff
@@ -3,11 +3,84 @@
...
+      <div class="stars__star active"></div>
```
Anna-Tkachenko, line 51: You should manage active stars with modifier for `stars` BEM block
You should not add any additional classes to the elements

Anna-Tkachenko (changes requested):
Add demo link and test report link to pull request description

src/style.css
```diff
@@ -0,0 +1,41 @@
...
+  position: absolute !important;
+  width: 1px !important;
+  height: 1px !important;
+  margin: -1px !important;
+  border: 0 !important;
+  padding: 0 !important;
+  overflow: hidden !important;
+  clip: rect(0,0,0,0) !important;
+  white-space: nowrap !important;
+}
```
dffUqp, line 15: do you really need it?🤔

src/index.html
```diff
@@ -3,11 +3,84 @@
...
+
+      <li class="stars__star"></li>
+
+      <li class="stars__star"></li>
+    </ul>
+
+    <ul class="stars stars--1">
+      <li class="stars__star"></li>
+
+
```
dffUqp, line 28:
```suggestion
```

dffUqp (changes requested):

src/style.css
```diff
@@ -0,0 +1,29 @@
...
+
+.stars--1 :nth-child(-n + 1),
+.stars--2 :nth-child(-n + 2),
+.stars--3 :nth-child(-n + 3),
+.stars--4 :nth-child(-n + 4),
+.stars--5 :nth-child(n) {
+  background-image: url(images/star-active.svg);
+  background-position: center;
+  background-repeat: no-repeat;
```
alexander-ignatow, line 28:
```
background-position: center;
background-repeat: no-repeat;
```

These styles are not needed here as you are already setting them in the `stars__star`

alexander-ignatow (changes requested):
Good job - let's make just one more small polish!

src/index.html
```diff
@@ -3,11 +3,81 @@
...
+    <ul class="stars stars--0">
+      <li class="stars__star"></li>
+
+      <li class="stars__star"></li>
+
+      <li class="stars__star"></li>
+
+      <li class="stars__star"></li>
+
+      <li class="stars__star"></li>
```
tiserett, line 20: when you have few same empty tags - you can avoid adding whitespaces between them

tiserett (approved):
LGTM 🔥

PR #3354, opened 2023-07-18
add task solution

Vannnkof:
- [DEMO LINK](https://Vannnkof.github.io/layout_stars/)
- [TEST REPORT LINK](https://Vannnkof.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -8,6 +8,47 @@
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Stars</h1>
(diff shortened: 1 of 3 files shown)


src/index.html
```diff
@@ -8,6 +8,47 @@
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Stars</h1>
+    <dev class="stars stars--0">
+      <div class="stars__star"></div>
```

Anna-Tkachenko (approved):

PR #3358, opened 2023-07-18
Complited task

Wesses:
- [DEMO LINK](https://Wesses.github.io/layout_stars/)
- [TEST REPORT LINK](https://Wesses.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -8,6 +8,47 @@
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Stars</h1>
+    <div class="stars stars--0">
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+    </div>
+    <div class="stars stars--1">
+      <div class="stars__star stars__star-active"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+    </div>
+    <div class="stars stars--2">
+      <div class="stars__star stars__star-active"></div>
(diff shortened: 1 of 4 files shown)


tiserett (changes requested):

src/index.html
```diff
@@ -8,6 +8,52 @@
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Stars</h1>
+    <div class="stars stars--0">
+      <div class="stars__star"></div>
```

src/style.css
```diff
@@ -0,0 +1,48 @@
+html {
+  margin: 0;
+  padding: 0;
+  box-sizing: border-box;
+}
```
Anna-Tkachenko, line 5: `html` tag does not have any default margins or paddings
remove styles for `html`

src/style.css
```diff
@@ -0,0 +1,48 @@
...
+  background-image: url("./images/star-active.svg");
+}
+
+.stars--5 :nth-of-type(-n + 5) {
+  background-image: url("./images/star-active.svg");
+}
+
+.stars__star:last-child {
+  margin-right: 0;
+}
```
Anna-Tkachenko, line 48: Check checklist ⚠️

Anna-Tkachenko (changes requested):

src/index.html
```diff
@@ -8,6 +8,76 @@
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Stars</h1>
+    <section class="stars stars--0">
+      <p class="stars__star"></p>
+
```
roman-mirzoian, line 13: Do not forget about good formatting, do not leave unnecessary empty lines.

roman-mirzoian (approved):
Well done 👍

PR #3359, opened 2023-07-18
add task solution

ave-phil:
- [DEMO LINK](https://ave-phil.github.io/layout_stars/)
- [TEST REPORT LINK](https://ave-phil.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -8,6 +8,52 @@
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Stars</h1>
+    <div class="stars stars--0">
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+    </div>
+
+    <div class="stars stars--1">
+      <div class="stars__star stars__star--active"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+    </div>
+
+    <div class="stars stars--2">
+      <div class="stars__star stars__star--active"></div>
+      <div class="stars__star stars__star--active"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+    </div>
+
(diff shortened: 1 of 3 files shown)


src/index.html
```diff
@@ -8,6 +8,52 @@
...
+    <div class="stars stars--0">
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+    </div>
+
+    <div class="stars stars--1">
+      <div class="stars__star stars__star--active"></div>
```
BudnikOleksii, line 20: You don't need to specify "star--active" class, the class "stars--1" will define how many stars should be active on it and you don't need to change it manually in the future.

src/style.css
```diff
@@ -0,0 +1,29 @@
...
+}
+
+.stars--0,
+.stars--1,
+.stars--2,
+.stars--3,
+.stars--4,
+.stars--5 {
+  display: flex;
+}
```
BudnikOleksii, line 12: You can apply this property to class "stars" as all starts have the same.

src/style.css
```diff
@@ -0,0 +1,29 @@
...
+  margin-right: 4px;
+}
+
+.stars__star:last-child {
+  margin-right: 0;
+}
+
+.stars__star--active {
+  background-image: url(./images/star-active.svg);
+}
```
BudnikOleksii, line 29: Remove this class and let's specify active stars in another way.

BudnikOleksii (changes requested):
Nice start👍🏻
Left a few comments, take a look. Let's get rid of active class.

src/index.html
```diff
@@ -8,6 +8,52 @@
...
   </head>
   <body>
-    <h1>Stars</h1>
+    <div class="stars stars--0">
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+    </div>
```
dffUqp, line 17: let's use semantic tags instead of `div`

dffUqp (changes requested):

PR #3361, opened 2023-07-18
add task solution

Vitalii-Petruk:
- [DEMO LINK](https://Vitalii-Petruk.github.io/layout_stars/)
- [TEST REPORT LINK](https://Vitalii-Petruk.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,28 @@
+body {
+  margin: 0;
+}
+
+.stars--0,
+.stars--1,
+.stars--2,
+.stars--3,
+.stars--4,
+.stars--5 {
+  display: flex;
+}
+
+.stars__star {
+  height: 16px;
+  width: 16px;
+  background-image: url(./images/star.svg);
+  background-repeat: no-repeat;
+  background-position: center;
+}
+
+.stars__star:not(:last-child) {
+  margin-right: 4px;
+}
+
+.stars__star.active-star {
+  background-image: url(./images/star-active.svg);
+}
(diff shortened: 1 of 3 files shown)


tiserett (changes requested):

src/style.css
```diff
@@ -0,0 +1,44 @@
...
+  background-image: url(./images/star-active.svg);
+}
+
+.stars--4 .stars__star:nth-child(-n + 4) {
+  background-image: url(./images/star-active.svg);
+}
+
+.stars--5 .stars__star:nth-child(-n + 5) {
+  background-image: url(./images/star-active.svg);
+}
```
alexander-ignatow, line 44: You can unite these definitions 

```suggestion
.stars--1 .stars__star:nth-child(1) ,
.stars--2 .stars__star:nth-child(-n + 2),
...
 {
  background-image: url(./images/star-active.svg);
}

```

src/style.css
```diff
@@ -0,0 +1,44 @@
...
+
+.stars--0,
+.stars--1,
+.stars--2,
+.stars--3,
+.stars--4,
+.stars--5 {
+  display: flex;
+}
```
alexander-ignatow, line 12: You are misusing the modifier: class `stars` is needed here
```suggestion
.stars {
  display: flex;
}
```

alexander-ignatow (changes requested):
Good progress 👍 

Little left to go

roman-mirzoian (approved):
Well done 👍

PR #3362, opened 2023-07-18
add task solution

NazarVotkanych:


src/style.css
```diff
@@ -0,0 +1,42 @@
...
+  background-image: url(./images/star-active.svg);
+}
+
+.stars--4 :nth-child(-n+4) {
+  background-image: url(./images/star-active.svg);
+}
+
+.stars--5 :nth-child(-n+5) {
+  background-image: url(./images/star-active.svg);
+}
```
tiserett, line 42: think how you can simplify this ;)

tiserett (changes requested):
LGTM, but you can make it better ;)

roman-mirzoian (approved):
Well done 👍

PR #3363, opened 2023-07-18
add solution for stars

pzwolenik:
- [DEMO LINK](https://pzwolenik.github.io/layout_stars/)
- [TEST REPORT LINK](https://pzwolenik.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -1,5 +1,5 @@
 <!doctype html>
-<html lang="en">
+<html lang="en" class="page">
   <head>
     <meta charset="UTF-8">
     <meta name="viewport"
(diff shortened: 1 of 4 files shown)


src/style.css
```diff
@@ -0,0 +1,25 @@
...
+}
+
+.stars {
+  display: flex;
+}
+
+.stars__star {
+  width: 16px;
+  height: 16px;
+  margin-right: 4px;
```
Viktor-Kost, line 12: last child shouldn't have margin-right. Use `:not(:last-child)`

src/style.css
```diff
@@ -0,0 +1,25 @@
...
+
+  background-image: url(./images/star.svg);
+  background-repeat: no-repeat;
+  background-position: center;
```
Viktor-Kost, line 16: you can apply background `shorthand`

src/index.html
```diff
@@ -1,13 +1,59 @@
...
   <head>
     <meta charset="UTF-8">
     <meta name="viewport"
           content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0">
     <title>Stars</title>
     <link rel="stylesheet" href="./style.css">
   </head>
-  <body>
-    <h1>Stars</h1>
+  <body class="page">
```
Viktor-Kost, line 10: no needs to add a class to a body. You can use just a tag selector

Viktor-Kost (changes requested):
Nice!
Some minor fixes needed

src/style.css
pzwolenik: I didn't know about this solution, thanks! :)

src/style.css
pzwolenik: Hi! Thank you very much :) 

Isn't it easier, however, to use a `:last-child?` Like this:

`.stars__star:last-child {
  margin-right: 0;
}`

DorotaLeniecDev (approved):
Good job 💪

PR #3364, opened 2023-07-18
add task solution

AntonHerasymov:
[DEMO LINK](https://AntonHerasymov.github.io/layout_stars/)
-[TEST REPORT LINK](https://AntonHerasymov.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -8,6 +8,54 @@
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Stars</h1>
+    <div class="stars__container">
+      <div class="stars__row stars--0">
+        <div class="star"></div>
+        <div class="star"></div>
+        <div class="star"></div>
+        <div class="star"></div>
+        <div class="star"></div>
+      </div>
+
+      <div class="stars__row stars--1">
+        <div class="star"></div>
+        <div class="star"></div>
+        <div class="star"></div>
+        <div class="star"></div>
+        <div class="star"></div>
+      </div>
+
+      <div class="stars__row stars--2">
+        <div class="star"></div>
+        <div class="star"></div>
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,40 @@
...
+  background-image: url(./images/star-active.svg);
+}
+
+.stars--4 :nth-child(-n+4) {
+  background-image: url(./images/star-active.svg);
+}
+
+.stars--5 :nth-child(-n+5) {
+  background-image: url(./images/star-active.svg);
+}
```
dffUqp, line 40: you can combine these selectors
```suggestion
.stars--1 .stars__star:nth-child(-n+1),
.stars--2 .stars__star:nth-child(-n+2),
.stars--3 .stars__star:nth-child(-n+3),
.stars--4 .stars__star:nth-child(-n+4),
.stars--5 .stars__star:nth-child(-n+5) {
  background-image: url(./images/star-active.svg);
}
```

dffUqp (changes requested):

src/index.html
```diff
@@ -8,6 +8,54 @@
...
   <body>
-    <h1>Stars</h1>
+    <div class="stars__container">
+      <div class="stars__row stars--0">
+        <div class="star"></div>
+        <div class="star"></div>
+        <div class="star"></div>
+        <div class="star"></div>
+        <div class="star"></div>
+      </div>
```
Anna-Tkachenko, line 18: I don't see any structure according to BEM methodology
Let's create separate stars block as a row, and each tag inside should be the element of the `stars` block

Anna-Tkachenko (changes requested):

Anna-Tkachenko (approved):
Good job!

PR #3372, opened 2023-07-19
add task solution

oleg-matvienko:
https://matvienko-oleg.github.io/layout_stars/
https://matvienko-oleg.github.io/layout_stars/report/html_report/

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -8,6 +8,47 @@
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Stars</h1>
+    <div class="stars stars--0">
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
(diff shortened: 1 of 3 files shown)


src/index.html
```diff
@@ -8,6 +8,47 @@
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Stars</h1>
+    <div class="stars stars--0">
+      <div class="stars__star"></div>
```

Anna-Tkachenko (changes requested):
Check test report link, it redirects me to 404 error page

src/index.html
Anna-Tkachenko: Will be better to use `ul` and `li` elements as it's more like the list of stars

Anna-Tkachenko (approved):

PR #3375, opened 2023-07-19
made stars

Aserg1992:
- [DEMO LINK](https://Aserg1992.github.io/layout_stars/)
- [TEST REPORT LINK](https://Aserg1992.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -8,6 +8,52 @@
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Stars</h1>
+    <div class="stars--0">
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+    </div>
+
+    <div class="stars--1">
+      <div class="stars__star active"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+    </div>
+
+    <div class="stars--2">
+      <div class="stars__star active"></div>
+      <div class="stars__star active"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+    </div>
+
+    <div class="stars--3">
+      <div class="stars__star active"></div>
+      <div class="stars__star active"></div>
+      <div class="stars__star active"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+    </div>
+
+    <div class="stars--4">
+      <div class="stars__star active"></div>
+      <div class="stars__star active"></div>
+      <div class="stars__star active"></div>
+      <div class="stars__star active"></div>
+      <div class="stars__star"></div>
+    </div>
+
+    <div class="stars--5">
+      <div class="stars__star active"></div>
+      <div class="stars__star active"></div>
+      <div class="stars__star active"></div>
+      <div class="stars__star active"></div>
+      <div class="stars__star active"></div>
+    </div>
   </body>
 </html>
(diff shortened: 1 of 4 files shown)


src/index.html
```diff
@@ -8,6 +8,52 @@
...
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+    </div>
+
+    <div class="stars--1">
+      <div class="stars__star active"></div>
+      <div class="stars__star"></div>
```
BudnikOleksii, line 21: It's better to use nth-of-type / nth-child pseudo-classes instead of class active in this case.

src/index.html
```diff
@@ -8,6 +8,52 @@
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Stars</h1>
+    <div class="stars--0">
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
```
BudnikOleksii, line 13:
```suggestion
    <div class="stars stars--0">
      <div class="stars__star"></div>
      <div class="stars__star"></div>
```
You can't use modifier class without block class in BEM

src/style.css
```diff
@@ -0,0 +1,27 @@
...
+.stars--0,
+.stars--1,
+.stars--2,
+.stars--3,
+.stars--4,
+.stars--5 {
+  display: flex;
+  justify-content: center;
+  padding-top: 2px;
+}
```
BudnikOleksii, line 14: You must have a block class such as "stars" to use a modifier, and if you will have it you can apply these properties to it and don't repeat them every time.

src/style.css
```diff
@@ -0,0 +1,27 @@
...
+  width: 14px;
+  background-image: url(../src/images/star.svg);
+  background-repeat: no-repeat;
+  margin: 0 2px;
+  padding: 0 2px;
+}
+
+.active {
+  background-image: url(../src/images/star-active.svg);
+}
```
BudnikOleksii, line 27: Let's get rid of active class.

BudnikOleksii (changes requested):
Nice start!
Let's get rid of the "active" class and also create a block class for our starts.

src/index.html
Aserg1992: Problem is fixed

src/index.html
Aserg1992: Problem is fixed

src/style.css
Aserg1992: Problem is fixed

src/style.css
Aserg1992: Problem is fixed

src/style.css
```diff
@@ -0,0 +1,22 @@
...
+  margin-right: 4px;
+  background: url(images/star.svg) no-repeat center;
+}
+
+.stars--1 .stars__star:nth-child(1),
+.stars--2 .stars__star:nth-child(-n+2),
+.stars--3 .stars__star:nth-child(-n+3),
+.stars--4 .stars__star:nth-child(-n+4),
+.stars--5 .stars__star:nth-child(-n+5) {
+  background: url(images/star-active.svg) no-repeat center;
```
alexander-ignatow, line 21: No need to alter whole background style - be more specific and change only what you need

```suggestion
background-image: url(../src/images/star-active.svg);
```

alexander-ignatow (changes requested):
So far - almost everything is fixed - good job!

Please fix one final issue and make sure to redeploy your latest solution - for now you still have some old solution deployed

src/style.css
Aserg1992: Done

Aserg1992: Thank you! Done

alexander-ignatow (approved):
LTGM

Good job!

PR #3383, opened 2023-07-22
Star task 

Masha-Cactus:
- [DEMO LINK](https://Masha-Cactus.github.io/layout_stars/)
- [TEST REPORT LINK](https://Masha-Cactus.github.io/layout_stars/report/html_report/)


src/index.html
```diff
@@ -5,9 +5,84 @@
...
+        <div class="stars__star"></div>
+
+        <div class="stars__star"></div>
+
+        <div class="stars__star"></div>
+      </div>
+    </div>
+
+    <!-- Every page should contain H1 for searching system and for screen readers. That`s why I left it -->
+    <h1 class="visually-hidden">Stars</h1>
```
xiimiik, line 86: Yes, semantically it is correct, but in this case we are laying out a block that will be reused, so it is not needed here

xiimiik (approved):

PR #3385, opened 2023-07-22
stars

Exess94:


nazarmatsevych (approved):
Great work but no need to add margin-right to the last stars in a row, you can fix it using :last-child

PR #3390, opened 2023-07-24
add task solution

LadyLiza:
- [DEMO LINK](https://LadyLiza.github.io/layout_stars/)
- [TEST REPORT LINK](https://LadyLiza.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,25 @@
+body {
+  margin: 0;
+}
+
+.stars {
+  display: flex;
+}
+
+.stars__star {
+  width: 12px;
+  height: 12px;
+  padding: 0;
+  margin: 2px 4px;
+  background-image: url("images/star.svg");
+  background-size: cover;
+  transform: translateX(-2px);
+}
+
+.stars--5 .stars__star,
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,25 @@
...
+
+.stars__star {
+  width: 12px;
+  height: 12px;
+  padding: 0;
+  margin: 2px 4px;
+  background-image: url("images/star.svg");
+  background-size: cover;
+  transform: translateX(-2px);
+}
```
Viktor-Kost, line 17: 1. no needs to reset padding;
2. transform is redundant
3. use one side margin with `:not(:last-child)`
4. perhaps you'll need to add background-repeat: no-repeat;
```suggestion
.stars__star {
  width: 12px;
  height: 12px;
  background-image: url("images/star.svg");
  background-size: cover;
}

.stars__star:not(:last-child) {
  margin-right: 4px;
}
```

Viktor-Kost (changes requested):
Nice!
Some moments can be improved

volodymyr-soltys97 (approved):
Well done 🔥

PR #3394, opened 2023-07-25
add task solution

Daniil-A4:
- [DEMO LINK](https://Daniil-A4.github.io/layout_stars/)
- [TEST REPORT LINK](https://Daniil-A4.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,29 @@
+body {
+  margin: 0;
+}
+
+.srars {
+  display: flex;
+}
+
+.stars__star {
+  background: url(images/star.svg);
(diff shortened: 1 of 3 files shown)


xiimiik (changes requested):

src/style.css
```diff
@@ -0,0 +1,28 @@
+body {
+  margin: 0;
+}
+
+.srars {
```
Esceype, line 5: Some minor comment )
What happens with this class name ?

Esceype (approved):
Perfect!

PR #3410, opened 2023-07-28
add task solution

Cy90sure:
- [DEMO LINK](https://Cy90sure.github.io/layout_stars/)
- [TEST REPORT LINK](https://Cy90sure.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -7,7 +7,54 @@
     <title>Stars</title>
     <link rel="stylesheet" href="./style.css">
   </head>
+
   <body>
-    <h1>Stars</h1>
+    <stars class="stars stars--0">
+      <star class="stars__star"></star>
+      <star class="stars__star"></star>
+      <star class="stars__star"></star>
+      <star class="stars__star"></star>
+      <star class="stars__star"></star>
+    </stars>
+
+    <stars class="stars stars--1">
+      <star class="stars__star"></star>
+      <star class="stars__star"></star>
+      <star class="stars__star"></star>
+      <star class="stars__star"></star>
+      <star class="stars__star"></star>
+    </stars>
+
+    <stars class="stars stars--2">
+      <star class="stars__star"></star>
+      <star class="stars__star"></star>
+      <star class="stars__star"></star>
+      <star class="stars__star"></star>
+      <star class="stars__star"></star>
+    </stars>
+
+    <stars class="stars stars--3">
+      <star class="stars__star"></star>
+      <star class="stars__star"></star>
+      <star class="stars__star"></star>
+      <star class="stars__star"></star>
+      <star class="stars__star"></star>
+    </stars>
+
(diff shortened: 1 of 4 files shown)


src/style.css
```diff
@@ -0,0 +1,44 @@
...
+  background-image: url(images/star-active.svg);
+}
+
+.stars--4 :nth-child(-n+4) {
+  background-image: url(images/star-active.svg);
+}
+
+.stars--5 :nth-child(-n+5) {
+  background-image: url(images/star-active.svg);
+}
```
xiimiik, line 44: ❗ Checklist point number '2 ❗

src/index.html
```diff
@@ -7,7 +7,54 @@
...
+
   <body>
-    <h1>Stars</h1>
+    <stars class="stars stars--0">
+      <star class="stars__star"></star>
+      <star class="stars__star"></star>
+      <star class="stars__star"></star>
+      <star class="stars__star"></star>
+      <star class="stars__star"></star>
+    </stars>
```
xiimiik, line 18: the `star` and `stars` tags do not exist, they must be existing tags for semantics

src/style.css
```diff
@@ -0,0 +1,44 @@
...
+  margin: 0;
+}
+
+.stars {
+  display: flex;
+  margin: 0;
+  padding: 0;
+  background-color: #fff;
+  width: fit-content;
+}
```
xiimiik, line 11: it's not a list, so we don't need to reset the styles for it
```suggestion
.stars {
  display: flex;
}
```

src/style.css
```diff
@@ -0,0 +1,44 @@
...
+
+.stars__star {
+  height: 16px;
+  width: 16px;
+  margin-right: 4px;
```
xiimiik, line 16: the outermost star should not be indented
image

xiimiik (changes requested):
GJ, left few commensts!

src/style.css
Cy90sure: How can I rewrite this code to make it look better? Each style has its own formula (-n+0, -n+1, -n+2 etc), so I don't know to change it properly.

src/style.css
```diff
@@ -0,0 +1,29 @@
...
+  margin: 0;
+}
+
+.stars {
+  display: flex;
+}
+
+.stars :last-child {
+  margin: 0;
+}
```
Andriy-Shymkiv, line 11: you can just 
```css
.stars__star:last-child {
  margin-right: 0;
}
```

Andriy-Shymkiv: but move it under stars__star styles

Andriy-Shymkiv (changes requested):

Andriy-Shymkiv (approved):
good job!

PR #3413, opened 2023-07-29
add task solution

SerhiiUnhurian:
- [DEMO LINK](https://SerhiiUnhurian.github.io/layout_stars/)
- [TEST REPORT LINK](https://SerhiiUnhurian.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,36 @@
+.page {
+  margin: 0;
+  padding: 0;
+}
+
+.stars {
+  display: flex;
+  justify-content: center;
+  align-items: center;
+}
+
+.stars__star {
+  height: 16px;
+  width: 16px;
+  margin: 0 2px;
+  background-image: url(./images/star.svg);
+  background-repeat: no-repeat;
+  background-position: center;
+}
+
+.stars :first-child {
+  margin-left: 0;
+}
+
+.stars :last-child {
+  margin-right: 0;
+}
+
+.stars--0 :nth-child(0),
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,36 @@
...
+.stars {
+  display: flex;
+  justify-content: center;
+  align-items: center;
+}
+
+.stars__star {
+  height: 16px;
+  width: 16px;
+  margin: 0 2px;
```
denys-danyliuk, line 15: use margin right instead

src/style.css
```diff
@@ -0,0 +1,36 @@
...
+  background-position: center;
+}
+
+.stars :first-child {
+  margin-left: 0;
+}
+
+.stars :last-child {
+  margin-right: 0;
+}
```
denys-danyliuk, line 27: if you will use margin-right than you will need only one of them

denys-danyliuk (changes requested):
you probably forgot to deploy your changes because your demo link as well as test report link aren't working

src/style.css
SerhiiUnhurian: okay)

src/style.css
SerhiiUnhurian: done)

SerhiiUnhurian:
changed also this
  margin-right: 4px;
}
but tests still fail

Andriy-Shymkiv (approved):
good job!

PR #3423, opened 2023-08-01
add task solution

noraaudonnet:
https://noraaudonnet.github.io/layout_stars/
https://noraaudonnet.github.io/layout_stars/report/html_report/


lerastarynets (approved):
You're showing incredible skills, keep it going!

PR #3425, opened 2023-08-01
add task solution

Metra25:
- [DEMO LINK](https://Metra25.github.io/layout_stars/)
- [TEST REPORT LINK](https://Metra25.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -7,7 +7,6 @@ body {
 }
 
 .stars__star {
-  display: flex;
   background-image: url(images/star.svg);
   background-repeat: no-repeat;
   background-position: center;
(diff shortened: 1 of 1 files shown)


src/style.css
```diff
@@ -0,0 +1,40 @@
+body {
+  margin: 0;
+}
+
+.stars {
+  display: flex;
+}
+
+.stars__star {
+  display: flex;
```
TarasHoliuk, line 10: redundant

TarasHoliuk (changes requested):
Very good
I added 2 small comments

xiimiik (approved):
Nice work!

PR #3428, opened 2023-08-02
Add task solution

AndriyBabiy:
- [DEMO LINK](https://AndriyBabiy.github.io/layout_stars/)
- [TEST REPORT LINK](https://AndriyBabiy.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -8,6 +8,78 @@
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Stars</h1>
+    <div>
+      <div class="stars stars--0">
+        <div class="stars__star stars__star-inactive"></div>
+
+        <div class="stars__star stars__star-inactive"></div>
+
+        <div class="stars__star stars__star-inactive"></div>
+
+        <div class="stars__star stars__star-inactive"></div>
+
+        <div class="stars__star stars__star-inactive"></div>
+      </div>
+
+      <div class="stars stars--1">
+        <div class="stars__star stars__star-active"></div>
+
+        <div class="stars__star stars__star-inactive"></div>
+
+        <div class="stars__star stars__star-inactive"></div>
+
+        <div class="stars__star stars__star-inactive"></div>
+
+        <div class="stars__star stars__star-inactive"></div>
+      </div>
+
+      <div class="stars stars--2">
+        <div class="stars__star stars__star-active"></div>
+
+        <div class="stars__star stars__star-active"></div>
+
+        <div class="stars__star stars__star-inactive"></div>
+
+        <div class="stars__star stars__star-inactive"></div>
+
+        <div class="stars__star stars__star-inactive"></div>
+      </div>
+
+      <div class="stars stars--3">
+        <div class="stars__star stars__star-active"></div>
+
+        <div class="stars__star stars__star-active"></div>
+
+        <div class="stars__star stars__star-active"></div>
+
+        <div class="stars__star stars__star-inactive"></div>
+
+        <div class="stars__star stars__star-inactive"></div>
+      </div>
+
+      <div class="stars stars--4">
+        <div class="stars__star stars__star-active"></div>
+
+        <div class="stars__star stars__star-active"></div>
+
+        <div class="stars__star stars__star-active"></div>
+
+        <div class="stars__star stars__star-active"></div>
+
+        <div class="stars__star stars__star-inactive"></div>
+      </div>
+
+      <div class="stars stars--5">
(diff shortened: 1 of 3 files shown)


src/index.html
```diff
@@ -8,6 +8,78 @@
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Stars</h1>
+    <div>
+      <div class="stars stars--0">
+        <div class="stars__star stars__star-inactive"></div>
```
mykhalenych, line 13: Screenshot 2023-08-02 at 16 12 23
```suggestion
        <div class="stars__star"></div>
```

src/style.css
```diff
@@ -0,0 +1,29 @@
...
+  display: flex;
+}
+
+.stars__star {
+  display: flex;
+  height: 16px;
+  width: 16px;
+  background-position: center;
+  background-repeat: no-repeat;
+  margin-right: 4px;
```
mykhalenych, line 16:
```suggestion
  margin-right: 4px;
  background-image: url(images/star.svg);
```

mykhalenych (changes requested):

src/index.html
AndriyBabiy: Adjusted the previous element of “stars__star-inactive” to “stars__star--inactive” to match the modifier styling.

src/style.css
AndriyBabiy: Removed the "stars__star--active" modifier and have that built into the "stars__star" as a default background image.

src/index.html
mykhalenych: Add one of the next extra classes (a modifier) stars--0, stars--1, stars--2 ... stars--5 to each block

you already added this modifier line before

mykhalenych (changes requested):
missed requirements, if you have questions add them to chat

src/index.html
AndriyBabiy: I just tried adding the stars--0 modifier to each of the blocks, in addition to the addition to the modifier in line 12, and got numerous errors. 
image

AndriyBabiy: Reclassified the parent div as "stars" and then the subsequenqent child blocks were given the modifiers of "stars--0" (...).
Also restructured the CSS element to attribute "display: flex" and "flex-direction: row" to each of the stars--N modifiers

AndriyBabiy: if this still misses the requirements, would you possibly have a different example of what this question is looking for as I am not 100% sure what I should be doing to fully fill the requirements of the task.

src/style.css
```diff
@@ -0,0 +1,28 @@
...
+
+.stars--0,
+.stars--1,
+.stars--2,
+.stars--3,
+.stars--4,
+.stars--5 {
+  display: flex;
+  flex-direction: row;
+}
```
Andriy-Shymkiv, line 14: just give one common class for all block, for ex. `stars`, and then you can go like this:
```css
.stars {
  display: flex;
}
```
and no need to set flex-direction: row, because it is row by default

src/style.css
```diff
@@ -0,0 +1,28 @@
...
+  width: 16px;
+  background-position: center;
+  background-repeat: no-repeat;
+  margin-right: 4px;
+  background-image: url(images/star-active.svg);
+}
+
+.stars__star--inactive {
+  background-image: url(images/star.svg);
+}
```
Andriy-Shymkiv, line 28: do this, but for each block (stars--0, stars--1...), and use `nth-child()` as i mentioned above

Andriy-Shymkiv (changes requested):

src/style.css
Andriy-Shymkiv: please, for better understanding of what you need to do - read requirements and checklist first

src/style.css
```diff
@@ -0,0 +1,38 @@
...
+  background-image: url(images/star-active.svg);
+}
+
+.stars--0 :nth-child(n + 1) {
+  background-image: url(images/star.svg);
+}
+
+.stars--1 :nth-child(n + 2) {
+  background-image: url(images/star.svg);
+}
```
TarasHoliuk, line 26:
```suggestion
.stars--0 :nth-child(n + 1),
.stars--1 :nth-child(n + 2) {
  background-image: url(images/star.svg);
}
```
you may join selectors with same styles (as you did before) - FYI

TarasHoliuk (approved):
GJ :+1: - Approved :heavy_check_mark:

src/style.css
AndriyBabiy: Ok, right. Have implemented! TYSM!

PR #3430, opened 2023-08-02
add task solution

DudeMood:
https://DudeMood.github.io/layout_stars/
https://DudeMood.github.io/layout_stars/report/html_report/

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -16,7 +16,6 @@ body {
 .stars__star {
   width: 16px;
   height: 16px;
-  margin: auto;
   background-image: url(../src/images/star.svg);
   background-position: center;
   background-repeat: no-repeat;
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,36 @@
+body {
+  margin: 0;
+}
+
+.main {
+  display: flex;
+  flex-direction: column;
+}
+
+.stars {
+  display: flex;
+}
+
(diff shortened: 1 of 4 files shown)


src/style.css
```diff
@@ -0,0 +1,35 @@
...
+
+.stars--1 .stars__star:nth-child(-n + 1),
+.stars--2 .stars__star:nth-child(-n + 2),
+.stars--3 .stars__star:nth-child(-n + 3),
+.stars--4 .stars__star:nth-child(-n + 4),
+.stars--5 .stars__star:nth-child(-n + 5) {
+  width: 16px;
+  height: 16px;
+  background-image: url(../src/images/star-active.svg);
+  background-repeat: no-repeat;
```
TarasHoliuk, line 34: no need to duplicate all properties
write here the only one you change

src/style.css
```diff
@@ -0,0 +1,35 @@
...
+.main {
+  display: flex;
+  flex-direction: column;
+}
+
+.stars {
+  display: flex;
+}
+
+.stars__star :not(:last-child) {
```
TarasHoliuk, line 14:
```suggestion
.stars__star:not(:last-child) {
```
it does matter :)
with space, it applies to nested in `stars__star` element

TarasHoliuk (changes requested):

PR #3436, opened 2023-08-04
add task solution

iurche:
[DEMO LINK](https://iurche.github.io/layout_stars/)
- [TEST REPORT LINK](https://iurche.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,38 @@
+body {
+  margin: auto;
+}
+
+.stars {
+  display: flex;
+  flex-direction: row;
+}
+
+.stars__star {
+  height: 16px;
+  width: 16px;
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,38 @@
+body {
+  margin: auto;
+}
+
+.stars {
+  display: flex;
+  flex-direction: row;
```
polosanya, line 7: That's actually a default value. No need to specify it, but it's not a problem

polosanya (changes requested):
Great! Almost perfect!
One small but important fix left 🙏

VitaliyBodnar (approved):

PR #3439, opened 2023-08-04
add task solution

Kalkorejz:
- [DEMO LINK](https://Kalkorejz.github.io/layout_stars/)
- [TEST REPORT LINK](https://Kalkorejz.github.io/layout_stars/report/html_report/)

My solution didn't pass the test but I'm very close to the correct result.
Pixels are the same as the layout

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -8,6 +8,54 @@
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Stars</h1>
+    <div class="container">
+      <div class="stars stars--0">
+        <div class="stars__star"></div>
+        <div class="stars__star"></div>
+        <div class="stars__star"></div>
+        <div class="stars__star"></div>
+        <div class="stars__star"></div>
+      </div>
+
+      <div class="stars stars--1">
+        <div class="stars__star active"></div>
+        <div class="stars__star"></div>
+        <div class="stars__star"></div>
+        <div class="stars__star"></div>
+        <div class="stars__star"></div>
+      </div>
+
+      <div class="stars stars--2">
+        <div class="stars__star active"></div>
+        <div class="stars__star active"></div>
+        <div class="stars__star"></div>
+        <div class="stars__star"></div>
+        <div class="stars__star"></div>
+      </div>
+
+      <div class="stars stars--3">
+        <div class="stars__star active"></div>
+        <div class="stars__star active"></div>
+        <div class="stars__star active"></div>
+        <div class="stars__star"></div>
+        <div class="stars__star"></div>
+      </div>
+
+      <div class="stars stars--4">
+        <div class="stars__star active"></div>
+        <div class="stars__star active"></div>
+        <div class="stars__star active"></div>
+        <div class="stars__star active"></div>
+        <div class="stars__star"></div>
+      </div>
+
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,28 @@
+body {
+  margin: 0;
+}
+
+.container {
+  display: inline-block;
+}
```
Esceype, line 7: You don`t need it

src/style.css
```diff
@@ -0,0 +1,28 @@
...
+
+.stars {
+  display: flex;
+  flex-direction: row;
```
Esceype, line 11: This is default value for flex-box so remove it

src/style.css
```diff
@@ -0,0 +1,28 @@
...
+  background-image: url(images/star.svg);
+  background-repeat: no-repeat;
+  height: 16px;
+  width: 16px;
+}
+
+.active {
+  background-image: url(images/star-active.svg);
+  background-repeat: no-repeat;
+}
```
Esceype, line 24: You should add a modifier to stars__star, not a new block
Checklist 2nd point

src/style.css
```diff
@@ -0,0 +1,28 @@
...
+  display: flex;
+  flex-direction: row;
+}
+
+.stars__star {
+  background-image: url(images/star.svg);
+  background-repeat: no-repeat;
+  height: 16px;
+  width: 16px;
+}
```
Esceype, line 19: Also add here margin-right or margin left and background position

Esceype (changes requested):
GJ!!! Read the following comments

src/style.css
Kalkorejz: I don't understand what i need to do here(
How can i get different stars without creating a new block

src/style.css
Kalkorejz: In previous commit i did margin-left for all stars exept first, isn't that more correct?
And why i need to background position here?

Kalkorejz:
Tywm for review, now i passed the test but still have questions

src/style.css
denys-danyliuk: you can use pseudo classes like nth-child

src/index.html
```diff
@@ -8,6 +8,52 @@
...
+      <div class="stars__star active"></div>
+      <div class="stars__star"></div>
+    </div>
+
+    <div class="stars stars--5">
+      <div class="stars__star active"></div>
+      <div class="stars__star active"></div>
+      <div class="stars__star active"></div>
+      <div class="stars__star active"></div>
+      <div class="stars__star active"></div>
```
denys-danyliuk, line 56: don't hardcode it like this, add classes to stars--5 modifier instead

denys-danyliuk (changes requested):

Kalkorejz:
i think i did it

src/style.css
```diff
@@ -0,0 +1,24 @@
...
+  display: flex;
+}
+
+.stars__star {
+  background-image: url(images/star.svg);
+  background-repeat: no-repeat;
+  background-position: center;
+  height: 16px;
+  width: 16px;
+  margin-right: 4px;
```
mykhalenych, line 15: the last star shouldn't have margin-right: 4px

PR #3446, opened 2023-08-06
add task solution

Val-Bryks:


volodymyr-soltys97 (changes requested):
There are no demo links, you need to note them and add them to the description of the PR
image

PR #3465, opened 2023-08-11
layout_stars

DudeMood:
https://DudeMood.github.io/layout_stars/
https://DudeMood.github.io/layout_stars/report/html_report/

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,31 @@
+body {
+  display: flex;
+  flex-direction: column;
+  padding: 0;
+  margin: 0;
+}
+
+.stars {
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,31 @@
...
+  padding: 0;
+  margin: 0;
+}
+
+.stars {
+  display: flex;
+}
+
+.stars__star {
+  box-sizing: border-box;
```
TarasHoliuk, line 13: FYI: always set `box-sizing: border-box;` in `*` selector

TarasHoliuk (approved):

PR #3467, opened 2023-08-12
add task solution

viktoriiadereza:
- [DEMO LINK](https://viktoriiadereza.github.io/vd_layout_stars/)
- [TEST REPORT LINK](https://viktoriiadereza.github.io/vd_layout_stars/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,27 @@
+body {
+  margin: 0;
+}
+
+.stars {
+  display: flex;
+}
+
+.stars__star {
+  display: flex;
+  background-image: url(./images/star.svg);
+  background-repeat: no-repeat;
+  background-position: center;
+  margin-right: 4px;
+  width: 16px;
+  height: 16px;
+
+}
+
+.stars--0 .stars__star:nth-child(-n + 0),
+.stars--1 .stars__star:nth-child(-n + 1),
+.stars--2 .stars__star:nth-child(-n + 2),
+.stars--3 .stars__star:nth-child(-n + 3),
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,27 @@
+body {
+  margin: 0;
+}
+
+.stars {
+  display: flex;
+}
+
+.stars__star {
+  display: flex;
```
lerastarynets, line 10: The .stars__star class is set to display flex, but it doesn't seem to be necessary as it doesn't have any child elements to control the layout of. If this is unintentional, you can remove it to simplify the code.

src/style.css
```diff
@@ -0,0 +1,27 @@
...
+  background-image: url(./images/star.svg);
+  background-repeat: no-repeat;
+  background-position: center;
+  margin-right: 4px;
+  width: 16px;
+  height: 16px;
+
+}
+
+.stars--0 .stars__star:nth-child(-n + 0),
```
lerastarynets, line 20: The :nth-child(-n + 0) selector will never match any elements because it selects elements that are the first child of their parent and also have no preceding siblings, which is impossible. You can remove the .stars--0 style completely as it doesn't affect the styling.

lerastarynets (changes requested):
Nice start! Let's fix and improve your work a little bit)

src/style.css
viktoriiadereza: Thank you! Updated the code.

Andriy-Shymkiv (approved):
well done!

PR #3468, opened 2023-08-12
task: layout_stars

iLavruk:
[DEMO LINK](https://iLavruk.github.io/layout_stars/)
[TEST REPORT LINK](https://iLavruk.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,24 @@
+body {
+  margin: 0;
+}
+
+.stars {
+  display: flex;
+}
+
+.stars__star {
+  width: 16px;
+  height: 16px;
+  display: flex;
+  margin-right: 4px;
+  background: url("images/star.svg") center no-repeat;
+}
+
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,24 @@
...
+
+.stars__star {
+  width: 16px;
+  height: 16px;
+  display: flex;
+  margin-right: 4px;
+  background: url("images/star.svg") center no-repeat;
+}
+
+.stars--0 .stars__star:nth-child(-n + 0),
```
lerastarynets, line 17:
```suggestion
```
The :nth-child(-n + 0) selector will never match any elements because it selects elements that are the first child of their parent and also have no preceding siblings, which is impossible. You can remove the .stars--0 style completely as it doesn't affect the styling.

lerastarynets (approved):
Great job overall! Left one tiny suggestion for the improvement

PR #3479, opened 2023-08-15
add task solution

Daniilart01:
- [DEMO LINK](https://Daniilart01.github.io/layout_stars/)
- [TEST REPORT LINK](https://Daniilart01.github.io/layout_stars/report/html_report/)
My solution fails 6 of 7 test except entire document test. In previous task i had the same trouble and I solved it with the help of a mentor and swaping some blocks in css, but I did not fully understand how I did it. In curent implemetation position of my rows of stars shifts a little to the left compare to the reference, but entire document test passed. This is probably due to `margin` because it is applied to every row, so i decide to set-up `.block` width manualy and use `justify-content: space-between`, but it didn't work for me with the same tests results. I ask to explain me, why does check-system behaves like this, so in the future I do not waste time solving the problem of shifting my implementation to the side (u can check this shift in tests)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -5,9 +5,57 @@
     <meta name="viewport"
           content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0">
     <title>Stars</title>
-    <link rel="stylesheet" href="./style.css">
+    <link rel="stylesheet" href="style.css">
   </head>
   <body>
-    <h1>Stars</h1>
+    <div class="block">
+      <div class="stars stars--0">
+        <div class="stars__star"></div>
+        <div class="stars__star"></div>
+        <div class="stars__star"></div>
+        <div class="stars__star"></div>
+        <div class="stars__star"></div>
+      </div>
+
+      <div class="stars stars--1">
+        <div class="stars__star"></div>
+        <div class="stars__star"></div>
+        <div class="stars__star"></div>
+        <div class="stars__star"></div>
+        <div class="stars__star"></div>
+      </div>
+
+      <div class="stars stars--2">
+        <div class="stars__star"></div>
+        <div class="stars__star"></div>
+        <div class="stars__star"></div>
+        <div class="stars__star"></div>
+        <div class="stars__star"></div>
+      </div>
+
+      <div class="stars stars--3">
+        <div class="stars__star"></div>
+        <div class="stars__star"></div>
+        <div class="stars__star"></div>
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,39 @@
...
+.stars {
+  display: flex;
+  align-items: center;
+  justify-content: space-between;
+}
+
+.stars__star {
+  width: 16px;
+  height: 16px;
+  margin: 0 2px;
```
mykhalenych, line 18:
```suggestion
  margin-right: 4px;
```

src/style.css
```diff
@@ -0,0 +1,39 @@
...
+  background-image: url(images/star.svg);
+  background-position: center;
+  background-repeat: no-repeat;
+}
+
+.stars__star:last-child {
+  margin-right: 0;
+}
+
+.stars .stars__star:first-child {
```
mykhalenych, line 28:
```suggestion
```

mykhalenych (changes requested):

Daniilart01:
Done. I`ve tried this way you suggested before and it doesn't work, as you see

src/style.css
Daniilart01: +

src/style.css
Daniilart01: +

src/index.html
```diff
@@ -5,9 +5,57 @@
     <meta name="viewport"
           content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0">
     <title>Stars</title>
-    <link rel="stylesheet" href="./style.css">
+    <link rel="stylesheet" href="style.css">
   </head>
   <body>
-    <h1>Stars</h1>
+    <div class="block">
```
denys-danyliuk, line 11: you don't need this wrapper

src/style.css
```diff
@@ -0,0 +1,35 @@
...
+  background-image: url(images/star.svg);
+  background-position: center;
+  background-repeat: no-repeat;
+}
+
+.stars__star:last-child {
+  margin-right: 0;
+}
+
+.stars--0 .stars__star:nth-child(-n+0),
```
denys-danyliuk, line 28:
```suggestion
```

denys-danyliuk (approved):

Daniilart01:
Yes, without `.block` and `justify-content: space-between` in `.stars`  it works, but why test-system doesn't work with `.block` wrapping? It looks more structured when group of similar items covered with a separate block.

PR #3494, opened 2023-08-18
add task solution

myehor:


src/style.css
```diff
@@ -0,0 +1,45 @@
...
+  background-image: url(./images/star-active.svg);
+}
+
+.stars--4 :nth-child(n+1):nth-child(-n+4) {
+  background-image: url(./images/star-active.svg);
+}
+
+.stars--5 :nth-child(n+1):nth-child(-n+5) {
+  background-image: url(./images/star-active.svg);
+}
```
etojeDenys, line 45: you can join it with `,`

etojeDenys (changes requested):
don't forget to add demo and test report links

src/style.css
myehor: How can i join it, when for every class i use different properties? And why all of tests are failed?

src/style.css
```diff
@@ -0,0 +1,33 @@
...
+  justify-content: center;
+  align-items: center;
+  width: 96px;
+  height: 96px;
+}
+
+.stars-star {
+  background-image: url(./images/star.svg);
+  background-repeat: no-repeat;
+  background-position: center;
```
loralevitska, line 16: Combine with background property

src/index.html
```diff
@@ -8,6 +8,49 @@
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Stars</h1>
+    <div class="container">
+      <div class="stars stars--0">
+        <div class="stars-star"></div>
```
loralevitska, line 13: Fix it everywhere
```suggestion
        <div class="stars__star"></div>
```

src/style.css
```diff
@@ -0,0 +1,33 @@
...
+  width: 16px;
+  height: 16px;
+}
+
+.stars {
+  display: flex;
+  justify-content: space-between;
```
loralevitska, line 23: remove this property
you should add `margin-right` property for each star except the last one

loralevitska (changes requested):
Almost done) Let's improve
If you have any questions, please ask them in fe_chat and tag mentors. It will be faster, because your answers in GitHub will be visible to mentor only after you re-request review

src/index.html
```diff
@@ -8,6 +8,49 @@
...
+        <div class="stars__star"></div>
+        <div class="stars__star"></div>
+        <div class="stars__star stars__star__last"></div>
+      </div>
+      <div class="stars stars--2">
+        <div class="stars__star"></div>
+        <div class="stars__star"></div>
+        <div class="stars__star"></div>
+        <div class="stars__star"></div>
+        <div class="stars__star stars__star__last"></div>
```
SanyaBratashchuk, line 31:
```suggestion
        <div class="stars__star"></div>
```
use pseudo-class :last-child

SanyaBratashchuk (changes requested):

SanyaBratashchuk (approved):

PR #3521, opened 2023-08-25
add task solution

VikiUA:
I don't know how to center a star in its div 16*16.

https://VikiUA.github.io/layout_stars/

https://VikiUA.github.io/layout_stars/report/html_report/

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -8,6 +8,49 @@
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Stars</h1>
+    <div class="container">
+      <container class="stars stars--0">
+        <div class="stars__star"></div>
+        <div class="stars__star"></div>
+        <div class="stars__star"></div>
+        <div class="stars__star"></div>
+        <div class="stars__star"></div>
+      </container>
+      <container class="stars stars--1">
+        <div class="stars__star-active"></div>
+        <div class="stars__star"></div>
+        <div class="stars__star"></div>
+        <div class="stars__star"></div>
+        <div class="stars__star"></div>
+      </container>
+      <container class="stars stars--2">
+        <div class="stars__star-active"></div>
+        <div class="stars__star-active"></div>
+        <div class="stars__star"></div>
+        <div class="stars__star"></div>
+        <div class="stars__star"></div>
+      </container>
+      <container class="stars stars--3">
+        <div class="stars__star-active"></div>
+        <div class="stars__star-active"></div>
+        <div class="stars__star-active"></div>
+        <div class="stars__star"></div>
+        <div class="stars__star"></div>
+      </container>
+      <container class="stars stars--4">
+        <div class="stars__star-active"></div>
+        <div class="stars__star-active"></div>
+        <div class="stars__star-active"></div>
+        <div class="stars__star-active"></div>
+        <div class="stars__star"></div>
+      </container>
+      <container class="stars stars--5">
+        <div class="stars__star-active"></div>
+        <div class="stars__star-active"></div>
+        <div class="stars__star-active"></div>
+        <div class="stars__star-active"></div>
+        <div class="stars__star-active"></div>
+      </container>
+    </div>
   </body>
 </html>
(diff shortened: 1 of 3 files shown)


src/index.html
```diff
@@ -8,6 +8,49 @@
...
+    <div class="container">
+      <container class="stars stars--0">
+        <div class="stars__star"></div>
+        <div class="stars__star"></div>
+        <div class="stars__star"></div>
+        <div class="stars__star"></div>
+        <div class="stars__star"></div>
+      </container>
+      <container class="stars stars--1">
+        <div class="stars__star-active"></div>
```
anastasiiavorobiova, line 20:
```suggestion
        <div class="stars__star"></div>
```
There is no need to use additional class

src/index.html
```diff
@@ -8,6 +8,49 @@
...
+        <div class="stars__star"></div>
+        <div class="stars__star"></div>
+        <div class="stars__star"></div>
+        <div class="stars__star"></div>
+      </container>
+      <container class="stars stars--2">
+        <div class="stars__star-active"></div>
+        <div class="stars__star-active"></div>
```
anastasiiavorobiova, line 28: The same issue here

src/index.html
```diff
@@ -8,6 +8,49 @@
...
+        <div class="stars__star"></div>
+        <div class="stars__star"></div>
+        <div class="stars__star"></div>
+      </container>
+      <container class="stars stars--3">
+        <div class="stars__star-active"></div>
+        <div class="stars__star-active"></div>
+        <div class="stars__star-active"></div>
```
anastasiiavorobiova, line 36: See my previous comment

src/index.html
```diff
@@ -8,6 +8,49 @@
...
+        <div class="stars__star"></div>
+      </container>
+      <container class="stars stars--5">
+        <div class="stars__star-active"></div>
+        <div class="stars__star-active"></div>
+        <div class="stars__star-active"></div>
+        <div class="stars__star-active"></div>
+        <div class="stars__star-active"></div>
```
anastasiiavorobiova, line 52: The same here

anastasiiavorobiova (changes requested):
Great start! Try to fix the tests, stars should be centered inside the block

src/index.html
VikiUA: But how then can I add two different pictures to the same class?

VikiUA: > But how then can I add two different pictures to the same class?
in class="stars__star-active" all stars are yellow, in another class they are gray

src/style.css
```diff
@@ -0,0 +1,34 @@
...
+}
+
+.stars__star {
+  background-image: url(images/star.svg);
+  width: 16px;
+  height: 16px;
+  background-repeat: no-repeat;
+}
+
+.stars__star-active {
```
loralevitska, line 24: Remove this class for fixing previous comments 
use [nth-child](https://developer.mozilla.org/en-US/docs/Web/CSS/:nth-child)

src/style.css
```diff
@@ -0,0 +1,34 @@
...
+  height: 96px;
+  width: 96px;
+  background-color: #fff;
+}
+
+.stars {
+  display: flex;
+  justify-content: center;
+  align-items: center;
+}
```
loralevitska, line 15: Remove redundant properties
```suggestion
.stars {
  display: flex;
  align-items: center;
}
```

loralevitska (changes requested):
Almost done) Try to pass all tests 
- check stars indents 
image

src/style.css
VikiUA: this command highlights the column at once, I don't know how to use it.

src/style.css
```diff
@@ -0,0 +1,28 @@
...
+  display: flex;
+  align-items: center;
+}
+
+.stars__star {
+  background-image: url(images/star.svg);
+  width: 16px;
+  height: 16px;
+  background-repeat: no-repeat;
+  background-position: center;
```
maxim2310, line 15: would be nice to use combine version of background

maxim2310 (approved):
Good work!!!

PR #3524, opened 2023-08-27
add stars solution

sviatikkondur:
[DEMO LINK](https://sviatikkondur.github.io/layout_stars/)
[TEST REPORT LINK](https://sviatikkondur.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -8,6 +8,52 @@
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Stars</h1>
+    <div class="stars--0 stars">
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+    </div>
+
+    <div class="stars--1 stars">
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+    </div>
+
+    <div class="stars--2 stars">
+      <div class="stars__star"></div>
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,22 @@
...
+}
+
+.stars {
+  display: flex;
+}
+
+.stars__star {
+  width: 16px;
+  height: 16px;
+  margin-right: 4px;
```
polosanya, line 12: image
Make sure that the last elements of the row don't have the right margin

src/index.html
```diff
@@ -8,6 +8,52 @@
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Stars</h1>
+    <div class="stars--0 stars">
```
polosanya, line 11:
```suggestion
    <div class="stars stars--0">
```
It's not critical, but I think it's more logical to set the default class name first, and then the modifier

polosanya (changes requested):
Almost perfect! One small but important fix left

src/style.css
```diff
@@ -0,0 +1,26 @@
...
+  background: url(images/star.svg) center no-repeat;
+}
+
+.stars__star:last-child {
+  margin-right: 0;
+}
```
StasSokolov, line 18: please make sure that this works right

StasSokolov (changes requested):
Great job, let's fix only one minor comments from me!)

src/style.css
StasSokolov: not fixed

src/style.css
sviatikkondur: My problem was that I didn't deploy changes. I thought it was an automatic process.

StasSokolov (approved):
Perfect🔥

PR #3529, opened 2023-08-30
implemented stars

ErnestMk1:
- [DEMO LINK](https://ErnestMk1.github.io/layout_stars/)
- [TEST REPORT LINK](https://ErnestMk1.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,46 @@
+:root {
+  --star-size: 16px;
+}
+
+body {
+  margin: 0;
+  padding: 0;
+}
+
+.stars {
+  padding: 0;
+  margin: 0;
+  list-style: none;
+  display: flex;
+  box-sizing: border-box;
+}
+
+.stars__star {
+  background-position: center;
+  background-repeat: no-repeat;
+  height: var(--star-size);
+  width: var(--star-size);
+  background-image: url(./images/star.svg);
+  padding: 0;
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,46 @@
...
+  margin: 0;
+  padding: 0;
+}
+
+.stars {
+  padding: 0;
+  margin: 0;
+  list-style: none;
+  display: flex;
+  box-sizing: border-box;
```
OleziO, line 15: A better practice is to place this property in a * selector and remove it from all other places.

src/style.css
```diff
@@ -0,0 +1,46 @@
...
+  background-image: url(./images/star-active.svg);
+}
+
+.stars--4 .stars__star:nth-child(-n + 4) {
+  background-image: url(./images/star-active.svg);
+}
+
+.stars--5 .stars__star:nth-child(-n + 5) {
+  background-image: url(./images/star-active.svg);
+}
```
OleziO, line 46: Combine these styles to optimize code.

OleziO (changes requested):

src/style.css
ErnestMk1: are talking about margin? Or box-sizing?

anastasiiavorobiova (approved):
Awesome!

PR #3537, opened 2023-09-01
add task solution

romanyshyns:


src/index.html
```diff
@@ -8,6 +8,54 @@
...
+      <div class="stars stars--0">
+        <div class="stars__star"></div>
+        <div class="stars__star"></div>
+        <div class="stars__star"></div>
+        <div class="stars__star"></div>
+        <div class="stars__star"></div>
+      </div>
+
+      <div class="stars stars--1">
+        <div class="stars__star stars--N"></div>
```
mykhalenych, line 21:
```suggestion
        <div class="stars__star"></div>
```
delete all stars--N classes

mykhalenych: Add one of the next extra classes (a modifier) stars--0, stars--1, stars--2 ... stars--5 to each block
The block with stars--N modifier should have exactly N first stars active

src/style.css
```diff
@@ -0,0 +1,22 @@
+body {
+  margin: 0;
+}
+
+.stars {
+  display: flex;
+}
+
+.stars__star {
+  margin-right: 4px;
```
mykhalenych, line 10: the last star shouldn't have margin-right: 4px;

mykhalenych (changes requested):

src/style.css
romanyshyns: What should he have?

PR #3560, opened 2023-09-05
stars task solution

yuliabakun:
- [DEMO LINK](https://yulabakun.github.io/layout_stars/)
- [TEST REPORT LINK](https://yulabakun.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -8,6 +8,52 @@
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Stars</h1>
+    <div class="stars__block stars--0">
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+    </div>
+
+    <div class="stars__block stars--1">
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+    </div>
+
+    <div class="stars__block stars--2">
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+    </div>
+
+    <div class="stars__block stars--3">
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+    </div>
+
+    <div class="stars__block stars--4">
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+    </div>
+
+    <div class="stars__block stars--5">
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+    </div>
   </body>
 </html>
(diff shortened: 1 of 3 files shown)


src/index.html
```diff
@@ -8,6 +8,52 @@
...
   </head>
   <body>
-    <h1>Stars</h1>
+    <div class="stars__block stars--0">
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+    </div>
```
33Orange, line 17: Block - `stars`
Element - `stars__star` 
Modifier - `stars--0`

```suggestion
    <div class="stars stars--0">
      <div class="stars__star"></div>
      <div class="stars__star"></div>
      <div class="stars__star"></div>
      <div class="stars__star"></div>
      <div class="stars__star"></div>
    </div>
```

src/style.css
```diff
@@ -0,0 +1,38 @@
...
+.stars--2 > .stars__star:nth-child(-n + 2),
+.stars--3 > .stars__star:nth-child(-n + 3),
+.stars--4 > .stars__star:nth-child(-n + 4),
+.stars--5 > .stars__star:nth-child(-n + 5) {
+  background-image: var(--img-star-active);
+  background-repeat: no-repeat;
+  background-position: center;
+  height: var(--star-size);
+  width: var(--star-size);
+}
```
33Orange, line 38: 1. You already have a `width` and `height` styles for star
2. Let's use `background:` shorthand.

```suggestion
.stars--1 > .stars__star:nth-child(1),
.stars--2 > .stars__star:nth-child(-n + 2),
.stars--3 > .stars__star:nth-child(-n + 3),
.stars--4 > .stars__star:nth-child(-n + 4),
.stars--5 > .stars__star:nth-child(-n + 5) {
  background: var(--img-star-active) no-repeat center;
}
```

src/style.css
```diff
@@ -0,0 +1,38 @@
+body {
+  margin: 0;
+
+  --star-size: 16px;
+  --img-star-inactive: url(images/star.svg);
+  --img-star-active: url(images/star-active.svg);
```
33Orange, line 6: I think it is unnecessary here. 
You use these variables only once

33Orange (changes requested):
Almost done

src/style.css
```diff
@@ -0,0 +1,30 @@
...
+}
+
+.stars__star {
+  margin-right: 4px;
+  background: url(images/star.svg) no-repeat center;
+  height: var(--star-size);
+  width: var(--star-size);
+}
+
+.stars__star:last-child {
```
vasylmiskiv, line 20:
```suggestion
.stars__star:last-child {
```

src/style.css
```diff
@@ -0,0 +1,30 @@
+body {
+  margin: 0;
+
```
vasylmiskiv, line 3: Don't forget remove empty spaces

src/style.css
```diff
@@ -0,0 +1,30 @@
...
+  --star-size: 16px;
```
vasylmiskiv, line 4: The variable is not necessary in this case but it is not a critical

src/style.css
```diff
@@ -0,0 +1,30 @@
...
+}
+
+.stars {
+  display: flex;
+  flex-direction: row;
```
vasylmiskiv, line 9: `flex-direction:row` is default behavior of elements inside flex so don't write this property unless you need to

src/style.css
```diff
@@ -0,0 +1,30 @@
...
+  background: url(images/star.svg) no-repeat center;
+  height: var(--star-size);
+  width: var(--star-size);
+}
+
+.stars__star:last-child {
+  margin-right: 0;
+}
+
+.stars--1 > .stars__star:nth-child(1),
```
vasylmiskiv, line 24: `>` is not necessary because we don't have other elements with different class inside `stars--1` for definition to which certain class should apply our properties. So you can remove it

vasylmiskiv (approved):

PR #3562, opened 2023-09-05
add task solution

mpohorenyi:
- [DEMO LINK](https://mpohorenyi.github.io/layout_stars/)
- [TEST REPORT LINK](https://mpohorenyi.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -8,6 +8,54 @@
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Stars</h1>
+    <div class="stars--block">
+      <div class="stars stars--0">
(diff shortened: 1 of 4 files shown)


src/index.html
```diff
@@ -8,6 +8,54 @@
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Stars</h1>
+    <div class="stars--block">
```
polosanya, line 11:
```suggestion
    <div class="stars-container">
```
It is not a modifier of stars block, so let's change the name

polosanya (approved):
Well done!

PR #3566, opened 2023-09-05
add stars

vlevik:
- [DEMO LINK](https://vlevik.github.io/layout_stars/)
- [TEST REPORT LINK](https://vlevik.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,26 @@
+body {
+  margin: 0;
+}
+
+.stars {
+  box-sizing: border-box;
+  display: flex;
+  justify-content: flex-start;
+  align-items: center;
+}
+
+.stars__star {
+  display: flex;
+  height: 16px;
+  width: 16px;
+  margin-right: 4px;
+  background: no-repeat center center;
+}
+
+.stars__star--star {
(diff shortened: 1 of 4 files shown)


StasSokolov (changes requested):
Good job, left to fix only few comments!

src/style.css
```diff
@@ -0,0 +1,27 @@
+body {
+  margin: 0;
+}
+
+.stars {
+  box-sizing: border-box;
+  display: flex;
+  justify-content: flex-start;
```
polosanya, line 8: That's a default value

src/style.css
```diff
@@ -0,0 +1,27 @@
...
+  align-items: center;
+}
+
+.stars__star {
+  display: flex;
+  height: 16px;
+  width: 16px;
+  margin-right: 4px;
+  background: no-repeat center center;
+  background-image: url(./images/star.svg);
```
polosanya, line 18: You can add it to `background` too, but it's not necessary 
https://developer.mozilla.org/en-US/docs/Web/CSS/background#syntax

polosanya (approved):
Great!

PR #3583, opened 2023-09-05
beta_version_of_BEM

Stan1slav123:
- [DEMO LINK](https://Stan1slav123.github.io/layout_stars/)
- [TEST REPORT LINK](https://Stan1slav123.github.io/layout_stars/report/html_report/)

Help me please with BEM i don't understand how to change default star to active only with star--N modifier

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -8,6 +8,47 @@
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Stars</h1>
+    <div class="stars stars--0">
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+    </div>
+    <div class="stars stars--1">
+      <div class="stars__star stars__star--active"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+    </div>
+    <div class="stars stars--2">
+      <div class="stars__star stars__star--active"></div>
+      <div class="stars__star stars__star--active"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+    </div>
+    <div class="stars stars--3">
+      <div class="stars__star stars__star--active"></div>
+      <div class="stars__star stars__star--active"></div>
+      <div class="stars__star stars__star--active"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+    </div>
+    <div class="stars stars--4">
+      <div class="stars__star stars__star--active"></div>
+      <div class="stars__star stars__star--active"></div>
+      <div class="stars__star stars__star--active"></div>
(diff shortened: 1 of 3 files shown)


src/index.html
```diff
@@ -8,6 +8,47 @@
...
-    <h1>Stars</h1>
+    <div class="stars stars--0">
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+    </div>
+    <div class="stars stars--1">
+      <div class="stars__star stars__star--active"></div>
```
maxim2310, line 19: remove it everywhere 
```suggestion
      <div class="stars__star "></div>
```

src/style.css
```diff
@@ -0,0 +1,25 @@
...
+  margin: 0 4px 0 0;
+  background: url(images/star.svg) no-repeat;
+  background-position: center;
+}
+
+.stars__star:last-child {
+  margin: 0;
+}
+
+.stars__star--active {
```
maxim2310, line 23:
```suggestion
.stars--1 > :nth-child(n+1):nth-child(-n+1),
(and other the same way) {
```

maxim2310 (changes requested):
Good work, but lets make some improvement

src/style.css
Stan1slav123: Why doesn't it work without it ">"?

Stan1slav123: But everything works without ":nth-child(n+1)"

src/style.css
```diff
@@ -0,0 +1,39 @@
...
+  background: url(images/star-active.svg) no-repeat center;
+}
+
+.stars--4 :nth-child(-n + 4) {
+  background: url(images/star-active.svg) no-repeat center;
+}
+
+.stars--5 :nth-child(-n + 5) {
+  background: url(images/star-active.svg) no-repeat center;
+}
```
Viktor-Kost, line 39: 1. we can use as a group
2. no needs to repeat styles that we've already applied
```suggestion
.stars--1 :nth-child(-n + 1),
.stars--2 :nth-child(-n + 2),
.stars--3 :nth-child(-n + 3),
.stars--4 :nth-child(-n + 4),
.stars--5 :nth-child(-n + 5) {
  background: url(images/star-active.svg);
}
```

src/style.css
```diff
@@ -0,0 +1,39 @@
...
+  height: 16px;
+  width: 16px;
+  margin: 0 4px 0 0;
+  background: url(images/star.svg) no-repeat;
+  background-position: center;
+}
+
+.stars__star:last-child {
+  margin: 0;
+}
```
Viktor-Kost, line 19: 1. we can simplify setting margin
2. use background `shorthand`

```suggestion
.stars__star {
  height: 16px;
  width: 16px;
  background: url(images/star.svg) no-repeat center;
}

.stars__star:not(:last-child) {
  margin-right: 4px;
}
```

Viktor-Kost (changes requested):

Esceype (approved):
Good job 👍 !!!

PR #3587, opened 2023-09-07
add task solution

OlgaFloria:
https://OlgaFloria.github.io/layout_stars/
https://OlgaFloria.github.io/layout_stars/report/html_report

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -8,6 +8,50 @@
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Stars</h1>
+    <main>
+      <div class="stars stars--0">
+        <div class="stars__star"></div>
+        <div class="stars__star"></div>
+        <div class="stars__star"></div>
+        <div class="stars__star"></div>
+        <div class="stars__star"></div>
+      </div>
+      <div class="stars stars--1">
+        <div class="stars__active"></div>
+        <div class="stars__star"></div>
+        <div class="stars__star"></div>
+        <div class="stars__star"></div>
+        <div class="stars__star"></div>
+      </div>
+      <div class="stars stars--2">
+        <div class="stars__active"></div>
+        <div class="stars__active"></div>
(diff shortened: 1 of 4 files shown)


src/index.html
```diff
@@ -8,6 +8,50 @@
...
+    <main>
+      <div class="stars stars--0">
+        <div class="stars__star"></div>
+        <div class="stars__star"></div>
+        <div class="stars__star"></div>
+        <div class="stars__star"></div>
+        <div class="stars__star"></div>
+      </div>
+      <div class="stars stars--1">
+        <div class="stars__active"></div>
```
etojeDenys, line 20: image
We should have the ability to change the number of active stars just by changing the modifier

etojeDenys (changes requested):

src/index.html
volodymyr-soltys97: Still not fixed from the previous review, you need to remove extra class

src/index.html
```diff
@@ -8,6 +8,50 @@
...
+        <div class="stars__star"></div>
+        <div class="stars__star"></div>
+        <div class="stars__star"></div>
+      </div>
+      <div class="stars stars--2">
+        <div class="stars__active"></div>
+        <div class="stars__active"></div>
+        <div class="stars__star"></div>
+        <div class="stars__star"></div>
+        <div class="stars__star"></div>
```
volodymyr-soltys97, line 31: Need to fix everywhere
```suggestion
        <div class="stars__star"></div>
        <div class="stars__star"></div>
        <div class="stars__star"></div>
        <div class="stars__star"></div>
        <div class="stars__star"></div>
```

volodymyr-soltys97 (changes requested):

etojeDenys (changes requested):
.

volodymyr-soltys97 (approved):
Well done 🔥

PR #3588, opened 2023-09-07
Create stars section

buffi4ka:
- [DEMO LINK](https://buffi4ka.github.io/layout_stars/)
- [TEST REPORT LINK](https://buffi4ka.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,35 @@
+* {
+  box-sizing: border-box;
+}
+
+body,
+ul,
+li {
+  margin: 0;
+  padding: 0;
+}
+
+.stars {
+  display: flex;
+  list-style: none;
+}
+
+.stars > :not(:first-child) {
+  margin-left: 4px;
+}
+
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,35 @@
+* {
+  box-sizing: border-box;
+}
+
+body,
+ul,
+li {
+  margin: 0;
+  padding: 0;
+}
```
superpooperxxx, line 10: `li` is redundant here

src/style.css
```diff
@@ -0,0 +1,35 @@
...
+
+.stars {
+  display: flex;
+  list-style: none;
+}
```
superpooperxxx, line 15: I would suggest moving `list-style: none` to the top and reseting it for all `<ul>`

superpooperxxx (approved):
Awesome job🔥🔥🔥

PR #3590, opened 2023-09-07
add task solution but failed the tests) Help

mr-prophet:
[DEMO LINK](https://mr-prophet.github.io/layout_stars/)
- [TEST REPORT LINK](https://mr-prophet.github.io/layout_stars/report/html_report/)
what is the problem?)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -8,6 +8,49 @@
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Stars</h1>
+    <div class="stars--block">
+      <ul class="stars stars--0">
+        <li class="stars__star"></li>
+        <li class="stars__star"></li>
+        <li class="stars__star"></li>
+        <li class="stars__star"></li>
+        <li class="stars__star"></li>
+      </ul>
+      <ul class="stars stars--1">
+        <li class="stars__star stars__star_active"></li>
+        <li class="stars__star"></li>
+        <li class="stars__star"></li>
+        <li class="stars__star"></li>
+        <li class="stars__star"></li>
+      </ul>
+      <ul class="stars stars--2">
+        <li class="stars__star stars__star_active"></li>
+        <li class="stars__star stars__star_active"></li>
+        <li class="stars__star"></li>
+        <li class="stars__star"></li>
+        <li class="stars__star"></li>
+      </ul>
+      <ul class="stars stars--3">
+        <li class="stars__star stars__star_active"></li>
+        <li class="stars__star stars__star_active"></li>
+        <li class="stars__star stars__star_active"></li>
+        <li class="stars__star"></li>
+        <li class="stars__star"></li>
+      </ul>
+      <ul class="stars stars--4">
+        <li class="stars__star stars__star_active"></li>
+        <li class="stars__star stars__star_active"></li>
+        <li class="stars__star stars__star_active"></li>
+        <li class="stars__star stars__star_active"></li>
+        <li class="stars__star"></li>
+      </ul>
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,63 @@
...
+}
+
+/* Block stars */
+
+.stars--block {
+  display: flex;
+  flex-direction: column;
+  justify-content: space-between;
+  height: 96px;
+  width: 96px;
```
Esceype, line 23: Checklist 3

src/style.css
```diff
@@ -0,0 +1,63 @@
...
+}
+
+.stars {
+  display: flex;
+  justify-content: space-between;
```
Esceype, line 28: You don`t need it after you will remove the fixed container size

src/style.css
```diff
@@ -0,0 +1,63 @@
...
+}
+
+.stars__star {
```
Esceype, line 31: Should have margin-right: 4px but not for the last one in row

src/style.css
```diff
@@ -0,0 +1,63 @@
...
+  background-image: url(images/star-active.svg);
+}
+
+.stars--4 .stars__star:nth-child(-n+4) {
+  background-image: url(images/star-active.svg);
+}
+
+.stars--5 .stars__star:nth-child(-n+5) {
+  background-image: url(images/star-active.svg);
+}
```
Esceype, line 63: Should be from 0-5, so enable commented code and make it like in checklist point number 2

src/index.html
```diff
@@ -8,6 +8,49 @@
...
+    <div class="stars--block">
+      <ul class="stars stars--0">
+        <li class="stars__star"></li>
+        <li class="stars__star"></li>
+        <li class="stars__star"></li>
+        <li class="stars__star"></li>
+        <li class="stars__star"></li>
+      </ul>
+      <ul class="stars stars--1">
+        <li class="stars__star stars__star_active"></li>
```
Esceype, line 20: You should have only one class here and in the same cases where you use additional class for active star. 
The number of active stars must be specified in the modifier, which is defined in the container for each row of stars

src/style.css
```diff
@@ -0,0 +1,63 @@
+/* Reset and base styles  */
+
+body {
+  margin: 0;
+  padding: 0;
+}
+
+ul,
+li {
+  display: flex;
```
Esceype, line 10: You don't need to specify additional styles props for tags where you use classes directly, class selectors are more specific, so this definition doesn't make sense

src/style.css
```diff
@@ -0,0 +1,63 @@
...
+  margin: 0;
+  list-style: none;
+}
+
+/* Block stars */
+
+.stars--block {
+  display: flex;
+  flex-direction: column;
+  justify-content: space-between;
```
Esceype, line 21: Remove it cause you don`t need space between rows

src/style.css
```diff
@@ -0,0 +1,63 @@
+/* Reset and base styles  */
+
+body {
+  margin: 0;
+  padding: 0;
```
Esceype, line 5: body haven`t paddings

Esceype (changes requested):
At all good job, leave several comments to improve code style and use it to try pass tests

mykhalenych (changes requested):
I don't see any changed from previous review, did you push your code?

mr-prophet:
yes

mr-prophet:
sorry, made a mistake while saving

mykhalenych (changes requested):
almost done

mykhalenych (approved):

PR #3598, opened 2023-09-11
add task solution

BodyaRespect:
- [DEMO LINK](https://BodyaRespect.github.io/layout_stars/)
- [TEST REPORT LINK](https://BodyaRespect.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,24 @@
+body {
+  margin: 0;
+}
+
+.stars {
+  display: flex;
+}
+
+.stars_star {
+  width: 16px;
+  height: 16px;
+  margin-right: 4px;
+  background-image: url(./images/star.svg);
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,24 @@
...
+  display: flex;
+}
+
+.stars_star {
+  width: 16px;
+  height: 16px;
+  margin-right: 4px;
+  background-image: url(./images/star.svg);
+  background-position: center;
+  background-repeat: no-repeat;
```
loralevitska, line 15: Combine with [`background` property](https://developer.mozilla.org/en-US/docs/Web/CSS/background)

loralevitska (approved):
Almost perfect) Check my comments for improvement

PR #3605, opened 2023-09-14
add task solution

indigo04:
https://indigo04.github.io/layout_stars/
https://indigo04.github.io/layout_stars/report/html_report/

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,34 @@
+body {
+  margin: 0;
+}
+
+.stars--0,
+.stars--1,
+.stars--2,
+.stars--3,
+.stars--4,
+.stars--5 {
+  display: flex;
+  align-items: center;
+  justify-content: center;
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,34 @@
...
+}
+
+.stars_star {
+  background-image: url(images/star.svg);
+  background-size: cover;
+  display: flex;
+  width: 12px;
+  height: 12px;
+  margin: 2px 6px;
+  margin-left: 2px;
```
nazarmatsevych, line 23: According to the mockup, you should only add margin-right to all stars except the last stars in a row
Screenshot 2023-09-14 at 12 31 12

nazarmatsevych (changes requested):

maxim2310 (approved):
Good work!!!

PR #3610, opened 2023-09-18
add task solution

ikorostylev:
[DEMO LINK](https://ikorostylev.github.io/layout_stars/)
[TEST REPORT LINK](https://ikorostylev.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -8,54 +8,54 @@
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <div class="container">
-      <div class="stars stars--0">
-        <div class="stars__star"></div>
-        <div class="stars__star"></div>
-        <div class="stars__star"></div>
-        <div class="stars__star"></div>
-        <div class="stars__star"></div>
-      </div>
 
-      <div class="stars stars--1">
-        <div class="stars__star__active"></div>
-        <div class="stars__star"></div>
-        <div class="stars__star"></div>
-        <div class="stars__star"></div>
-        <div class="stars__star"></div>
-      </div>
-
-      <div class="stars stars--2">
-        <div class="stars__star__active"></div>
-        <div class="stars__star__active"></div>
-        <div class="stars__star"></div>
-        <div class="stars__star"></div>
-        <div class="stars__star"></div>
-      </div>
-
-      <div class="stars stars--3">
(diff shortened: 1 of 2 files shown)


src/style.css
```diff
@@ -0,0 +1,32 @@
...
+
+}
+
+.stars__star {
+  background-position: center;
+  background-image: url("../src/images/star.svg");
+  width: 16px;
+  height: 16px;
+  background-repeat: no-repeat;
+  margin-left: 2px;
```
mykhalenych, line 22: add margin-right: 4px for all stars instead the last

src/index.html
```diff
@@ -8,6 +8,54 @@
...
+      <div class="stars stars--0">
+        <div class="stars__star"></div>
+        <div class="stars__star"></div>
+        <div class="stars__star"></div>
+        <div class="stars__star"></div>
+        <div class="stars__star"></div>
+      </div>
+
+      <div class="stars stars--1">
+        <div class="stars__star__active"></div>
```
mykhalenych, line 21: Screenshot 2023-09-18 at 17 20 31

delete __active and use modifier for adding active star

mykhalenych (changes requested):

src/style.css
```diff
@@ -0,0 +1,34 @@
...
+  padding: 0;
+
+}
+
+.stars__star {
+  background: url(./images/star.svg) no-repeat center;
+  height: 16px;
+  width: 16px;
+  margin: 0 2px;
+}
```
VitaliyBondarenko1982, line 18:
```suggestion
}

.stars__star:not(:last-child) {
  margin-right: 4px;
}
```
better to make bargin from one side.
and with such selector will be settled margin for all elements except last.

src/style.css
```diff
@@ -0,0 +1,34 @@
...
+  background: url(./images/star-active.svg) no-repeat center;
+}
+
+.stars__star:first-child {
+  margin-left: 0;
+}
+
+.stars__star:last-child {
+  margin-right: 0;
+}
```
VitaliyBondarenko1982, line 34:
```suggestion
```
if to set margin to one side except last element - wil be no need this styles.

VitaliyBondarenko1982 (approved):
Well done.
Check comments for next tasks.

PR #3617, opened 2023-09-21
Stars task Liubyn_V

VolodymyrLiubyn:


src/index.html
```diff
@@ -8,6 +8,49 @@
...
+    <div class="container">
+      <div class="stars stars--0">
+        <div class="stars__star"></div>
+        <div class="stars__star"></div>
+        <div class="stars__star"></div>
+        <div class="stars__star"></div>
+        <div class="stars__star"></div>
+      </div>
+      <div class="stars stars--1">
+        <div class="stars__star is_active"></div>
```
etojeDenys, line 20: you should implement it without the additional class

etojeDenys (changes requested):

src/index.html
VolodymyrLiubyn: Ok. I removed an additional class

Andriy-Shymkiv: @etojeDenys mean `is-active`, you should remove it and set bgc directly from css

src/style.css
```diff
@@ -0,0 +1,40 @@
+body {
+  margin: 0;
+  box-sizing: border-box;
+}
+
+.container {
+  box-sizing: border-box;
+  display: flex;
+  width: 100px;
+  height: 96px;
```
Andriy-Shymkiv, line 10: do not hardcode sizes here
```suggestion
```

src/style.css
```diff
@@ -0,0 +1,40 @@
...
+  align-items: center;
+  margin: auto;
+
+}
+
+.stars {
+  display: flex;
+  flex-direction: row;
+  justify-content: space-between;
+  width: 96px;
```
Andriy-Shymkiv, line 22: same
```suggestion
```

src/style.css
```diff
@@ -0,0 +1,40 @@
...
+  flex-direction: column;
+  justify-content: center;
+  align-items: center;
+  margin: auto;
+
+}
+
+.stars {
+  display: flex;
+  flex-direction: row;
```
Andriy-Shymkiv, line 20: it is row by default
```suggestion
```

src/style.css
```diff
@@ -0,0 +1,40 @@
...
+  justify-content: space-between;
+  width: 96px;
+  padding-right: 4px;
```
Andriy-Shymkiv, line 23: set `margin-right`, not `padding`

src/style.css
```diff
@@ -0,0 +1,40 @@
...
+  background-position: center;
+}
+
+div.is_active {
+  width: 16px;
+  height: 16px;
+  background-image: url(./images/star-active.svg);
+  background-repeat: no-repeat;
+  background-position: center;
+}
```
Andriy-Shymkiv, line 40: 1. do not use selector by tag,
2. you should use `nth-child` pseudo class here to set active stars

Andriy-Shymkiv (changes requested):
please fix previous comment about class

src/style.css
VolodymyrLiubyn: Fixed inaccuracies in the code, used pseudo-classes nth-child for active stars

sTorba24 (changes requested):
Well done!

next time add DEMO link to description here

image

sTorba24 (approved):

PR #3630, opened 2023-09-27
Develop

Vasyl1991:
https://Vasyl1991.github.io/layout_stars/
https://Vasyl1991.github.io/layout_stars/report/html_report/

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -2,12 +2,60 @@
 <html lang="en">
   <head>
     <meta charset="UTF-8">
-    <meta name="viewport"
-          content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0">
+    <meta
+          name="viewport"
+          content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0"
+    >
     <title>Stars</title>
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Stars</h1>
+    <div class="stars stars--0">
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+    </div>
+
+    <div class="stars stars--1">
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+    </div>
+
+    <div class="stars stars--2">
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,37 @@
...
+  background-image: url(./images/star-active.svg);
+}
+
+.stars--4 > div:nth-child(-n + 4) {
+  background-image: url(./images/star-active.svg);
+}
+
+.stars--5 > * {
+  background-image: url(./images/star-active.svg);
+}
```
etojeDenys, line 37: you can join it using the comma

etojeDenys (changes requested):

src/index.html
```diff
@@ -2,12 +2,60 @@
 <html lang="en">
   <head>
     <meta charset="UTF-8">
-    <meta name="viewport"
-          content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0">
+    <meta
+          name="viewport"
+          content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0"
+    >
```
DarkMistyRoom, line 8:
```suggestion
    <meta
      name="viewport"
      content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0"
    >
```

DarkMistyRoom: 2 spaces

src/style.css
```diff
@@ -0,0 +1,25 @@
+html {
+  margin: 32px 22px;
+}
```
DarkMistyRoom, line 3: don't set margin to `html`. It may be a reason why the tests failed

src/style.css
```diff
@@ -0,0 +1,25 @@
...
+  height: 16px;
+  width: 16px;
+  background-image: url(./images/star.svg);
+  background-position: center;
+  background-size: auto;
+  background-repeat: no-repeat;
+  margin-right: 4px;
+}
+
+.stars--1 > div:nth-child(1),
```
DarkMistyRoom, line 19:
```suggestion
.stars--1 > stars__star:nth-child(1),
```

DarkMistyRoom: use class instead of tag

DarkMistyRoom: fix it for other selectors

src/style.css
```diff
@@ -0,0 +1,25 @@
...
+.stars--2 > div:nth-child(-n + 2),
+.stars--3 > div:nth-child(-n + 3),
+.stars--4 > div:nth-child(-n + 4),
+.stars--5 > * {
```
DarkMistyRoom, line 23: don't use *. What if we need to add some elements inside the block(not stars__star)

DarkMistyRoom (changes requested):
Good job! But pay attention to the suggestions

VitaliyBodnar (approved):

PR #3648, opened 2023-10-04
add task solution

valost:
https://valost.github.io/layout_stars/
https://valost.github.io/layout_stars/report/html_report/

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,30 @@
+body {
+  margin: 0;
+}
+
+.stars {
+  display: flex;
+  justify-content: center;
+  align-items: center;
+}
+
+.stars__star {
+  width: 16px;
+  height: 16px;
+  background-image: url(images/star.svg);
+  background-repeat: no-repeat;
+  background-position: center;
+}
+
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,30 @@
...
+  justify-content: center;
+  align-items: center;
+}
+
+.stars__star {
+  width: 16px;
+  height: 16px;
+  background-image: url(images/star.svg);
+  background-repeat: no-repeat;
+  background-position: center;
```
sTorba24, line 16: background shorthand property can be used

sTorba24 (changes requested):
Well done! Fix tests please

src/style.css
valost: @sTorba24 I'm not sure I understand how to fix the tests? Is there a problem in my code or sth else?

valost:
Hi! I tried to fix the tests with the mentor's help but nothing worked. The mentor told me just to re-request review and notify you that the tests don't work.

Viktor-Kost (approved):

PR #3657, opened 2023-10-10
Develop

katenevm:
I'm so sory, I know that test was failed, but I have no idea why, because it seems like everything looks like in the reference and I dont understand from the npm test mistakes what should I correct, so can you please help me

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -8,6 +8,116 @@
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Stars</h1>
+    <div class="box">
+      <table>
+        <tr>
+          <td>
+            <div class="star"></div>
+          </td>
+          <td>
+            <div class="star"></div>
+          </td>
+          <td>
+            <div class="star"></div>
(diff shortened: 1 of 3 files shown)


src/index.html
```diff
@@ -8,6 +8,54 @@
...
   </head>
   <body>
-    <h1>Stars</h1>
+    <div class="line stars--0">
+      <div class="star"></div>
+      <div class="star"></div>
+      <div class="star"></div>
+      <div class="star"></div>
+      <div class="star"></div>
+    </div>
```
etojeDenys, line 17:
```suggestion
    <div class="stars stars--0">
      <div class="stars__star"></div>
      <div class="stars__star"></div>
      <div class="stars__star"></div>
      <div class="stars__star"></div>
      <div class="stars__star"></div>
    </div>
```
let's name classes using the BEM

etojeDenys (changes requested):

maxim2310 (approved):
Good work!!!

PR #3659, opened 2023-10-10
stars

DanyloSemenov:


DanyloSemenov:
- [TEST REPORT LINK](https://DanyloSemenov.github.io/layout_stars/report/html_report/)
- [DEMO LINK](https://DanyloSemenov.github.io/layout_stars/)

src/style.css
```diff
@@ -0,0 +1,34 @@
...
+.stars--2 .stars__star:nth-child(2),
+.stars--3 .stars__star:nth-child(1),
+.stars--3 .stars__star:nth-child(2),
+.stars--3 .stars__star:nth-child(3),
+.stars--4 .stars__star:nth-child(1),
+.stars--4 .stars__star:nth-child(2),
+.stars--4 .stars__star:nth-child(3),
+.stars--4 .stars__star:nth-child(4) {
+  background-image: url("images/star-active.svg");
+}
```
JoyCoffeeAddict, line 34: You have to use `nth-child(n)` syntax, read about it in the [documentation](https://developer.mozilla.org/en-US/docs/Web/CSS/:nth-child)

JoyCoffeeAddict (changes requested):

src/style.css
```diff
@@ -0,0 +1,40 @@
...
+
+.stars--4 .stars__star:nth-child(-n+4) {
+  background-image: url("images/star-active.svg");
+}
+
+.stars--5 .stars__star:nth-child(-n+5) {
+  background-image: url("images/star-active.svg");
+}
```
mvjl000, line 40: Values area repeated. Let's create a combined selector (separated with comma `,`) and it's gonna be perfect!

mvjl000 (changes requested):

src/style.css
```diff
@@ -0,0 +1,28 @@
...
+
+.stars {
+  display: flex;
+
+}
+
+.stars__star {
+  height: 16px;
+  width: 16px;
```
mvjl000, line 12: Those values could be in a variable

src/style.css
```diff
@@ -0,0 +1,28 @@
...
+  background-image: url("images/star.svg");
+  background-repeat: no-repeat;
+  background-position: center;
```
mvjl000, line 15: According to the checklist - background properties should be combined using a shorthand

mvjl000 (changes requested):
Your solution is good, there are just two small things to make it ideal. So sorry I didn't catch it before :/

mvjl000 (approved):

PR #3663, opened 2023-10-10
add task solution

Sibunas:


src/style.css
```diff
@@ -0,0 +1,34 @@
...
+.stars--2 .stars__star:nth-child(2),
+.stars--3 .stars__star:nth-child(1),
+.stars--3 .stars__star:nth-child(2),
+.stars--3 .stars__star:nth-child(3),
+.stars--4 .stars__star:nth-child(1),
+.stars--4 .stars__star:nth-child(2),
+.stars--4 .stars__star:nth-child(3),
+.stars--4 .stars__star:nth-child(4) {
+  background-image: url("images/star-active.svg");
+}
```
JoyCoffeeAddict, line 34: You have to use `nth-child(n)` syntax, read about it in the [documentation](https://developer.mozilla.org/en-US/docs/Web/CSS/:nth-child)

JoyCoffeeAddict (changes requested):

mvjl000 (approved):
Well done!

PR #3664, opened 2023-10-10
task solution

evoora:
https://evoora.github.io/layout_stars/
https://evoora.github.io/layout_stars/report/html_report/


nazarmatsevych (approved):
Great work!

PR #3667, opened 2023-10-10
solution

aureliagorzna:


src/index.html
```diff
@@ -1,13 +1,55 @@
...
   <body>
-    <h1>Stars</h1>
+    <main class="main">
+      <div class="stars__star stars--0">
+        <div class="star--card inactive"></div>
+        <div class="star--card inactive"></div>
+        <div class="star--card inactive"></div>
+        <div class="star--card inactive"></div>
+        <div class="star--card inactive"></div>
+      </div>
```
mvjl000, line 17: 1. I would change the name of this div's class. I think simply `stars` would be better and more clear since it is a row of stars. Not a single star.
2. Divs inside should have only one class. According to BEM it should be named stars__star

mvjl000 (changes requested):
Overall, you solution looks good but it should be done with different syntax. Please refactor your styles so you use nth pseudo classes

src/style.css
```diff
@@ -0,0 +1,39 @@
...
+  margin: 0;
+}
+
+* {
+  margin: 0;
+  padding: 0;
+  box-sizing: border-box;
+}
+
+main {
```
mvjl000, line 11: Actually you don't need any wrapper for this. You can leave the star's rows only

src/style.css
```diff
@@ -0,0 +1,39 @@
...
+
+.active {
+  background-image: url(../src/images/star-active.svg);
+  background-repeat: no-repeat;
+}
+
+.inactive {
+  background-image: url(../src/images/star.svg);
+  background-repeat: no-repeat;
+}
```
mvjl000, line 39: As I said before, this must be done using :nth-child pseudo class. Check [the docs](https://developer.mozilla.org/en-US/docs/Web/CSS/:nth-child) please (remember to use functional notation)

mvjl000 (changes requested):

mbruhler (approved):
Approved :)

PR #3668, opened 2023-10-10
add task solution

mosowiecka:


mosowiecka:
- [DEMO LINK](https://mosowiecka.github.io/layout_stars/)
- [TEST REPORT LINK](https://mosowiecka.github.io/layout_stars/report/html_report/)

src/style.css
```diff
@@ -0,0 +1,39 @@
...
+  background-image: url("images/star-active.svg");
+}
+
+.stars--4 .stars__star:nth-of-type(-n+4) {
+  background-image: url("images/star-active.svg");
+}
+
+.stars--5 .stars__star:nth-of-type(-n+5) {
+  background-image: url("images/star-active.svg");
+}
```
mvjl000, line 39: Please combine those selectors with comma `,`
DRY - don't repeat yourself

src/style.css
```diff
@@ -0,0 +1,39 @@
...
+  display: flex;
+}
+
+.stars__star {
+  background-image: url("images/star.svg");
+  height: 16px;
+  width: 16px;
+  background-repeat: space;
+  background-position: center;
+  margin-right: 4px;
```
mvjl000, line 15: Note that you don't need margin for the last star ;)

src/style.css
```diff
@@ -0,0 +1,39 @@
...
+
+.stars {
+  display: flex;
+}
+
+.stars__star {
+  background-image: url("images/star.svg");
+  height: 16px;
+  width: 16px;
+  background-repeat: space;
```
mvjl000, line 13: `space` works here but `no-repeat` is the exact value we want to use in this case.
Check [the docs](https://developer.mozilla.org/en-US/docs/Web/CSS/background-repeat)

src/style.css
```diff
@@ -0,0 +1,39 @@
...
+}
+
+.stars {
+  display: flex;
+}
+
+.stars__star {
+  background-image: url("images/star.svg");
+  height: 16px;
+  width: 16px;
```
mvjl000, line 12: This dimensions could be in a variable

mvjl000 (changes requested):
Please apply my suggestions and your solution will be accepted. Good job overall! 🚀

src/style.css
```diff
@@ -0,0 +1,53 @@
...
+  height: var(--height);
+  width: var(--width);
+  background-repeat: no-repeat;
+  background-position: center;
+  margin-right: 4px;
+}
+
+/* .stars__star:not(:last-child) {
+  margin: 0;
+} */
```
mvjl000, line 25: If this causes your tests to fail, try to use only :last-child pseudo class

mvjl000 (changes requested):
Please remove comments so it looks cleaner 🙏

src/style.css
mosowiecka: tests end successfully

mvjl000 (approved):
Good!

PR #3669, opened 2023-10-10
add task solution

WiktoriaBielec:
- [DEMO LINK](https://WiktoriaBielec.github.io/layout_stars/)
- [TEST REPORT LINK](https://WiktoriaBielec.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -8,6 +8,49 @@
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Stars</h1>
+    <div class="container">
+      <div class="stars stars--0">
+        <div class="stars__star"></div>
+        <div class="stars__star"></div>
+        <div class="stars__star"></div>
+        <div class="stars__star"></div>
+        <div class="stars__star"></div>
+      </div>
+      <div class="stars stars--1">
+        <div class="stars__star stars__star-active"></div>
+        <div class="stars__star"></div>
+        <div class="stars__star"></div>
+        <div class="stars__star"></div>
+        <div class="stars__star"></div>
+      </div>
+      <div class="stars stars--2">
+        <div class="stars__star stars__star-active"></div>
+        <div class="stars__star stars__star-active"></div>
+        <div class="stars__star"></div>
+        <div class="stars__star"></div>
+        <div class="stars__star"></div>
+      </div>
+      <div class="stars stars--3">
+        <div class="stars__star stars__star-active"></div>
+        <div class="stars__star stars__star-active"></div>
+        <div class="stars__star stars__star-active"></div>
+        <div class="stars__star"></div>
+        <div class="stars__star"></div>
+      </div>
+      <div class="stars stars--4">
+        <div class="stars__star stars__star-active"></div>
+        <div class="stars__star stars__star-active"></div>
+        <div class="stars__star stars__star-active"></div>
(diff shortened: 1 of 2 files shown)


src/style.css
```diff
@@ -0,0 +1,33 @@
...
+}
+
+.stars__star {
+  background-image: url(\images\star.svg);
+  height: 16px;
+  width: 16px;
+  background-position: center;
+  object-fit: cover;
+  box-sizing: border-box;
+  border: solid white 2px;
```
mvjl000, line 22: You don't need border here. If you see multiple stars - add `background-repeat: no-repeat`

src/style.css
```diff
@@ -0,0 +1,33 @@
...
+
+.stars {
+  display: flex;
+  justify-content: space-between;
+}
+
+.stars__star {
+  background-image: url(\images\star.svg);
+  height: 16px;
+  width: 16px;
```
mvjl000, line 18: Please create a variables for repeated values

src/index.html
```diff
@@ -8,6 +8,49 @@
...
+        <div class="stars__star stars__star-active"></div>
+        <div class="stars__star"></div>
+        <div class="stars__star"></div>
+      </div>
+      <div class="stars stars--4">
+        <div class="stars__star stars__star-active"></div>
+        <div class="stars__star stars__star-active"></div>
+        <div class="stars__star stars__star-active"></div>
+        <div class="stars__star stars__star-active"></div>
+        <div class="stars__star"></div>
```
mvjl000, line 45: This is not the approach we want in this case.
Star should have only one class and you should change background-image basing on parent's class (in this case `stars--4`. You can use `:nth` pseudo class to make it work 😸)

Btw modifiers should have two dashes ;)

src/style.css
```diff
@@ -0,0 +1,33 @@
...
+
+.stars__star-active {
+  background-image: url(\images\star-active.svg);
+  height: 16px;
+  width: 16px;
+  background-position: center;
+  object-fit: cover;
+  box-sizing: border-box;
+  border: solid white 2px;
+}
```
mvjl000, line 33: Not like this :/
I left some instructions in previous comment

mvjl000 (changes requested):

src/style.css
```diff
@@ -0,0 +1,37 @@
...
+
+.container {
+  height: var(--container-size);
+  width: var(--container-size);
+}
+
+.stars {
+  display: flex;
+  justify-content: space-between;
+}
```
JoyCoffeeAddict, line 18: You should position your stars by using margin between each star, not by using `justify-content: space-between;` This is causing your tests to fail.

JoyCoffeeAddict (changes requested):

src/style.css
```diff
@@ -0,0 +1,44 @@
...
+  width: var(--container-size);
+}
+
+.stars {
+  display: flex;
+}
+
+.stars:last-child {
+  margin-right: 0;
+}
```
JoyCoffeeAddict, line 21: `:last-child` is a pseudo-class that chooses the last element between the siblings, not from the parent's perspective. Also, you set `.stars__star:last-child ` later correctly.

JoyCoffeeAddict (approved):
Other than one unused part of the code, your solution is correct, and there is some issue with the tests

PR #3672, opened 2023-10-11
stars task solution

adrian200002:
- [DEMO LINK](https://<adrian200002>.github.io/layout_stars/)
- [TEST REPORT LINK](https://<adrian200002>.github.io/layout_stars/report/html_report/)

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -1,7 +1,7 @@
 # Stars block
 Replace `<your_account>` with your Github username and copy the links to Pull Request description:
-- [DEMO LINK](https://<your_account>.github.io/layout_stars/)
-- [TEST REPORT LINK](https://<your_account>.github.io/layout_stars/report/html_report/)
+- [DEMO LINK](https://<adrian200002>.github.io/layout_stars/)
+- [TEST REPORT LINK](https://<adrian200002>.github.io/layout_stars/report/html_report/)
 
 > Follow [this instructions](https://mate-academy.github.io/layout_task-guideline)
 ___
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,47 @@
+body {
+  margin: 0;
+}
+
+.stars {
+  display: flex;
+}
+
+.stars__star {
+  background-image: url("images/star-active.svg");
+  background-repeat: no-repeat;
+  background-position: center;
+  width: 16px;
+  height: 16px;
+  margin-right: 4px;
+}
+
+.stars--0 .stars__star {
+  background-image: url("images/star.svg");
+  width: 16px;
+  height: 16px;
+}
+
+.stars--1 div:not(:first-child) {
+  background-image: url("images/star.svg");
+  width: 16px;
+  height: 16px;
+}
+
+.stars--3 div:nth-child(4),
+.stars--3 div:nth-child(5) {
+  background-image: url("images/star.svg");
+  width: 16px;
+  height: 16px;
+}
+
+.stars--4 div:nth-child(5) {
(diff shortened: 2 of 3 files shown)


readme.md
```diff
@@ -1,7 +1,7 @@
 # Stars block
 Replace `<your_account>` with your Github username and copy the links to Pull Request description:
-- [DEMO LINK](https://<your_account>.github.io/layout_stars/)
-- [TEST REPORT LINK](https://<your_account>.github.io/layout_stars/report/html_report/)
+- [DEMO LINK](https://<adrian200002>.github.io/layout_stars/)
+- [TEST REPORT LINK](https://<adrian200002>.github.io/layout_stars/report/html_report/)
```
mvjl000, line 4: Please pay attention to those links... (don't put `<>` inside of them)
And also attach them to PR description
```suggestion
- [DEMO LINK](https://adrian200002.github.io/layout_stars/)
- [TEST REPORT LINK](https://adrian200002.github.io/layout_stars/report/html_report/)
```

src/style.css
```diff
@@ -0,0 +1,47 @@
...
+.stars {
+  display: flex;
+}
+
+.stars__star {
+  background-image: url("images/star-active.svg");
+  background-repeat: no-repeat;
+  background-position: center;
+  width: 16px;
+  height: 16px;
```
mvjl000, line 14: This value `(16px)` could be in a variable

src/style.css
```diff
@@ -0,0 +1,47 @@
...
+  margin-right: 4px;
```
mvjl000, line 15: Note that the last star doesn't need marigin (try something with `:not `and `:last-child` pseudo classes)

src/style.css
```diff
@@ -0,0 +1,47 @@
...
+  background-image: url("images/star.svg");
+  width: 16px;
+  height: 16px;
+}
+
+.stars--2 div:not(:nth-child(1)):not(:nth-child(2)) {
+  background-image: url("images/star.svg");
+  width: 16px;
+  height: 16px;
+}
```
mvjl000, line 47: A couple of thing to change here:

1.  No need to specify height and width again. You want to put here only the values that are changing,
2. Instead of targeting `div`, use `stars__star` class,
3. As you can see, the values inside of selectors are repeating. Why don't you combine those selectors with comma `,`,
4. There is a better way to specify which child should be affected. Instead of using fixed values like `:nth-child(4)` try to make a use of functional notation. Check it out [in docs](https://developer.mozilla.org/en-US/docs/Web/CSS/:nth-child)

mvjl000 (changes requested):

src/style.css
```diff
@@ -0,0 +1,44 @@
+:root {
+  --starheight: 16px;
+  --starwidth: 16px;
```
mvjl000, line 3: Star has the same height and width. Only one variable needed.
```suggestion
  --star-dimensions: 16px;
```

src/style.css
```diff
@@ -0,0 +1,44 @@
...
+  background-image: url("images/star.svg");
+}
+
+.stars--4 .stars__star:nth-child(n + 5) {
+  background-image: url("images/star.svg");
+}
+
+.stars--2 .stars__star:nth-child(n + 3) {
+  background-image: url("images/star.svg");
+}
```
mvjl000, line 44: Please combine those selectors with comma `,`

mvjl000 (changes requested):
Just two little adjustments and it's perfect

mvjl000 (approved):
Excellent!

PR #3677, opened 2023-10-11
task

dawidpizon:


src/style.css
```diff
@@ -0,0 +1,28 @@
...
+  display: flex;
+}
+
+.stars__star {
+  width: 50px;
+  height: 50px;
+  background-image: url(images/star.svg);
+  margin: 0;
+  flex: 1;
+}
```
mvjl000, line 11: 1. Star's dimensions should be waaay smaller. Try 16px and put it in a variable since this value will repeat
2. Use `background-repeat: no-repeat;`. Keep in mind that according to the checklist all background properties should be combined using a shorthand.

src/style.css
```diff
@@ -0,0 +1,28 @@
+.starts {
```
mvjl000, line 1: I think it is a typo ;)
```suggestion
.stars {
```

src/style.css
```diff
@@ -0,0 +1,28 @@
...
+
+.stars__star {
+  width: 50px;
+  height: 50px;
+  background-image: url(images/star.svg);
+  margin: 0;
+  flex: 1;
+}
+
+.body {
```
mvjl000, line 13: selector for body would look a lot better at the top of the document

Btw it is not affecting anything because you are targeting `body` class. Remove the dot and it will be applied to body tag ;)
```suggestion
body {
```

src/style.css
```diff
@@ -0,0 +1,28 @@
...
+}
+
+.stars--0 .stars__star:nth-child(1),
+.stars--1 .stars__star:nth-child(2),
+.stars--2 .stars__star:nth-child(3),
+.stars--3 .stars__star:nth-child(4),
+.stars--4 .stars__star:nth-child(5),
+.stars--5 .stars__star:nth-child(5) {
+  background-image: url("images/star-active.svg");
+}
```
mvjl000, line 28: You shouldn't target explicitly i.e. `:nth-child(4)`.  Instead use functional notation (like `-n + 4`). Please Look for it [in the docs](https://developer.mozilla.org/en-US/docs/Web/CSS/:nth-child).

mvjl000 (changes requested):

mvjl000 (changes requested):
Please apply my suggestions before re-requesting the review

PR #3678, opened 2023-10-11
Stars task

szymanski-marcin:
- [DEMO LINK](https://szymanski-marcin.github.io/layout_stars/)
- [TEST REPORT LINK](https://szymanski-marcin.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,30 @@
+.main-content {
+  margin: 0;
+}
+
+.stars {
+  display: flex;
+}
+
+.stars__item {
+  width: 16px;
+  height: 16px;
+  margin-right: 4px;
+  background: url(./images/star.svg);
+  background-repeat: no-repeat;
+  background-position: center;
+}
+
+.stars__item:last-child {
+  margin-right: 0;
+}
+
+.stars--1 .stars__item:nth-child(-n + 1),
+.stars--2 .stars__item:nth-child(-n + 2),
+.stars--3 .stars__item:nth-child(-n + 3),
+.stars--4 .stars__item:nth-child(-n + 4),
+.stars--5 .stars__item:nth-child(-n + 5) {
+  background: url(./images/star-active.svg);
+  background-repeat: no-repeat;
+  background-position: center;
+}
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,30 @@
...
+  margin: 0;
+}
+
+.stars {
+  display: flex;
+}
+/*  s*/
+.stars__item {
+  width: 16px;
+  height: 16px;
```
mvjl000, line 11: This value could be in a variable

src/style.css
```diff
@@ -0,0 +1,30 @@
...
+  margin-right: 4px;
+  background: url(./images/star.svg);
+  background-repeat: no-repeat;
+  background-position: center;
+}
+
+.stars__item:last-child {
+  margin-right: 0;
+}
```
mvjl000, line 20: You can set `margin-right: 4px;` here and make a use of :not pseudo class to exclude last child

src/style.css
```diff
@@ -0,0 +1,30 @@
...
+
+.stars--1 .stars__item:nth-child(-n + 1),
+.stars--2 .stars__item:nth-child(-n + 2),
+.stars--3 .stars__item:nth-child(-n + 3),
+.stars--4 .stars__item:nth-child(-n + 4),
+.stars--5 .stars__item:nth-child(-n + 5) {
+  background: url(./images/star-active.svg);
+  background-repeat: no-repeat;
+  background-position: center;
```
mvjl000, line 29: Those 2 values are not changing so you can get rid of them

szymanski-marcin: I had to shorten them into one line, otherwise tests were failed. All other recommendations are fixed.

src/style.css
```diff
@@ -0,0 +1,31 @@
...
+.stars__item:not(:last-child) {
+  margin-right: 4px;
+}
+
+.stars--1 .stars__item:nth-child(-n + 1),
+.stars--2 .stars__item:nth-child(-n + 2),
+.stars--3 .stars__item:nth-child(-n + 3),
+.stars--4 .stars__item:nth-child(-n + 4),
+.stars--5 .stars__item:nth-child(-n + 5) {
+  background: url(./images/star-active.svg) no-repeat center;
```
mvjl000, line 30: Please note that the only value you want to change here is a background image.
```suggestion
  background-image: url(./images/star-active.svg);
```

mvjl000 (changes requested):
One little thing and the task is done!

mvjl000 (approved):
Nailed it!

PR #3680, opened 2023-10-11
add task solution

K4rol4:


src/style.css
```diff
@@ -0,0 +1,27 @@
...
+  align-items: center;
+}
+
+.stars__star {
+  width: 16px;
+  height: 16px;
+  background-image: url("images/star.svg");
+  background-size: contain;
+  background-repeat: no-repeat;
+}
```
mvjl000, line 19: You need some space between stars. Try to add margin ;)

Also I don't think you will need `background-size`, `background-position: center;` will do the trick.

mvjl000 (changes requested):
You are very close!

One more thing... Please add correct github pages links and paste them into PR description. It should look like this:
`https://k4rol4.github.io/layout_stars/` 
not this:
`https://github.com/K4rol4/layout_stars`
You just need to replace <> with your username

mvjl000 (approved):
That's gooooood! 🔥

PR #3727, opened 2023-10-20
Stars

Denys209818:
I had troubles when I launched tests but they are related to image format because in tests was indicated PNG format but project stars have SVG code. Perhaps, May I improve my code?

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -8,6 +8,131 @@
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Stars</h1>
+    <div class="stars stars--0">
+      <div class="stars__star">
+
+      </div>
+
+      <div class="stars__star">
+
+      </div>
+
+      <div class="stars__star">
+
+      </div>
+
+      <div class="stars__star">
+
+      </div>
+
+      <div class="stars__star">
+
+      </div>
+    </div>
+    <div class="stars stars--1">
+      <div class="stars__star">
+
+      </div>
+
+      <div class="stars__star">
+
+      </div>
+
+      <div class="stars__star">
+
+      </div>
+
+      <div class="stars__star">
+
+      </div>
+
+      <div class="stars__star">
+
+      </div>
+    </div>
+    <div class="stars stars--2">
+      <div class="stars__star">
+
+      </div>
+
+      <div class="stars__star">
+
+      </div>
+
+      <div class="stars__star">
+
+      </div>
+
+      <div class="stars__star">
+
+      </div>
+
+      <div class="stars__star">
+
+      </div>
+    </div>
+    <div class="stars stars--3">
+      <div class="stars__star">
+
+      </div>
+
+      <div class="stars__star">
+
+      </div>
+
+      <div class="stars__star">
+
+      </div>
+
+      <div class="stars__star">
+
+      </div>
+
+      <div class="stars__star">
+
+      </div>
+    </div>
+    <div class="stars stars--4">
+      <div class="stars__star">
+
+      </div>
+
+      <div class="stars__star">
+
+      </div>
+
+      <div class="stars__star">
+
+      </div>
+
+      <div class="stars__star">
+
+      </div>
+
+      <div class="stars__star">
+
+      </div>
+    </div>
+    <div class="stars stars--5">
+      <div class="stars__star">
+
+      </div>
+
+      <div class="stars__star">
+
+      </div>
+
+      <div class="stars__star">
+
+      </div>
+
+      <div class="stars__star">
+
+      </div>
+
+      <div class="stars__star">
+
+      </div>
+    </div>
   </body>
 </html>
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,94 @@
+body {
+  margin: 0;
+}
+
+.stars {
+  display: flex;
+  justify-content: space-between;
+  flex-direction: row-reverse;
+  width: 96px;
+}
+
+.stars__star {
+  width: 16px;
+  height: 16px;
+  border: 2px solid transparent;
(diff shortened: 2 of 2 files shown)


Denys209818:
I found a problem. I eliminated space-between and added margin-right. In addition, I should implement width: fit-content, because the content would be displayed on the right side of a page.

src/style.css
```diff
@@ -0,0 +1,94 @@
...
+  background-position: center;
+  background-repeat: no-repeat;
+}
+
+.stars--5 > .stars__star {
+  background-image: url(images/star.svg);
+  background-size: cover;
+  background-position: center;
+  background-repeat: no-repeat;
+}
```
anastasiiavorobiova, line 59:
```suggestion

```
Avoid code repeating

src/style.css
```diff
@@ -0,0 +1,94 @@
+body {
+  margin: 0;
+}
+
+.stars {
+  display: flex;
+  flex-direction: row-reverse;
+  width: fit-content;
```
anastasiiavorobiova, line 8: background-image: url(images/star.svg);
  background-size: cover;
```suggestion
  width: fit-content;
  background-image: url(images/star.svg);
  background-size: cover;
  background-position: center;
  background-repeat: no-repeat;
```

src/style.css
```diff
@@ -0,0 +1,94 @@
...
+  background-position: center;
+  background-repeat: no-repeat;
+}
+
+.stars--5 > .stars__star:nth-of-type(n) {
+  background-image: url(images/star-active.svg);
+  background-size: cover;
+  background-position: center;
+  background-repeat: no-repeat;
+}
```
anastasiiavorobiova, line 94:
```suggestion
.stars--1 > .stars__star:last-child,
....other selectors separated by comme
.stars--3 > .stars__star:nth-of-type(n+3) {
  background-image: url(images/star-active.svg);
}
```
The same issue here

anastasiiavorobiova (changes requested):
Could you please add a demo and test report links here

src/style.css
```diff
@@ -0,0 +1,32 @@
...
+.stars {
+  display: flex;
+  flex-direction: row-reverse;
+  width: fit-content;
+}
+
+.stars__star {
+  width: 16px;
+  height: 16px;
+  border: 2px solid transparent;
```
Andriy-Shymkiv, line 14: why you need this border? it is strange

src/style.css
```diff
@@ -0,0 +1,32 @@
...
+
+.stars--1 > .stars__star:last-child,
+.stars--2 > .stars__star:nth-of-type(n+4),
+.stars--3 > .stars__star:nth-of-type(n+3),
+.stars--4 > .stars__star:nth-of-type(n+2),
+.stars--5 > .stars__star:nth-of-type(n) {
+  background-image: url(images/star-active.svg);
+  background-size: cover;
+  background-position: center;
+  background-repeat: no-repeat;
```
Andriy-Shymkiv, line 31: you don't need this styles here, you already set it above

src/style.css
```diff
@@ -0,0 +1,32 @@
+body {
+  margin: 0;
+}
+
+.stars {
+  display: flex;
+  flex-direction: row-reverse;
+  width: fit-content;
```
Andriy-Shymkiv, line 8: please remove it, this styles is useless. also after removing you should see that the starts are placed in reversed order, so you need to fix it in `nth-of-type` pseudo class below
```suggestion
```

Andriy-Shymkiv (changes requested):

src/style.css
Denys209818: If I hadn't added border, there would be stars without right proportions, and tests wouldn't work correctly!

src/style.css
Denys209818: It doesn't work correctly

src/index.html
```diff
@@ -8,6 +8,115 @@
...
+      </div>
+
+      <div class="stars__star">
+        <div class="star--item"></div>
+      </div>
+
+      <div class="stars__star">
+        <div class="star--item"></div>
+      </div>
+    </div>
```
OleziO, line 31:
```suggestion
    <div class="stars stars--0">
      <div class="stars__star"></div>
      <div class="stars__star"></div>
      <div class="stars__star"></div>
      <div class="stars__star"></div>
      <div class="stars__star"></div>
    </div>
```

Optimize code

OleziO: Check all other similar cases.

src/style.css
```diff
@@ -0,0 +1,33 @@
...
+  background-repeat: no-repeat;
+}
+
+.stars--1 > .stars__star:first-child > .star--item,
+.stars--2 > .stars__star:nth-of-type(-n + 2) > .star--item,
+.stars--3 > .stars__star:nth-of-type(-n + 3) > .star--item,
+.stars--4 > .stars__star:nth-of-type(-n + 4) > .star--item,
+.stars--5 > .stars__star:nth-of-type(n) > .star--item {
+  background-image: url(images/star-active.svg);
+}
```
OleziO, line 33: After optimization it will be like:

```suggestion
.stars--0 :nth-child(-n + 1),
.stars--1 :nth-child(-n + 2),
.stars--2 :nth-child(-n + 3),
.stars--3 :nth-child(-n + 4),
.stars--4 :nth-child(-n + 5) {
  background-image: url(images/star-active.svg);
}
```

OleziO (changes requested):

src/style.css
VitaliyBodnar: you need to fix that by using paddings and margins but not borders

VitaliyBodnar (approved):

PR #3728, opened 2023-10-21
created stars

Dev1NA:
- [DEMO LINK](https://Dev1NA.github.io/layout_stars/)
- [TEST REPORT LINK](https://Dev1NA.github.io/layout_stars/report/html_report/)

[x] Code follows all the Code Style Rules ❗️

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -8,6 +8,47 @@
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Stars</h1>
+    <div class="stars stars--0">
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+    </div>
+    <div class="stars stars--1">
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+    </div>
+    <div class="stars stars--2">
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,52 @@
...
+}
+
+/* .stars--0:nth-child(n),
+.stars--1:nth-child(n + 2),
+.stars--2:nth-child(n + 3),
+.stars--3:nth-child(n + 4),
+.stars--4:nth-child(n + 5) {
+  background-image: url("./images/star.svg");
+} */
+
```
lerastarynets, line 29: why did you comment loop? pls use loop for styles instead of adding styleone by one

src/index.html
```diff
@@ -8,6 +8,47 @@
...
   <body>
-    <h1>Stars</h1>
+    <div class="stars stars--0">
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+    </div>
+    <div class="stars stars--1">
```
lerastarynets, line 18:
```suggestion
    </div>
    
    <div class="stars stars--1">
```
pls add spaces between sibling `stars stars--[number]` blocks

lerastarynets (changes requested):
Nice start! Let's fix the solution according to the checklist

src/style.css
```diff
@@ -0,0 +1,28 @@
...
+.stars {
+  display: flex;
+}
+
+.stars__star:first-child {
+  margin-left: 0;
+}
+
+.stars--1 .stars__star:first-child,
```
OleziO, line 22:
```suggestion
.stars--1 .stars__star:nth-child(-n + 1),
```

Replace with this code to make all code the same style.

OleziO (changes requested):

Andriy-Shymkiv (approved):

PR #3730, opened 2023-10-23
add task solution

Xilston:
- [DEMO LINK](https://xilston.github.io/layout_stars/)
- [TEST REPORT LINK](https://xilston.github.io/layout_stars/report/html_report/)
- [+ ] Yellow stars are added with container modifier + pseudo-selector (NO extra classes)
- [+ ] Each BEM block has its own separate file
- [ +] All `Typical Mistakes` from `BEM` lesson theory are checked.
- [+ ] Code follows all the [Code Style Rules ❗️](./checklist.md)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,39 @@
+* {
+  margin: 0;
+}
+
+.stars {
+  display: flex;
+}
+
+.stars__star {
+  background-image: url(images/star.svg);
+  width: 16px;
+  height: 16px;
+  background-repeat: no-repeat;
+  background-position: center;
+}
+
+.stars__star:not(:last-child) {
+  margin-right: 4px;
+}
+
+.stars--1 .stars__star:nth-child(-n+1) {
+  background-image: url(images/star-active.svg);
+}
+
+.stars--2 .stars__star:nth-child(-n+2) {
+  background-image: url(images/star-active.svg);
+}
+
+.stars--3 .stars__star:nth-child(-n+3) {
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,39 @@
...
+
+.stars {
+  display: flex;
+}
+
+.stars__star {
+  background-image: url(images/star.svg);
+  width: 16px;
+  height: 16px;
+  background-repeat: no-repeat;
```
artemforsoff, line 13: It would be prettier if the properties for the background were grouped together ✨.

src/style.css
```diff
@@ -0,0 +1,39 @@
...
+.stars__star:not(:last-child) {
+  margin-right: 4px;
+}
+
+.stars--1 .stars__star:nth-child(-n+1) {
+  background-image: url(images/star-active.svg);
+}
+
+.stars--2 .stars__star:nth-child(-n+2) {
+  background-image: url(images/star-active.svg);
```
artemforsoff, line 26: Here you have used the same CSS property with the same value for 5 different selectors, it would be better to merge them. Check the example [here](https://developer.mozilla.org/en-US/docs/Learn/CSS/Building_blocks/Selectors#selector_lists).

artemforsoff:
There is space not only between columns but also between rows. Fix it 🤟🏻
image
image

artemforsoff (approved):
Very good work 🤩
There are minor edits, check them out.

PR #3731, opened 2023-10-23
layout_stars

Poviakalo:
- [DEMO LINK](https://Poviakalo.github.io/layout_stars/)
- [TEST REPORT LINK](https://Poviakalo.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -3,13 +3,7 @@ div {
   margin: 0;
 }
 
-/* .block {
-  width: 96px;
-  height: 96px;
-} */
-
 .stars {
-  /* width: 96px; */
   display: flex;
   justify-content: flex-start;
 }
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,30 @@
+body,
+div {
+  margin: 0;
+}
+
+.stars {
+  display: flex;
+  justify-content: flex-start;
```
xiimiik, line 8: its a default value, no need to write it

src/style.css
```diff
@@ -0,0 +1,30 @@
+body,
+div {
+  margin: 0;
+}
```
xiimiik, line 4: div has no marigin by default

xiimiik (approved):
nice work! left some comments :)

PR #3745, opened 2023-10-27
stars task solution 1

GeorgiusX:
- [DEMO LINK](https://GeorgiusX.github.io/layout_stars/)
- [TEST REPORT LINK](https://GeorgiusX.github.io/layout_stars/report/html_report/)

- [x ] Yellow stars are added with container modifier + pseudo-selector (NO extra classes)
- [x ] All `Typical Mistakes` from `BEM` lesson theory are checked.
- [x ] Code follows all the [Code Style Rules ❗️](./checklist.md)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -8,6 +8,54 @@
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Stars</h1>
+    <main class="container">
+      <ul class="stars stars--0">
+        <li class="stars__star"></li>
+        <li class="stars__star"></li>
+        <li class="stars__star"></li>
+        <li class="stars__star"></li>
+        <li class="stars__star"></li>
+      </ul>
+
+      <ul class="stars stars--1">
+        <li class="stars__star"></li>
+        <li class="stars__star"></li>
+        <li class="stars__star"></li>
+        <li class="stars__star"></li>
+        <li class="stars__star"></li>
+      </ul>
+
+      <ul class="stars stars--2">
+        <li class="stars__star"></li>
+        <li class="stars__star"></li>
+        <li class="stars__star"></li>
+        <li class="stars__star"></li>
+        <li class="stars__star"></li>
+      </ul>
+
+      <ul class="stars stars--3">
+        <li class="stars__star"></li>
+        <li class="stars__star"></li>
+        <li class="stars__star"></li>
+        <li class="stars__star"></li>
+        <li class="stars__star"></li>
+      </ul>
+
+      <ul class="stars stars--4">
+        <li class="stars__star"></li>
+        <li class="stars__star"></li>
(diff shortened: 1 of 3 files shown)


GeorgiusX:
I can not space out the individual starts. I cant set the width and height of 16px... It feels like i can not treat each star as a seperate elelment because they are a cut from one mash of starts in the background picture...Im confused here.

src/index.html
```diff
@@ -8,6 +8,54 @@
...
   <body>
-    <h1>Stars</h1>
+    <main class="container">
+      <ul class="stars stars--0">
+        <li class="stars__star"></li>
+        <li class="stars__star"></li>
+        <li class="stars__star"></li>
+        <li class="stars__star"></li>
+        <li class="stars__star"></li>
+      </ul>
```
VitaliyBondarenko1982, line 18: main 
```suggestion
    <main class="container">
      <div class="stars stars--0">
        <div class="stars__star"></div>
        <div class="stars__star"></div>
        <div class="stars__star"></div>
        <div class="stars__star"></li>
        <div class="stars__star"></div>
      </div>
```
stars row shouldn't be a list. it is just a row of stars and not have any additional information for which will be nessasary to make a list.

src/style.css
```diff
@@ -0,0 +1,88 @@
...
+  height: 78px;
+} */
+
+.stars {
+  display: flex;
+  margin: 0;
+  padding: 0;
+  width: 80px;
+  height: 12px;
+}
```
VitaliyBondarenko1982, line 17:
```suggestion
.stars {
  display: flex;
}
```
do not set widht and height for stars. 
content of stars will make size of row
margin and padding will be redundant after change ul to div

src/style.css
```diff
@@ -0,0 +1,88 @@
...
+  background-image: url(images/star-active.svg);
+}
+
+.stars--5 li:nth-child(4) {
+  background-image: url(images/star-active.svg);
+}
+
+.stars--5 li:nth-child(5) {
+  background-image: url(images/star-active.svg);
+}
```
VitaliyBondarenko1982, line 88: do not hardcode each active star.
it can be more flexible solution. 
also use classes instead tags for styling. and if  several selectors should have same styles - they are can be styling throw comma.
```suggestion
.stars--1 .stars__star:nth-child(-n+1),
.stars--2 .stars__star:nth-child(-n+2),
.stars--3 .stars__star:nth-child(-n+3),
.stars--4 .stars__star:nth-child(-n+4),
.stars--5 .stars__star:nth-child(-n+5) {
  background-image: url("./images/star-active.svg");

```

VitaliyBondarenko1982 (changes requested):
Please check comments and try to fix tests

IvanFesenko (approved):
Well done

PR #3769, opened 2023-11-06
add task solution

OksanaNastenko:
- [DEMO LINK](https://OksanaNastenko.github.io/layout_stars/)
- [TEST REPORT LINK](https://OksanaNastenko.github.io/layout_stars/report/html_report/)

- [x] Code follows all the [Code Style Rules ❗️](https://github.com/mate-academy/layout_stars/blob/master/checklist.md)

### DIFF
--- a/src/styles/blocks/stars.css
+++ b/src/styles/blocks/stars.css
@@ -1,0 +1,3 @@
+.stars {
+  display: flex;
+}
--- a/readme.md
+++ b/readme.md
@@ -21,8 +21,8 @@ Follow
 
 ❗️ Replace `<your_account>` with your Github username and copy the links to `Pull Request` description:
 
-- [DEMO LINK](https://<your_account>.github.io/layout_stars/)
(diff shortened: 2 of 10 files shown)


maxim2310 (changes requested):
you dont need separate files for every modificator, combine it and move these styles to star.css
image

src/styles/blocks/stars.css
```diff
@@ -0,0 +1,23 @@
+.stars {
+  display: flex;
+}
+
+.stars--1 :first-child {
+  background-image: url(../../images/star-active.svg);
```
maxim2310, line 6: this line is duplicated 5 times, need to combine this 5 rules

maxim2310 (changes requested):
Good work, but lets make some improvement 

star is not a block, star is a element of stars block so need to move all these styles to file stars.css
image

volodymyr-soltys97 (approved):
Well done 🔥

PR #3796, opened 2023-11-14
Develop

manch0ffline:
- [DEMO LINK](https://manch0ffline.github.io/layout_stars/)
- [TEST REPORT LINK](https://manch0ffline.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -8,6 +8,47 @@
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Stars</h1>
+    <div class="stars stars--0">
+      <div class="stars_star"></div>
+      <div class="stars_star"></div>
+      <div class="stars_star"></div>
+      <div class="stars_star"></div>
+      <div class="stars_star"></div>
+    </div>
+    <div class="stars stars--1">
+      <div class="stars_star"></div>
+      <div class="stars_star"></div>
+      <div class="stars_star"></div>
+      <div class="stars_star"></div>
+      <div class="stars_star"></div>
+    </div>
+    <div class="stars stars--2">
+      <div class="stars_star"></div>
+      <div class="stars_star"></div>
+      <div class="stars_star"></div>
+      <div class="stars_star"></div>
+      <div class="stars_star"></div>
+    </div>
+    <div class="stars stars--3">
+      <div class="stars_star"></div>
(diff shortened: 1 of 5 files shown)


src/style.css
```diff
@@ -0,0 +1,25 @@
...
+  padding: 0;
+}
+
+.stars_star {
+  width: 16px;
+  height: 16px;
+  background-image: url(images/star.svg);
+  background-repeat: no-repeat;
+  background-position: center;
+  margin-right: 4px;
```
Viktor-Kost, line 12: last star shouldn't have `margin-right`, you can use `:not(:last-child)`

src/style.css
```diff
@@ -0,0 +1,25 @@
...
+  margin: 0;
+  padding: 0;
+}
+
+.stars_star {
+  width: 16px;
+  height: 16px;
+  background-image: url(images/star.svg);
+  background-repeat: no-repeat;
+  background-position: center;
```
Viktor-Kost, line 11: here is you can use background `shorthand`

src/style.css
```diff
@@ -0,0 +1,25 @@
+* {
```
Viktor-Kost, line 1: it's not a good practice to reset values like this (using `*` selector). 
Reset `margin` and `padding` for tag selectors instead

src/index.html
```diff
@@ -8,6 +8,47 @@
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Stars</h1>
+    <div class="stars stars--0">
+      <div class="stars_star"></div>
```
Viktor-Kost, line 12: following BEM methodology we should use `__` for block elements
```suggestion
      <div class="stars__star"></div>
```

Viktor-Kost (changes requested):

src/style.css
```diff
@@ -0,0 +1,27 @@
+body {
+  margin: 0;
+  padding: 0;
+}
+
+.stars__star {
+  width: 16px;
+  height: 16px;
+  background: url(images/star.svg) no-repeat center;
+  margin-right: 4px;
```
Viktor-Kost, line 10: no needs to set this value since you've used `.stars__star:not(:last-child)`
```suggestion
```

Viktor-Kost (approved):

PR #3801, opened 2023-11-15
add task solution

nyudenkova:
- [DEMO LINK](https://nyudenkova.github.io/layout_stars/)
- [TEST REPORT LINK](https://nyudenkova.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,29 @@
+body {
+  margin: 0;
+}
+
+/* #region stars */
+.stars-block {
+  display: flex;
+  justify-content: center;
+  align-items: center;
+}
+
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,29 @@
+body {
+  margin: 0;
+}
+
+/* #region stars */
```
nazarmatsevych, line 5: You can remove regions in your final solution

nazarmatsevych (approved):
Awesome 👍

src/style.css
nyudenkova: Okay, thank you!

PR #3806, opened 2023-11-16
add task solution

pivkopa:
- [DEMO LINK](https://pivkopa.github.io/layout_stars/)
- [TEST REPORT LINK](https://pivkopa.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -8,8 +8,7 @@
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <!-- <h1>Stars</h1> -->
-
+    
     <div class="stars stars--0">
       <div class="stars__star"></div>
       <div class="stars__star"></div>
--- a/src/style.css
+++ b/src/style.css
@@ -14,7 +14,10 @@ body {
   background-repeat: no-repeat;
   margin-right: 4px;
   background-position: center;
-  flex-shrink: 0;
+}
+
+.stars__star:last-child {
+  margin-right: 0;
 }
 
 .stars--5 .stars__star,


src/index.html
```diff
@@ -8,6 +8,55 @@
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Stars</h1>
+    <!-- <h1>Stars</h1> -->
```
artemforsoff, line 11: remove commented code
with git history there is no need for commented code.
and in this case, it does not bring anything useful

src/style.css
```diff
@@ -0,0 +1,26 @@
+body {
+  margin: 0;
+}
+
+.stars {
+  display: flex;
+}
+
+.stars__star {
+  display: flex;
```
artemforsoff, line 10: what is this rule for?
I checked and it works without this rule
image

src/style.css
artemforsoff, line 1: no space between rows
image

src/style.css
```diff
@@ -0,0 +1,26 @@
...
+  width: 16px;
+  height: 16px;
+  background-image: url(images/star.svg);
+  background-repeat: no-repeat;
+  margin-right: 4px;
```
artemforsoff, line 15: the last star has `margin-right`, I think it better to remove it for last start

artemforsoff (changes requested):
great job 🤟🏻
there are a couple of non-critical points, look at them

src/style.css
artemforsoff: hm, you are right
I see the requirement _Don't add vertical margins between blocks._
this can be solved with paddings 😁, but okay, we can skip this

artemforsoff (approved):
well done 🦾

PR #3821, opened 2023-11-19
Develop

combine143:
- [DEMO LINK](https://combine143.github.io/layout_stars/)
- [TEST REPORT LINK](https://combine143.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -5,9 +5,57 @@
     <meta name="viewport"
           content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0">
     <title>Stars</title>
-    <link rel="stylesheet" href="./style.css">
+    <link rel="stylesheet" href="style.css">
   </head>
   <body>
-    <h1>Stars</h1>
+
+    <h1 class="stars stars--0">
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+    </h1>
+
+    <h1 class="stars stars--1">
+      <div class="stars__star active"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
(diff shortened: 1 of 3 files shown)


src/index.html
```diff
@@ -8,6 +8,54 @@
...
+      <div class="stars__star active"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+    </h1>
+
+    <h1 class="stars stars--2">
+      <div class="stars__star active"></div>
+      <div class="stars__star active"></div>
```
loralevitska, line 30: remove class active, and use [nth-child](https://developer.mozilla.org/en-US/docs/Web/CSS/:nth-child)

src/style.css
```diff
@@ -0,0 +1,23 @@
+body {
+  margin: 0;
+}
+
+.stars__star {
+  background-image: url(images/star.svg);
+  height: 16px;
+  width: 16px;
+  background-repeat: no-repeat;
+  background-position: center;
```
loralevitska, line 10: Combine with [`background`](https://developer.mozilla.org/en-US/docs/Web/CSS/background) shorthand property

loralevitska (changes requested):
Almost perfect) Let's improve

src/index.html
```diff
@@ -8,6 +8,54 @@
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Stars</h1>
+
+    <h1 class="stars stars--0">
```
anastasiiavorobiova, line 12:
```suggestion
    <div class="stars stars--0">
```
h1-6 tags are tags for headings, consider using divs or uls

anastasiiavorobiova (changes requested):
Great work! Almost done

src/index.html
combine143: Thank you! Fixed it

anastasiiavorobiova (approved):
Awesome work!

PR #3829, opened 2023-11-20
add task solution

zakhar1996:
[DEMO LINK](https://zakhar1996.github.io/layout_stars/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,1 +1,27 @@
 /* add styles here */
+body {
+  margin: 0;
+}
+
+.star-block {
+  display: flex;
+}
+
+.stars__star {
+  background-image: url(images/star.svg);
+  width: 16px;
+  height: 16px;
+  background-repeat: no-repeat;
+  background-position: center;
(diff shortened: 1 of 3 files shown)


nazarmatsevych (changes requested):
Good work,  let's make some improvements:

image

src/style.css
```diff
@@ -1 +1,34 @@
 /* add styles here */
```
OleziO, line 1: Remove all comments from the code.

OleziO (changes requested):

src/style.css
```diff
@@ -1 +1,34 @@
...
+  width: 16px;
+  height: 16px;
+  background-repeat: no-repeat;
+  background-position: center;
+  margin-right: 4px;
+
+}
+
+
+.stars__star:nth-child(5n) {
```
nazarmatsevych, line 21: Add just one empty line between styles next time

nazarmatsevych (approved):
Well done 👍

PR #3833, opened 2023-11-22
add task solution

IvanVaniuk:
- [DEMO LINK](https://IvanVaniuk.github.io/layout_stars/)
- [TEST REPORT LINK](https://IvanVaniuk.github.io/layout_stars/report/html_report/)


nazarmatsevych (changes requested):
Looks great but let's remove margin-right from the last stars in a row
image

PR #3867, opened 2023-11-29
add task solution

yaneznayou:
- [DEMO LINK](https://yaneznayou.github.io/layout_stars/)
- [TEST REPORT LINK](https://yaneznayou.github.io/layout_stars/report/html_report/)

- [x ] Yellow stars are added with container modifier + pseudo-selector (NO extra classes)
- [x ] All `Typical Mistakes` from `BEM` lesson theory are checked.
- [ x] Code follows all the [Code Style Rules ❗️](./checklist.md)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,1 +1,21 @@
-/* add styles here */
+body {
+  margin: 0;
+  padding: 0;
+}
+
+.stars {
+  display: flex;
+}
+
+.star {
+  flex: 0 0 auto;
+  height: 16px;
+  width: 16px;
+  background-image: url('images/star.svg');
+  background-size: contain;
+  margin-right: 4px;
+}
+
+.star--active {
+  background-image: url('images/star-active.svg');
+}
--- a/readme.md
+++ b/readme.md
@@ -21,8 +21,8 @@ Follow
 
 ❗️ Replace `<your_account>` with your Github username and copy the links to `Pull Request` description:
 
-- [DEMO LINK](https://<your_account>.github.io/layout_stars/)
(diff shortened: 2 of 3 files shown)


artemforsoff:
The syntax of the checkboxes is broken, so they are displayed incorrectly for you.
Only one space is allowed in case of the inactive state, for the active state any spaces are not allowed.
- [ ] - inactive
- [x] - active

src/style.css
```diff
@@ -1 +1,21 @@
...
+.stars {
+  display: flex;
+}
+
+.star {
+  flex: 0 0 auto;
+  height: 16px;
+  width: 16px;
+  background-image: url('images/star.svg');
+  background-size: contain;
```
artemforsoff, line 15: the background image should not stretch to the entire size of the element
check design

src/style.css
```diff
@@ -1 +1,21 @@
...
+  margin-right: 4px;
```
artemforsoff, line 16: the last elements have `margin-right`, although this is not necessary

src/style.css
```diff
@@ -1 +1,21 @@
...
+  margin: 0;
+  padding: 0;
+}
+
+.stars {
+  display: flex;
+}
+
+.star {
+  flex: 0 0 auto;
```
artemforsoff, line 11: this rule doesn't affect anything

artemforsoff (changes requested):
overall good
I left some comments, please pay attention and fix them

SerhiiKirik (approved):
Nice 🔥 
But don't forget redeploy your solution after push with `npm run deploy`

PR #3883, opened 2023-12-04
add task solution

Onoprijcha:
- [DEMO LINK](https://onoprijcha.github.io/layout_stars/)
- [TEST REPORT LINK](https://onoprijcha.github.io/layout_stars/report/html_report/)

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -21,12 +21,12 @@ Follow
 
 ❗️ Replace `<your_account>` with your Github username and copy the links to `Pull Request` description:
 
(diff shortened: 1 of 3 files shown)


nazarmatsevych (changes requested):
Good start, let's make some improvements:

Remove margin-right from the last stars in a row (use last-child rule)

image

anastasiiavorobiova (changes requested):
Make sure to deploy your solution

PR #3923, opened 2023-12-13
add stars

AlinaYamchuk:

- [DEMO LINK](https://AlinaYamchuk.github.io/layout_stars/)
- [TEST REPORT LINK](https://AlinaYamchuk.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -1,6 +1,6 @@
 <!doctype html>
 <html lang="en">
-  <head>
+  <head class="header">
     <meta charset="UTF-8">
     <meta name="viewport"
           content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0">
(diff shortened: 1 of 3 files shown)


src/index.html
```diff
@@ -1,13 +1,60 @@
...
     <meta name="viewport"
           content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0">
     <title>Stars</title>
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Stars</h1>
+    <div class="rating__stars">
+
+      <div class="stars__stars--0">
```
nazarmatsevych, line 13:
```suggestion
      <div class="stars stars--0">
```

stars--0 should be a modifier class, fix everywhere to pass tests

nazarmatsevych (changes requested):
Good start, let's make some improvements:

image

src/index.html
AlinaYamchuk: Fixed

PR #3938, opened 2023-12-19
Add stars

AlinaYamchuk:

- [DEMO LINK](https://AlinaYamchuk.github.io/layout_stars/)
- [TEST REPORT LINK](https://AlinaYamchuk.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -9,7 +9,6 @@
   </head>
   <body>
     <div class="rating__stars">
-
       <div class="stars stars--0">
         <div class="stars__star"></div>
         <div class="stars__star"></div>
--- a/src/style.css
+++ b/src/style.css
@@ -16,6 +16,10 @@ body {
   margin-right: 4px;
 }
 
+.stars__star:last-child {
+  margin: 0;
+}
+
 .stars--1 :nth-child(-n + 1),
 .stars--2 :nth-child(-n + 2),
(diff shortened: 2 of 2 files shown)


src/index.html
```diff
@@ -1,13 +1,63 @@
...
           content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0">
-    <title>Stars</title>
     <link rel="stylesheet" href="./style.css">
+    <title>Stars</title>
   </head>
   <body>
-    <h1>Stars</h1>
+    <div class="rating__stars">
+
+      <div class="stars stars--0">
```
nazarmatsevych, line 13: Redundant empty line between parent and child elements

nazarmatsevych (changes requested):
Remove margin-right from the last stars in a row (use last-child rule)
image

src/index.html
AlinaYamchuk: Done

src/style.css
```diff
@@ -1 +1,29 @@
-/* add styles here */
+body {
+  margin: 0;
+}
+
+.stars {
+  display: flex;
+  flex-direction: row;
```
Viktor-Kost, line 7: this is default flex value
```suggestion
```

PR #3946, opened 2023-12-22
add solution

AChubai89:
- [DEMO LINK](https://AChubai89.github.io/layout_stars/)
- [TEST REPORT LINK](https://AChubai89.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,1 +1,40 @@
 /* add styles here */
+body {
+margin: 0;
+}
+
+.stars {
+display: flex;
+justify-content: center;
+align-items: center;
+}
+
+.stars__star {
+margin-right: 4px;
+width: 16px;
+height: 16px;
+background-image: url(./images/star.svg);
+background-repeat: no-repeat;
+background-position: center;
+flex-shrink: 0;
+}
+
+.stars--1 .stars__star:nth-child(-n + 1) {
+background-image: url(./images/star-active.svg);
+}
+
+.stars--2 .stars__star:nth-child(-n + 2) {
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -1 +1,40 @@
...
+background-image: url(./images/star-active.svg);
+}
+
+.stars--4 .stars__star:nth-child(-n + 4) {
+background-image: url(./images/star-active.svg);
+}
+
+.stars--5 .stars__star:nth-child(-n + 5) {
+background-image: url(./images/star-active.svg);
+}
```
nazarmatsevych, line 40: combine selectors using comma,  example below:

.selector1,
.selector2 {
  your styles here...
}

nazarmatsevych (changes requested):
image

src/style.css
AChubai89: When I remove margin-right from the last stars in a row, i use the following: 
.stars__star:last-child {
  margin: 0;
} ...and after that I can't pass any tests.

src/style.css
```diff
@@ -1 +1,28 @@
 /* add styles here */
+body {
+margin: 0;
+}
+
+.stars {
+display: flex;
+justify-content: center;
+align-items: center;
+}
```
etojeDenys, line 10:
```suggestion
.stars {
  display: flex;
  justify-content: center;
  align-items: center;
}
```
looks like you forgot some spaces, let's fix it everywhere

etojeDenys (changes requested):

nazarmatsevych (approved):
🔥

PR #3949, opened 2023-12-23
Stars solution

PavloStelmaschuk:
- [DEMO LINK](https://PavloStelmaschuk.github.io/layout_stars/)
- [TEST REPORT LINK](https://PavloStelmaschuk.github.io/layout_stars/report/html_report/)
- [x]  All Typical Mistakes from BEM lesson theory are checked.
- [x]  Code follows all the [Code Style Rules ❗️](https://github.com/mate-academy/layout_stars/blob/master/checklist.md)


src/styles/blocks/stars.css
```diff
@@ -0,0 +1,32 @@
+.stars {
+  display: flex;
+}
+
+.stars__star {
+  background-image: url(../../images/star-active.svg);
```
VitaliyBondarenko1982, line 6:
```suggestion
  background-image: url(../../images/star.svg);
```

src/styles/blocks/stars.css
```diff
@@ -0,0 +1,32 @@
...
+  background-image: url(../../images/star.svg);
+}
+
+.stars--3 .stars__star:nth-child(n+4) {
+  background-image: url(../../images/star.svg);
+}
+
+.stars--4 .stars__star:nth-child(n+5) {
+  background-image: url(../../images/star.svg);
+}
```
VitaliyBondarenko1982, line 32:
```suggestion
.stars--1 :nth-child(-n+1),
.stars--2 :nth-child(-n+2),
.stars--3 :nth-child(-n+3),
.stars--4 :nth-child(-n+4),
.stars--5 :nth-child(-n+5) {
  background-image: url(../../images/star-active.svg);
}
```
suggest to make selectors more consistent with modifiers (same number in modifier and nth-child) and set by default star and in midifiers star-active.
it is possible to reduce amount of code by separate selector with same styles by comma.

VitaliyBondarenko1982 (changes requested):
Good job.
Suggest some refactor.

src/index.html
```diff
@@ -1,13 +1,54 @@
...
-    <h1>Stars</h1>
+  <body class="page__body">
+    <div class="stars stars--0">
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+    </div>
+    <div class="stars stars--1">
```
nazarmatsevych, line 18:
```suggestion
    <div class="stars stars--0">
      <div class="stars__star"></div>
      <div class="stars__star"></div>
      <div class="stars__star"></div>
      <div class="stars__star"></div>
      <div class="stars__star"></div>
    </div>
    
    <div class="stars stars--1">
```

Add an empty line between multiple sibling blocks of html,  fix all such cases

nazarmatsevych (changes requested):
Remove margin-right from the last stars in a row (use :last-child rule)
image

src/styles/blocks/stars.css
```diff
@@ -0,0 +1,25 @@
+.stars {
+  display: flex;
+}
+
+
+.stars__star {
+  background-image: url(../../images/star.svg);
+  background-repeat: no-repeat;
+  background-position: center;
```
loralevitska, line 9: `background` shorthand can be used

loralevitska (approved):
Well done)

PR #3958, opened 2023-12-29
Add solution

Vasya564:
- [DEMO LINK](https://Vasya564.github.io/layout_stars/)
- [TEST REPORT LINK](https://Vasya564.github.io/layout_stars/report/html_report/)


src/styles/stars-1.css
```diff
@@ -0,0 +1,3 @@
+.stars--1 > .stars__star:nth-child(1) {
+  background-image: url(../images/star-active.svg);
```
polosanya, line 2: There is no need to create a separate file for each stars modifier. Please delete them.
You can also combine the selectors with common style properties

polosanya (changes requested):
Almost done! Let's refactor your code a bit)

src/styles/stars-1.css
Vasya564: I had misunderstanding with this point "Each BEM block has its own separate file" so that is why I decided to seperate each stars modifier. I'll try to improve my code

stassokolov1 (approved):
Looks good to me, i'm approving it with only one minor comment🔥

PR #3986, opened 2024-01-06
Develop

tZsenya:

- [DEMO LINK](https://tZsenya.github.io/layout_stars/)
- [TEST REPORT LINK](https://tZsenya.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -8,6 +8,54 @@
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Stars</h1>
+    <container class="container">
+      <div class="stars--0">
+        <div class="darkstars"></div>
+        <div class="darkstars"></div>
+        <div class="darkstars"></div>
+        <div class="darkstars"></div>
+        <div class="darkstars"></div>
+      </div>
+
+      <div class="stars--1">
+        <div class="goldstars"></div>
+        <div class="darkstars"></div>
+        <div class="darkstars"></div>
+        <div class="darkstars"></div>
+        <div class="darkstars"></div>
+      </div>
+
+      <div class="stars--2">
+        <div class="goldstars"></div>
+        <div class="goldstars"></div>
+        <div class="darkstars"></div>
+        <div class="darkstars"></div>
+        <div class="darkstars"></div>
+      </div>
+
+      <div class="stars--3">
+        <div class="goldstars"></div>
+        <div class="goldstars"></div>
+        <div class="goldstars"></div>
+        <div class="darkstars"></div>
+        <div class="darkstars"></div>
+      </div>
+
+      <div class="stars--4">
+        <div class="goldstars"></div>
+        <div class="goldstars"></div>
+        <div class="goldstars"></div>
+        <div class="goldstars"></div>
+        <div class="darkstars"></div>
+      </div>
+
+      <div class="stars--5">
+        <div class="goldstars"></div>
+        <div class="goldstars"></div>
+        <div class="goldstars"></div>
+        <div class="goldstars"></div>
+        <div class="goldstars"></div>
+      </div>
+    </container>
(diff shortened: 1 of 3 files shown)


nazarmatsevych (changes requested):
remove margin-right from the last stars in a row
image

Proofread your code with task requirements
image

Don't add extra classes, use nth-child rule to change gray stars to gold one

src/index.html
```diff
@@ -8,6 +8,54 @@
...
   <body>
-    <h1>Stars</h1>
+    <container class="container">
+      <div class="stars--0">
+        <div class="darkstars"></div>
+        <div class="darkstars"></div>
+        <div class="darkstars"></div>
+        <div class="darkstars"></div>
+        <div class="darkstars"></div>
+      </div>
```
VitaliyBondarenko1982, line 18:
```suggestion
      <div class="stars stars--0">
        <div class="stars__star"></div>
        <div class="stars__star"></div>
        <div class="stars__star"></div>
        <div class="stars__star"></div>
        <div class="stars__star"></div>
      </div>
```
fix prev comments about task requirements.
no need make different classes for active or non-active star. use nth-child and stars--0, stars--1...stars--5 modifier.
set `stars_star` class for each star and `stars` for each stars row.

src/style.css
```diff
@@ -1 +1,69 @@
...
+  display: flex;
+}
+
+.stars--4 {
+  display: flex;
+}
+
+.stars--5 {
+  display: flex;
+}
```
VitaliyBondarenko1982, line 27:
```suggestion
.stars {
  display: flex;
}
```
it will be style for .stars block

src/style.css
```diff
@@ -1 +1,69 @@
...
+  margin: 0;
+}
+
+.stars--4 :last-child {
+  margin: 0;
+}
+
+.stars--5 :last-child {
+  margin: 0;
+}
```
VitaliyBondarenko1982, line 69:
```suggestion
.stars__star:last-child {
  margin-right: 0;
```
it can be shorter as every star will have same class

src/style.css
```diff
@@ -1 +1,69 @@
...
+
+.stars--4 {
+  display: flex;
+}
+
+.stars--5 {
+  display: flex;
+}
+
+.darkstars {
```
VitaliyBondarenko1982, line 29:
```suggestion
.stars__star {
```

src/style.css
```diff
@@ -1 +1,69 @@
...
+}
+
+.goldstars {
+  background-image: url(images/star-active.svg);
+  width: 16px;
+  height: 16px;
+  background-repeat: no-repeat;
+  background-position: center;
+  margin-right: 4px;
+}
```
VitaliyBondarenko1982, line 45: do not set here styles as already be settled for .stars__star.
here should be settled only background-image. use :nth-child pseudo class and stars--1 ... stars--5 modifiers for it

VitaliyBondarenko1982 (changes requested):
Good job.
Please check and fix comments.

src/style.css
```diff
@@ -1 +1,40 @@
...
+  background-image: url(images/star-active.svg);
+}
+
+.stars--4 :nth-child(-n+4) {
+  background-image: url(images/star-active.svg);
+}
+
+.stars--5 :nth-child(-n+5) {
+  background-image: url(images/star-active.svg);
+}
```
volodymyr-soltys97, line 36: These selectors must be separated by comma to avoid duplicate styles

volodymyr-soltys97 (changes requested):
Hi, your demo link doesn't work, you need to fix it
image

volodymyr-soltys97 (approved):
Good job 👍 
The margin right for the last star still not removed
image

PR #3991, opened 2024-01-07
Implementation of readme.md requirements

under-a:
- [DEMO LINK](https://under-a.github.io/layout_stars/)
- [TEST REPORT LINK](https://under-a.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,1 +1,28 @@
-/* add styles here */
+body {
+  margin: 0;
+}
+
+.stars {
+  display: inline-block;
+}
+
+.stars__group {
+  display: flex;
+}
+
+.stars__star {
+  background-image: url(./images/star.svg);
+  background-repeat: no-repeat;
+  background-position: center;
+  padding: 8px;
+  margin: 0;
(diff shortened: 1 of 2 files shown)


src/style.css
```diff
@@ -1 +1,28 @@
...
+.stars__group {
+  display: flex;
+}
+
+.stars__star {
+  background-image: url(./images/star.svg);
+  background-repeat: no-repeat;
+  background-position: center;
+  padding: 8px;
+  margin: 0;
```
nazarmatsevych, line 18:
```suggestion
  width: 16px;
  height: 16px;
```

16px due to figma

nazarmatsevych (changes requested):
Remove margin-right form the last stars in a row, use :last-child rule to fix it
image

under-a:
> Remove margin-right form the last stars in a row, use :last-child rule to fix it image

But tests still require this piece, should I replace it with something?

image

SanyaBratashchuk (approved):

PR #4027, opened 2024-01-15
Stars =)

Ellworts:
Yellow stars are added with container modifier + pseudo-selector (NO extra classes)
 [x] Code follows all the [Code Style Rules ❗️](https://github.com/Ellworts/layout_stars/blob/master/checklist.md)


anastasiiavorobiova (approved):
Great work!

PR #4048, opened 2024-01-20
addtask solution

Julia-na:
https://Julia-na.github.io/layout_stars/
https://Julia-na.github.io/layout_stars/report/html_report/

- [x] Code follows all the [Code Style Rules !](https://github.com/Julia-na/layout_stars/blob/master/checklist.md)


Esceype (approved):
Well done!

PR #4061, opened 2024-01-24
Create stars

Anothar:
- [DEMO LINK](https://anothar.github.io/layout_stars/)
- [TEST REPORT LINK](https://anothar.github.io/layout_stars/report/html_report/)
- [x]Each BEM block has its own separate file
- [x]All Typical Mistakes from BEM lesson theory are checked.
- [x]Code follows all the [Code Style Rules ❗️](https://github.com/Anothar/layout_stars/blob/master/checklist.md)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,1 +1,70 @@
 /* add styles here */
+html {
+  --active-star: url(/src/images/star-active.svg);
+  --star: url(/src/images/star.svg);
+}
+
+body {
+  margin: 0;
+  padding: 22px;
+  display: inline-block;
+}
+
+.stars {
+  display: flex;
+}
+
+.stars__star {
+  width: 16;
+  height: 16;
+  background-repeat: no-repeat;
+  background-position: center;
+}
+
+.stars:not(.stars--5) {
+  margin-bottom: 4px;
+}
+
+.stars__star:nth-last-child(n + 2) {
+  margin-right: 4px;
+}
+
+.stars--0> :nth-child(n + 1) {
+  background-image: var(--star);
+}
+
+.stars--1> :nth-child(-n + 1) {
+  background-image: var(--active-star);
+}
+
+.stars--1> :nth-child(n + 2) {
+  background-image: var(--star);
+}
+
+.stars--2> :nth-child(-n + 2) {
+  background-image: var(--active-star);
+}
+
+.stars--2> :nth-child(n + 3) {
+  background-image: var(--star);
+}
+
+.stars--3> :nth-child(-n + 3) {
+  background-image: var(--active-star);
+}
+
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -1 +1,70 @@
...
+
+.stars__star {
+  width: 16;
+  height: 16;
+  background-repeat: no-repeat;
+  background-position: center;
+}
+
+.stars:not(.stars--5) {
+  margin-bottom: 4px;
```
nazarmatsevych, line 25: redundant margin-bottom

src/style.css
```diff
@@ -1 +1,70 @@
 /* add styles here */
+html {
+  --active-star: url(/src/images/star-active.svg);
+  --star: url(/src/images/star.svg);
```
nazarmatsevych, line 4: You can set a default image by default to the `stars__star `class and change it using nth-chilt rule

src/style.css
```diff
@@ -1 +1,70 @@
...
+  background-image: var(--active-star);
+}
+
+.stars--4> :nth-child(n + 5) {
+  background-image: var(--star);
+}
+
+.stars--5> :nth-child(-n + 5) {
+  background-image: var(--active-star);
+}
```
nazarmatsevych, line 70: Combine selectors using comma, example below

.selector1,
.selector2 {
  your styles here...
}

nazarmatsevych (changes requested):
Good start, let's make some improvements

src/style.css
```diff
@@ -1 +1,34 @@
...
+.stars__star {
+  width: 16;
+  height: 16;
+  background-repeat: no-repeat;
+  background-position: center;
+  background-image: url(/src/images/star-active.svg);
+}
+
+.stars:not(.stars--5) {
+  margin-bottom: 4px;
```
TarasHoliuk, line 21: Do not ignore comments:
image

Write some answer or ask for more info (in the chat) if something isn't clear :slightly_smiling_face: 
The reason you added it is clear - but because of it tests will fail for example (BTW Backstop Report link doesn't work)

src/style.css
```diff
@@ -1 +1,34 @@
...
+}
+
+.stars__star:nth-last-child(n + 2) {
```
TarasHoliuk, line 24: `.stars__star:not(:last-child)` should also work but it looks a bit more understandable

FYI

TarasHoliuk (approved):
Approved :heavy_check_mark: 

But added 2 comments just FYI

src/style.css
Anothar: We discussed this issue with Nazar during the Q&A session and agreed that there is no need to do it, as the Figma layout has been updated

PR #4064, opened 2024-01-25
Layout__stars

Rusik-best86:
- [DEMO LINK](https://Rusik-best86.github.io/layout_stars/)
- [TEST REPORT LINK](https://Rusik-best86.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,1 +1,31 @@
-/* add styles here */
+@charset "UTF-8";
+
+* {
+  margin: 0;
+ }
+
+.stars {
+ display:flex;
+}
+
+.stars__star {
+  background-image:url(./images/star.svg);
+  width: 16px;
+  height: 16px;
+  background-position: 50%;
+  background-repeat: no-repeat;
+ }
+
+.stars__star:not(:first-child) {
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -1 +1,31 @@
-/* add styles here */
+@charset "UTF-8";
+
+* {
+  margin: 0;
+ }
```
polosanya, line 5: image

src/style.css
```diff
@@ -1 +1,31 @@
-/* add styles here */
+@charset "UTF-8";
```
polosanya, line 1: Why do you need it?

polosanya (changes requested):
Almost perfect!

src/style.css
```diff
@@ -1 +1,29 @@
-/* add styles here */
+body {
+  margin: 0;
+ }
```
Moroz-Dmytro, line 3:
```suggestion
body {
  margin: 0;
}
```

Moroz-Dmytro (approved):
Good work 🚀 
Really small comments

PR #4088, opened 2024-01-30
add task solution

DmitryKot96:
- [DEMO LINK](https://DmitryKot96.github.io/layout_stars/)
- [TEST REPORT LINK](https://DmitryKot96.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -8,6 +8,52 @@
     <link rel="stylesheet" href="./style.css">
   </head>
   <body>
-    <h1>Stars</h1>
+    <div class="stars stars--0">
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+    </div>
+
+    <div class="stars stars--1">
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+    </div>
+
+    <div class="stars stars--2">
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -1 +1,88 @@
...
+.stars--5 > .stars__star:nth-child(-n+5){
+  background-image: url(images/star-active.svg);
+  height: 16px;
+  width: 16px;
+  background-position: center;
+  background-repeat: no-repeat;
+  margin-right: 4px;
+}
+
+/* #endregion */
```
dffUqp, line 88: 1. Check how we can combine these selectors (try to use `,`)

2. I think we already apply these styles to `.stars__star`, so let's remove it here: 
```css
  height: 16px;
  width: 16px;
  background-position: center;
  background-repeat: no-repeat;
  margin-right: 4px;
  ```

File: src/style.css [hunk 2]
```diff
@@ -1 +1,88 @@
...
+  width: 16px;
+  background-position: center;
+  background-repeat: no-repeat;
+  margin-right: 4px;
+  flex-shrink: 0;
+}
+
+/* #endregion */
+
+/* #region 1star */
```
dffUqp, line 25: Let's remove comments like this

src/index.html
```diff
@@ -8,6 +8,52 @@
...
   </head>
   <body>
-    <h1>Stars</h1>
+    <div class="stars stars--0">
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+    </div>
```
dffUqp, line 17: Use semantic tags instead of `div` ones

dffUqp (changes requested):
LGTM 🔥 but needs a few minor fixes 🙂

BudnikOleksii (approved):
👍

PR #4096, opened 2024-01-30
add task solution

dionechik:
- [DEMO LINK](https://dionechik.github.io/layout_stars/)
- [TEST REPORT LINK](https://dionechik.github.io/layout_stars/report/html_report/)

- [х] Yellow stars are added with container modifier + pseudo-selector (NO extra classes)
- [х] Each BEM block has its own separate file
- [х] All `Typical Mistakes` from `BEM` lesson theory are checked.
- [х] Code follows all the [Code Style Rules ❗️](./checklist.md)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,1 +1,28 @@
-/* add styles here */
+body {
+  margin: 0;
+  padding: 0;
+}
+
+.stars {
+  display: flex;
+}
+
+.stars__star {
+  height: 16px;
+  width: 16px;
+  background-image: url(../src/images/star.svg);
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -1 +1,28 @@
-/* add styles here */
+body {
+  margin: 0;
+  padding: 0;
```
Viktor-Kost, line 3: body doesn't have default padding
```suggestion
```

Viktor-Kost (approved):
Great!

PR #4104, opened 2024-01-31
add task solution

IRONH4MMER:
- [x] Code follows all the [Code Style Rules ❗️](https://github.com/mate-academy/layout_stars/blob/master/checklist.md)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -1,13 +1,59 @@
 <!doctype html>
 <html lang="en">
+
   <head>
     <meta charset="UTF-8">
     <meta name="viewport"
-          content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0">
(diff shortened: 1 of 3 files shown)


src/index.html
```diff
@@ -1,13 +1,61 @@
...
+        <div class="stars__star"></div>
+        <div class="stars__star"></div>
+        <div class="stars__star"></div>
+        <div class="stars__star"></div>
+      </div>
+    </div>
+
   </body>
+
 </html>
```
polosanya, line 61:
```suggestion
    </div>
  </body>
</html>
```

src/style.css
```diff
@@ -1 +1,25 @@
...
+}
+
+.stars {
+  display: flex;
+}
+
+.stars__star {
+  height: 16px;
+  width: 16px;
+  margin-right: 4px;
```
polosanya, line 13: Make sure that the last star doesn't have margin-right

polosanya (approved):
Nice! I'm approving it, but please pay attention to the comments)
Please don't forget to add demo and test links to the PR description

PR #4109, opened 2024-01-31
taskComplete

JustinNowoczek:
- [DEMO LINK](https://tidetwo.github.io/layout_stars/)
- [TEST REPORT LINK](https://tidetwo.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -1,13 +1,56 @@
-<!doctype html>
+<!DOCTYPE html>
 <html lang="en">
   <head>
-    <meta charset="UTF-8">
-    <meta name="viewport"
-          content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0">
+    <meta charset="UTF-8" >
+    <meta
+      name="viewport"
+      content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0"
+    >
     <title>Stars</title>
-    <link rel="stylesheet" href="./style.css">
+    <link rel="stylesheet" href="./styles/style.css" >
   </head>
   <body>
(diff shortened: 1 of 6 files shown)


src/index.html
```diff
@@ -1,13 +1,56 @@
...
+      content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0"
+    >
     <title>Stars</title>
-    <link rel="stylesheet" href="./style.css">
+    <link rel="stylesheet" href="./styles/style.css" >
   </head>
   <body>
-    <h1>Stars</h1>
+    <div class="stars stars--0">
+      <div class="stars__star"></div>
```
asakevych, line 14: you can use self-closing tag here, because you don't have any content inside -> `<div ... />`

src/styles/stars.css
```diff
@@ -0,0 +1,26 @@
+/* #region Stars */
+
+.stars {
+  display: flex;
+  flex-direction: row;
+}
+
+.stars__star {
+  margin-right: 4px;
```
asakevych, line 9: please exclude last star of every row, there is no need for it to have some margin

src/styles/stars.css
```diff
@@ -0,0 +1,26 @@
+/* #region Stars */
+
+.stars {
+  display: flex;
+  flex-direction: row;
```
asakevych, line 5: row is a default property, no need to repeat it

asakevych (changes requested):
Good job!! :rocket:  just some small corrections are required

src/index.html
TideTwo: This is not something I can implement, as the linter has a rule against it. 
image

choeqq (approved):

PR #4112, opened 2024-02-01
add task solution

IrenaPopovych7:
- [ x] Yellow stars are added with container modifier + pseudo-selector (NO extra classes)
- [ x]  All Typical Mistakes from BEM lesson theory are checked.
- [x ]  Code follows all the [Code Style Rules ❗️](https://github.com/IrenaPopovych7/layout_stars/blob/master/checklist.md)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,1 +1,44 @@
 /* add styles here */
+body {
+  margin: 0;
+  display: flex;
+  flex-direction: column;
+  align-items: center;
+  justify-content: center;
+}
+
+.stars {
+  display: flex;
+  flex-direction: row;
+  align-items: center;
+
+}
+
+.stars__star {
+  margin-right: 4px;
+  width: 16px;
+  height: 16px;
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -1 +1,44 @@
...
+.stars__star {
+  margin-right: 4px;
+  width: 16px;
+  height: 16px;
+  background-image: url('./images/star.svg');
+  background-repeat: no-repeat;
+  background-position: center center;
+}
+
+.stars--1 :nth-child(-n + 1) {
```
anastasiiavorobiova, line 26:
```suggestion
.stars--1 :nth-child(-n + 1),
.stars--2 :nth-child(-n + 2),
... {
```
Consider using commas to group your selectors

anastasiiavorobiova (changes requested):
Great work! Almost done!

anastasiiavorobiova (approved):
Awesome!

PR #4122, opened 2024-02-02
first version

konradwasilewski444:
- [DEMO LINK](https://konradwasilewski444.github.io/layout_stars/)
- [TEST REPORT LINK](https://konradwasilewski444.github.io/layout_stars/report/html_report/)

 [x]- Yellow stars are added with container modifier + pseudo-selector (NO extra classes)
 [x]- Each BEM block has its own separate file
 [x]- All Typical Mistakes from BEM lesson theory are checked.
 [x]- Code follows all the [Code Style Rules ❗️


src/style.css
```diff
@@ -1 +1,44 @@
...
+
+.stars {
+  display: flex;
+  flex-direction: row;
+  align-items:center;
+}
+
+.stars:first-child {
+  margin-top: 0;
+}
```
nataliaklonowska, line 16: you don't need it :) margin is already 0
```suggestion
```

src/style.css
```diff
@@ -1 +1,44 @@
...
+.stars--3 .stars__star:nth-child(1),
+.stars--3 .stars__star:nth-child(2),
+.stars--3 .stars__star:nth-child(3),
+.stars--4 .stars__star:nth-child(1),
+.stars--4 .stars__star:nth-child(2),
+.stars--4 .stars__star:nth-child(3),
+.stars--4 .stars__star:nth-child(4),
+.stars--5 .stars__star:nth-child(n){
+  background-image: url(images/star-active.svg);
+}
```
nataliaklonowska, line 43: don't repeat all possible child elements instead use formula (an + b)
you can read more about it [here](https://css-tricks.com/how-nth-child-works/)

nataliaklonowska (changes requested):
Great work! 🚀 
just improve the use of :nth-child selector

asakevych (approved):
Great job!! :rocket: :fire: :clap:

PR #4151, opened 2024-02-11
Develop

VSkits:
- [DEMO LINK](https://VSkits.github.io/layout_stars/)
- [TEST REPORT LINK](https://VSkits.github.io/layout_stars/report/html_report/)


etojeDenys (changes requested):
.

volodymyr-soltys97 (approved):
Good job 👍 
Consider will be better if you remove the margin right for the last child
image

PR #4160, opened 2024-02-14
Develop

KrisPetrienko:
- [DEMO LINK](https://KrisPetrienko.github.io/layout_stars/)
- [TEST REPORT LINK](https://KrisPetrienko.github.io/layout_stars/report/html_report/)

+ Yellow stars are added with container modifier + pseudo-selector (NO extra classes)
+ Each BEM block has its own separate file
+ All from lesson theory are checked. Typical MistakesBEM
+ Code follows all the [Code Style Rules ❗️](https://github.com/KrisPetrienko/layout_stars/blob/master/checklist.md)

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -21,8 +21,8 @@ Follow
 
 ❗️ Replace `<your_account>` with your Github username and copy the links to `Pull Request` description:
 
-- [DEMO LINK](https://<your_account>.github.io/layout_stars/)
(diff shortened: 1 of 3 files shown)


maxim2310 (changes requested):
Great job overall! Just one tiny thing needs to be fixed before approval)
last star in every row must be without margin-right

vadiimvooo (approved):
LGTM

PR #4178, opened 2024-02-19
add task solution

D-Kondratiev:
- [DEMO LINK](https://D-Kondratiev.github.io/layout_stars/)
- [TEST REPORT LINK](https://D-Kondratiev.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -1,13 +1,65 @@
 <!doctype html>
 <html lang="en">
+
   <head>
     <meta charset="UTF-8">
     <meta name="viewport"
-          content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0">
+      content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0">
     <title>Stars</title>
     <link rel="stylesheet" href="./style.css">
   </head>
+
   <body>
-    <h1>Stars</h1>
+
+    <div class="stars stars--0">
+      <a href="#" class="stars__star"></a>
+      <a href="#" class="stars__star"></a>
+      <a href="#" class="stars__star"></a>
+      <a href="#" class="stars__star"></a>
+      <a href="#" class="stars__star"></a>
+    </div>
+
+    <div class="stars stars--1">
+      <a href="#" class="stars__star"></a>
+      <a href="#" class="stars__star"></a>
+      <a href="#" class="stars__star"></a>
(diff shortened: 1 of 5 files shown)


src/index.html
```diff
@@ -1,13 +1,65 @@
...
+      content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0">
     <title>Stars</title>
     <link rel="stylesheet" href="./style.css">
   </head>
+
   <body>
-    <h1>Stars</h1>
+
+    <div class="stars stars--0">
+      <a href="#" class="stars__star"></a>
```
polosanya, line 15: Actually, it shouldn't be a link, as this block should just show the amount of stars. They are not clickable

src/index.html
```diff
@@ -1,13 +1,65 @@
...
+      <a href="#" class="stars__star"></a>
+      <a href="#" class="stars__star"></a>
+      <a href="#" class="stars__star"></a>
+      <a href="#" class="stars__star"></a>
+    </div>
+
   </body>
+
 </html>
```
polosanya, line 64:
```suggestion
    </div>
  </body>
</html>
```

src/style.css
```diff
@@ -1 +1,64 @@
...
+
+.stars {
+  display: flex;
+}
+
+.stars__star {
+  background-image: url(images/star.svg);
+  width: 16px;
+  height: 16px;
+  margin-right: 4px;
```
polosanya, line 15: Make sure that the last star in the row doesn't have it

src/style.css
```diff
@@ -1 +1,64 @@
...
+
+}
+
+.stars--1 :nth-child(-n + 1) {
+  background-image: url(images/star-active.svg);
+  width: 16px;
+  height: 16px;
+  margin-right: 4px;
+  background-repeat: no-repeat;
+  background-position: center;
```
polosanya, line 27: You don't need to duplicate all styles. Just change the property that should be changed.
Also, you can combine the selectors with coma

polosanya (changes requested):
Nice! Let's improve your code a bit)

undeletable (approved):
The solution looks fine, so I'm approving it.

The only thing I suggest you to pay attention to:
According to the checklist, each BEM block should have its own separate CSS file. Here, styles for both `body` tag and `stars` block are placed in the single file.

In this case it's not really critical, as there aren't many styles. But for application of larger scale, separation is important. So be sure to follow all the items in checklist.

PR #4185, opened 2024-02-21
fixed bugs

Levik-24:
- [DEMO LINK](https://Levik-24.github.io/layout_stars/)
- [TEST REPORT LINK](https://Levik-24.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,1 +1,71 @@
-/* add styles here */
+body {
+  margin: 0;
+}
+
+.stars {
+  display: flex;
+}
+
+.stars__star {
+  width: 16px;
+  height: 16px;
+  background-repeat: no-repeat;
+  margin-right: 4px;
+  background-position: center;
+}
+
+.stars--0 .stars__star {
+  background-image: url(/src/images/star.svg);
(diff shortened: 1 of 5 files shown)


src/style.css
```diff
@@ -1 +1,71 @@
...
+
+.stars--2 .stars__star:nth-child(1),
+.stars--2 .stars__star:nth-child(2) {
+  background-image: url(/src/images/star-active.svg);
+}
+
+.stars--2 .stars__star:nth-child(3),
+.stars--2 .stars__star:nth-child(4),
+.stars--2 .stars__star:nth-child(5) {
+  background-image: url(/src/images/star.svg);
```
etojeDenys, line 46: it's too complicated, you can just use something like

`.stars--2 .stars__star:nth-child(-n + 2),`

etojeDenys (changes requested):

src/style.css
Levik-24: fixed

GUSILLUS (approved):
Good job! You can move to next task)

PR #4187, opened 2024-02-22
add stars_layout

metrakonon:
- [DEMO LINK](https://metrakonon.github.io/layout_stars/)
- [TEST REPORT LINK](https://metrakonon.github.io/layout_stars/report/html_report/)

- [X] Code follows all the [Code Style Rules ❗️]

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,1 +1,37 @@
-/* add styles here */
+.page__body {
+  margin: 0;
+}
+
+.stars {
+  display: flex;
+}
+
+.stars__star {
+  height: 16px;
+  width: 16px;
+  margin-right: 4px;
+
+  background-image: url(/src/images/star.svg);
+  background-position: center;
+  background-repeat: no-repeat;
+}
+
+.stars.stars--1 :nth-child(-n + 1) {
+  background-image: url(/src/images/star-active.svg);
+}
+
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -1 +1,37 @@
...
+}
+
+.stars {
+  display: flex;
+}
+
+.stars__star {
+  height: 16px;
+  width: 16px;
+  margin-right: 4px;
```
polosanya, line 12: Make sure that the last star in the row doesn't have it

src/style.css
```diff
@@ -1 +1,37 @@
...
+  background-image: url(/src/images/star.svg);
+  background-position: center;
+  background-repeat: no-repeat;
+}
+
+.stars.stars--1 :nth-child(-n + 1) {
+  background-image: url(/src/images/star-active.svg);
+}
+
+.stars.stars--2 :nth-child(-n + 2) {
```
polosanya, line 23: Let's combine the selectors with same styles to avoid duplicating yourself

polosanya (changes requested):
Nice! Let's improve your code a bit)

src/style.css
metrakonon: all fixed )

DanilWeda (approved):
Great! 
Approved.
Thanks!

PR #4200, opened 2024-02-25
Develop 'task completed'

denyssheremeta:
[DEMO LINK](https://denyssheremeta.github.io/layout_stars/)
[TEST REPORT LINK](https://denyssheremeta.github.io/layout_stars/report/html_report/)

[x] Code follows all the [Code Style Rules ❗️](https://github.com/denyssheremeta/layout_stars/blob/develop/checklist.md)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,1 +1,36 @@
-/* add styles here */
+body {
+  margin: 0;
+}
+
+.stars {
+  display: flex;
+}
+
+.stars__star {
+  height: 16px;
+  width: 16px;
+  margin-right: 4px;
+  background-image: url(./images/star.svg);
+  background-position: center;
+  background-repeat: no-repeat;
+}
+
+.stars--1 :nth-child(-n + 1) {
+  background-image: url(./images/star-active.svg);
+}
+
+.stars--2 :nth-child(-n + 2) {
+  background-image: url(./images/star-active.svg);
+}
+
+.stars--3 :nth-child(-n + 3) {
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -1 +1,36 @@
...
+}
+
+.stars {
+  display: flex;
+}
+
+.stars__star {
+  height: 16px;
+  width: 16px;
+  margin-right: 4px;
```
polosanya, line 12: Make sure that the last element of the row doesn't have margin-right

src/style.css
```diff
@@ -1 +1,36 @@
...
+  background-position: center;
+  background-repeat: no-repeat;
+}
+
+.stars--1 :nth-child(-n + 1) {
+  background-image: url(./images/star-active.svg);
+}
+
+.stars--2 :nth-child(-n + 2) {
+  background-image: url(./images/star-active.svg);
```
polosanya, line 23: Let's combine the selectors with common styles

polosanya (changes requested):
Almost done!

undeletable (approved):
The solution looks fine, so I'm approving it.

The only thing I suggest you to pay attention to:
According to the checklist, each BEM block should have its own separate CSS file. Here, styles for both `body` tag and `stars` block are placed in the single file.

In this case it's not really critical, as there aren't many styles. But for application of larger scale, separation is important. So be sure to follow all the items in checklist.

PR #4204, opened 2024-02-26
Develop

JohnyJDX:
- [DEMO LINK](https://JohnyJDX.github.io/layout_stars/)
- [TEST REPORT LINK](https://JohnyJDX.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -13,6 +13,47 @@
     />
   </head>
   <body>
-    <h1>Stars</h1>
+    <div class="stars stars--0">
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+    </div>
+    <div class="stars stars--1">
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+    </div>
+    <div class="stars stars--2">
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+    </div>
+    <div class="stars stars--3">
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+    </div>
+    <div class="stars stars--4">
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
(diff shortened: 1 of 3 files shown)


src/index.html
```diff
@@ -13,6 +13,47 @@
     />
   </head>
   <body>
-    <h1>Stars</h1>
+    <div class="stars stars--0">
```
undeletable, line 16: Please check how class name for block with modifier should look like according to BEM. Here, it's not used as expected for all the `stars` blocks.

src/style.css
```diff
@@ -1 +1,39 @@
-/* add styles here */
+body {
```
undeletable, line 1: According to the checklist, each block should have its own CSS file. Here, all the styles are placed in the single file.

src/style.css
```diff
@@ -1 +1,39 @@
...
+.stars__star {
+  background-image: url(images/star.svg);
+  background-position: center;
+  background-repeat: no-repeat;
+  width: 16px;
+  height: 16px;
+}
+
+.stars__star:not(:last-child) {
+  margin-right: 4px;
```
undeletable, line 18: Not a critical note: this style works fine, but it can be even simpler if you use `gap` property instead.

src/index.html
```diff
@@ -13,6 +13,47 @@
...
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+    </div>
+    <div class="stars stars--1">
```
undeletable, line 23: Please add empty lines between multiline sibling blocks of HTML.

src/style.css
```diff
@@ -1 +1,39 @@
...
+}
+
+.stars--1 :nth-child(-n + 1) {
```
undeletable, line 21: The style for this selector is the same as the following ones. They can be grouped into a single style block.

undeletable (changes requested):
The usage of `stars--{N}` modifiers for `stars` blocks is still not correct. See examples of modifiers usage at https://en.bem.info/methodology/quick-start/#modifier.
Other than that, everything's fine.

JohnyJDX:
> 
in theory to the topic on the MA website
image
but on en.bem.info has different information
image
I don't understand which one to use

undeletable:
@JohnyJDX Thank you for pointing me at this info. I'm sorry for the confusion. I'll sort this out with the team and come back to you.

undeletable (approved):

undeletable:
@JohnyJDX 
According to the guideline used for the studies:
> Follow naming convention block__element--modifier

```
<div class="block-name block-name--modifier-name--modifier-value">
  <p class="block-name__element-name block-name__element-name--modifier-name">
    text
  </p>
</div>
```
So the syntax you used is correct. Thank you for pointing me and sorry for confusion.
Overall, well done!

PR #4208, opened 2024-02-27
Stars

HigorPedroso:
[DEMO LINK](https://HigorPedroso.github.io/layout_stars/)
[TEST REPORT LINK](https://HigorPedroso.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -25,3 +25,4 @@ body {
 .stars--5 .stars__star:nth-child(-n + 5) {
   background-image: url(images/star-active.svg);
 }
+
--- a/src/style.css
+++ b/src/style.css
@@ -1,4 +1,3 @@
-/* add styles here */
 body {
   margin: 0;
 }
--- a/src/style.css
+++ b/src/style.css
@@ -9,7 +9,7 @@ body {
 .stars__star {
   width: 16px;
   height: 16px;
-  background-image: url('images/star-active.svg');
+  background-image: url('images/star.svg');
   background-repeat: no-repeat;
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -1 +1,39 @@
...
+  background-image: url('images/star.svg');
+}
+
+.stars--3 .stars__star:nth-child(n + 4) {
+  background-image: url('images/star.svg');
+}
+
+.stars--4 .stars__star:nth-child(n + 5) {
+  background-image: url('images/star.svg');
+}
```
EdPirro, line 39: This is a great implementation, but as a suggestion, the default behavior of stars should be inactive and adding the --N modifiers should activate the N first children of it, can you think of a way to invert your solution to follow this logic?

This is an important change (though not requested by the exercise, this is actually more of a personal suggestion) to improve your overall code maintainability and readability.
In this way it is way more complicated to understand what exactly the modifier does, but most importantly, imagine you now has to implement this logic for 10 stars; with this approach you'll have to change every single one of the classes to make sure it works as intended, the opposite approach would be able to reuse all the previous defined classes and simply add the remaining ones;

This is already a very long comment, if you have any questions please ask them in the qna chat

EdPirro (changes requested):
Great job, @HigorPedroso! I have only a small suggestion that I will actually request you to go through for I think you should be able to easily implement it, and it would be great for you to develop since early the mentality of creating as reusable and easy to maintain code as possible!

Other than that your solution solves the requirements perfectly 😊

PR #4211, opened 2024-02-28
add task solution

MarchAlbion:
- [DEMO LINK](https://MarchAlbion.github.io/layout_stars/)
- [TEST REPORT LINK](https://MarchAlbion.github.io/layout_stars/report/html_report/)

- [X] Code follows all the [Code Style Rules ❗️](./checklist.md)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -9,10 +9,51 @@
     <title>Stars</title>
     <link
       rel="stylesheet"
-      href="./style.css"
+      href="./styles/style.css"
     />
   </head>
   <body>
-    <h1>Stars</h1>
+    <div class="stars stars--0">
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+    </div>
+    <div class="stars stars--1">
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+    </div>
+    <div class="stars stars--2">
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+    </div>
+    <div class="stars stars--3">
+      <div class="stars__star"></div>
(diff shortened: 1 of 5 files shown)


src/index.html
```diff
@@ -9,10 +9,51 @@
     <title>Stars</title>
     <link
       rel="stylesheet"
-      href="./style.css"
+      href="./styles/style.css"
     />
   </head>
   <body>
-    <h1>Stars</h1>
+    <div class="stars stars--0">
```
undeletable, line 16: Please check how class name for block with modifier should look like according to BEM. Here, it's not used as expected for all the `stars` blocks.

src/index.html
```diff
@@ -9,10 +9,51 @@
...
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+    </div>
+    <div class="stars stars--1">
```
undeletable, line 23: Please add empty lines between multiline sibling blocks of HTML.

src/styles/block/stars.css
```diff
@@ -0,0 +1,32 @@
...
+.stars__star {
+  background-image: url(../../images/star.svg);
+  background-repeat: no-repeat;
+  background-position: center;
+  width: 16px;
+  height: 16px;
+  margin-right: 4px;
+}
+
+.stars--1 > .stars__star:nth-child(1) {
```
undeletable, line 14: The style for this selector is the same as the following ones. They can be grouped into a single style block.

undeletable (changes requested):
The result looks fine. But please check the usage of classnames according to BEM and HTML code guidelines.

src/index.html
undeletable: UPD. Please ignore this comment. I was referring to the guideline different from the one used for the studies. I'm sorry for the confusion.

undeletable: @JohnyJDX 
According to the guideline used for the studies:
> Follow naming convention block__element--modifier

```
<div class="block-name block-name--modifier-name--modifier-value">
  <p class="block-name__element-name block-name__element-name--modifier-name">
    text
  </p>
</div>
```
So the syntax you used is correct. Please ignore the first comment of this thread. Sorry for confusion.

tiserett (approved):

PR #4234, opened 2024-03-05
add task solution

lmmaarriiaa:
- [DEMO LINK](https://lmmaarriiaa.github.io/layout_stars/)
- [TEST REPORT LINK](https://lmmaarriiaa.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -13,6 +13,78 @@
     />
   </head>
   <body>
-    <h1>Stars</h1>
+    <container class="container">
+      <div class="stars stars--0">
+        <div class="stars__star"></div>
+
+        <div class="stars__star"></div>
+
+        <div class="stars__star"></div>
+
+        <div class="stars__star"></div>
+
+        <div class="stars__star"></div>
+      </div>
+
+      <div class="stars stars--1">
+        <div class="stars__star"></div>
+
+        <div class="stars__star"></div>
+
+        <div class="stars__star"></div>
+
+        <div class="stars__star"></div>
+
+        <div class="stars__star"></div>
+      </div>
+
+      <div class="stars stars--2">
+        <div class="stars__star"></div>
+
+        <div class="stars__star"></div>
+
+        <div class="stars__star"></div>
+
+        <div class="stars__star"></div>
+
+        <div class="stars__star"></div>
+      </div>
+
+      <div class="stars stars--3">
+        <div class="stars__star"></div>
+
+        <div class="stars__star"></div>
+
+        <div class="stars__star"></div>
+
+        <div class="stars__star"></div>
+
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -1 +1,39 @@
-/* add styles here */
+html {
+  --star: url(/src/images/star.svg);
```
mshepel, line 2: Using CSS variables - nice!

src/index.html
```diff
@@ -13,6 +13,78 @@
     />
   </head>
   <body>
-    <h1>Stars</h1>
+    <container class="container">
+      <div class="stars stars--0">
+        <div class="stars__star"></div>
+
```
mshepel, line 19: We don't need empty lines between `stars__star` elements. According to the guideline, they should be added between multiline siblings - and `stars__star` are single-line.

undeletable, line 19: We don't need empty lines between stars__star elements. According to the guideline, they should be added between multiline siblings - and stars__star are single-line.

src/style.css
undeletable, line 2: Using CSS variables - nice!

src/style.css
```diff
@@ -1 +1,39 @@
...
+  --star-activated: url(/src/images/star-active.svg);
+}
+
+body {
+  margin: 0;
+}
+
+.container {
```
undeletable, line 10: According to the task conditions, each BEM block should have its separate CSS file. You have two BEM blocks defined, but place all the styles in the single file.

src/index.html
```diff
@@ -13,6 +13,78 @@
     />
   </head>
   <body>
-    <h1>Stars</h1>
+    <container class="container">
```
undeletable, line 16: `container` is not a valid HTML tag. Please use a different tag.

src/style.css
```diff
@@ -1 +1,39 @@
...
+  height: 16px;
+  width: 16px;
+  margin-left: 4px;
+}
+
+.stars__star:first-child {
+  margin: 0;
+}
+
+.stars--1 .stars__star:nth-child(-n + 1),
```
undeletable, line 33: According to the conditions, yellow stars should be added using container modifier + pseudo-selector only, without extra classes. So `.stars__star` should not be a part of these selectors.

undeletable (changes requested):
The result is nice, but some changes are needed for compliance with checklist and guidelines.

lmmaarriiaa:
> The result is nice, but some changes are needed for compliance with checklist and guidelines.

Thank you for your feedback! I`ll try to fix everything

mrkovaliv (approved):
Nice work 👍

PR #4242, opened 2024-03-06
add task solution

Follder:


undeletable:
Please perform the steps mentioned in [the checklist for this task](https://github.com/mate-academy/layout_stars?tab=readme-ov-file#checklist).

src/style.css
```diff
@@ -1 +1,32 @@
...
+  padding: 22px;
+  display: flex;
+  flex-direction: column;
+  justify-content: space-between;
+
+  height: 116px;
+  width: 96px;
+}
+
+.stars {
```
undeletable, line 12: According to the task conditions, each BEM block should have its separate CSS file. In your solution all the styles are placed in the single file.

src/style.css
```diff
@@ -1 +1,32 @@
-/* add styles here */
+body {
+  margin: 0;
+  padding: 22px;
```
undeletable, line 3: It's not a good practice to use add such styles for body element. It's a common container for the whole web page, so the only styles which make sense for body are: the ones for resetting browser's default styles for body (`margin: 0` in this case), and some general styles applied to all elements on the page (like font size, text color etc.).
In you case it's better to wrap your `stars` blocks in another container block, and define styles for this block

src/style.css
```diff
@@ -1 +1,32 @@
...
+  display: flex;
+  flex-direction: column;
+  justify-content: space-between;
+
+  height: 116px;
```
undeletable, line 8: All 7 tests fail:
stars-failure
Looks like that's because the distance between rows and the padding is different from expected design. I assume that incorrect padding and height values are causing this. Please take a look at that, most likely these properties can be removed at all.

undeletable (changes requested):
Basically the solution looks good, but the appearance is slightly different from the expected result, which most likely causes tests failure. Also please pay attention to checklist and guidelines.

Follder:
- [DEMO LINK](https://follder.github.io/layout_stars/)
- [TEST REPORT LINK](https://follder.github.io/layout_stars/report/html_report/)

I don`t understand why the tests are failed. I did the same as in the figma`s layout.
I can`t see what is not good, because the reference in the test report has unexpected errror
image

undeletable:
@Follder This looks like an intermittent issue with test run. I've launched tests for your solution on my PC, and here's the result:
start-4242-failure
See the different vertical distance between rows of stars in reference view and your solution.
After you push the changes, the tests would be re-launched, so you should see their result. If after that you still see that 'Unexpected error', please let us know in the chat on platform.

undeletable:
@Follder By the way, for self-check you can run tests on your PC before pushing the changes - using `npm run test:only` command.

Follder:
fixed all errors

- [DEMO LINK](https://follder.github.io/layout_stars/)
- [TEST REPORT LINK](https://follder.github.io/layout_stars/report/html_report/)

PR #4247, opened 2024-03-06
add task solution

lesia-grapeniuk:
[DEMO LINK](https://lesia-grapeniuk.github.io/layout_stars/)
- [TEST REPORT LINK](https://lesia-grapeniuk.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,1 +1,30 @@
-/* add styles here */
+body {
+  margin: 0;
+}
+
+.stars {
+  display: flex;
+}
+
+.stars__star {
+  background-image: url(images/star-active.svg);
+  background-repeat: no-repeat;
+  background-position: center;
+
+  width: 16px;
+  height: 16px;
+  margin-right: 4px;
+}
+
+.stars__star:first-child {
+  margin-left: 0;
+}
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -1 +1,30 @@
...
+  background-position: center;
+
+  width: 16px;
+  height: 16px;
+  margin-right: 4px;
+}
+
+.stars__star:first-child {
+  margin-left: 0;
+}
```
Esceype, line 21:
```suggestion
.stars__star:last-child {
  margin-right: 0;
}
```
You should disable margin-right for last child because you exactly this property for each star

Esceype (changes requested):
GJ!
Read the following comment and make redeploy because your test report demo link doesn`t work.

src/style.css
```diff
@@ -1 +1,34 @@
...
```
Esceype, line 21:
```suggestion
```
And now it should be deleted, you don`t need it

Esceype (changes requested):
Your demo link still doesn't work and read please read this one comment:

PR #4282, opened 2024-03-12
add task solution

SlawikH:
- [DEMO LINK](https://SlawikH.github.io/layout_stars/)
- [TEST REPORT LINK](https://SlawikH.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -13,6 +13,52 @@
     />
   </head>
   <body>
-    <h1>Stars</h1>
+    <div class="stars stars--0">
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+    </div>
+
+    <div class="stars stars--1">
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+    </div>
+
+    <div class="stars stars--2">
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -1 +1,24 @@
...
+
+.stars {
+  display: flex;
+}
+
+.stars__star {
+  background-image: url('./images/star.svg');
+  background-repeat: no-repeat;
+  background-position: center;
+  margin-right: 4px;
```
polosanya, line 13: Make sure that the last element of the row doesn't have it

polosanya (changes requested):
Demo link is not correct(
Please make sure that you deployed your work correctly
https://mate-academy.github.io/layout_task-guideline/#codenpm-run-deploycode-to-publish-to-github-pages

src/index.html
```diff
@@ -13,6 +13,52 @@
...
+      <div class="stars__star"></div>
+    </div>
+
+    <div class="stars stars--5">
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+    </div>
```
DanilWeda, line 62: Try to use ul and li in this case.
it's preferred because we have one repeated element inside div.
In this case best variant use list)

src/style.css
```diff
@@ -1 +1,28 @@
...
+  width: 16px;
+  height: 16px;
+}
+
+.stars__star:last-child {
+  margin-right: 0;
+}
```
tiserett, line 20: you can set `margin-right: 4px;` here, and use `:not` with `:last-child` to prevent setting classes for the last element

tiserett (approved):
please, fix comments left by Danylo

src/style.css
SlawikH: got you, thanks!

PR #4283, opened 2024-03-12
add solution

stepanyukvova200:

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -9,10 +9,57 @@
     <title>Stars</title>
     <link
       rel="stylesheet"
-      href="./style.css"
+      href="styles/style.css"
     />
   </head>
+
   <body>
-    <h1>Stars</h1>
+    <div class="stars stars--0">
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+    </div>
+
+    <div class="stars stars--1">
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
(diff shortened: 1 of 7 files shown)


src/styles/blocks/stars-modifiers.css
```diff
@@ -0,0 +1,19 @@
...
+  background-image: url('../../images/star-active.svg');
+}
+
+.stars--4 :nth-child(-n + 4) {
+  background-image: url('../../images/star-active.svg');
+}
+
+.stars--5 :nth-child(-n + 5) {
+  background-image: url('../../images/star-active.svg');
+}
```
DanilWeda, line 19: You can write all selectors together, and use one line with a property  `background-image: url('../../images/star-active.svg'`

src/index.html
```diff
@@ -9,10 +9,57 @@
...
   </head>
+
   <body>
-    <h1>Stars</h1>
+    <div class="stars stars--0">
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
```
DanilWeda, line 22: Try to use `ul` and `li` in this case.
it's preferred because we have one repeated element inside div. 
In this case best variant use list)

DanilWeda (changes requested):
Good job! 
Please, check some improvements. 
Also, add a demo link in a readme file. 
Many thanks!

stepanyukvova200:
i hope that i need to put it here )
- [DEMO LINK](https://stepanyukvova200.github.io/layout_stars/)
- [TEST REPORT LINK](https://stepanyukvova200.github.io/layout_stars/report/html_report/)

PR #4290, opened 2024-03-13
add task solution

Pokhvalenkos:
- [DEMO LINK](https://Pokhvalenkos.github.io/layout_stars/)
- [TEST REPORT LINK](https://Pokhvalenkos.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,1 +1,26 @@
 /* add styles here */
+.stars {
+  display: flex;
+}
+
+.stars__star {
+  background-image: url('images/star.svg');
+  background-position: center;
+  background-repeat: no-repeat;
+  height: 16px;
+  width: 16px;
+  margin-right: 4px;
+}
+
+.stars--1 .stars__star:nth-child(-n + 1),
+.stars--2 .stars__star:nth-child(-n + 2),
+.stars--3 .stars__star:nth-child(-n + 3),
+.stars--4 .stars__star:nth-child(-n + 4),
+.stars--5 .stars__star:nth-child(-n + 5) {
+  background-image: url('./images/star-active.svg');
+}
+
+body {
+  margin: 0;
+  padding: 0;
+}
--- a/readme.md
+++ b/readme.md
@@ -21,8 +21,8 @@ Follow
 
 ❗️ Replace `<your_account>` with your Github username and copy the links to `Pull Request` description:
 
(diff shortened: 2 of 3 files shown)


src/style.css
```diff
@@ -1 +1,26 @@
...
+  display: flex;
+}
+
+.stars__star {
+  background-image: url('images/star.svg');
+  background-position: center;
+  background-repeat: no-repeat;
+  height: 16px;
+  width: 16px;
+  margin-right: 4px;
```
etojeDenys, line 12: remove the margin-right in the last star

etojeDenys (changes requested):
try to fix your demo links and re-request me

feel free to ask for some help with that in the chat

Pokhvalenkos:
> try to fix your demo links and re-request me
> 
> feel free to ask for some help with that in the chat

idk why, but tests link doesn't work, however u can compare link with demo, my nickname is correct there

src/style.css
Viktor-Kost: Isn't fixed
You should use `:not(:last-child)` to add margin to all items besides last one
```
.stars__star {
 .......
}

  margin-right: 4px;
}
```

src/style.css
```diff
@@ -1 +1,26 @@
...
+.stars--2 .stars__star:nth-child(-n + 2),
+.stars--3 .stars__star:nth-child(-n + 3),
+.stars--4 .stars__star:nth-child(-n + 4),
+.stars--5 .stars__star:nth-child(-n + 5) {
+  background-image: url('./images/star-active.svg');
+}
+
+body {
+  margin: 0;
+  padding: 0;
```
Viktor-Kost, line 25: body doesn't have default padding
```suggestion
```

src/style.css
```diff
@@ -1 +1,26 @@
 /* add styles here */
+.stars {
+  display: flex;
+}
+
+.stars__star:not(:last-child) {
```
Viktor-Kost, line 6: not here
```suggestion
.stars__star {
```

Viktor-Kost (changes requested):
Hi!)
After adding those fixes i've written at comments redeploy your work. If you still see 404 page feel free to ask at fe_chat. Probably the reason is test failure

Viktor-Kost (approved):
Nice!

PR #4307, opened 2024-03-17
add task solution

vorotilinas:
[x] All `Typical Mistakes` from `BEM` lesson theory are checked.
[x] Code follows all the [Code Style Rules ❗️](./checklist.md)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,1 +1,37 @@
 /* add styles here */
+body {
+  margin: 0;
+}
+
+.stars {
+  display: flex;
+}
+
+.stars__star {
+  height: 16px;
+  width: 16px;
+  background-image: url(images/star.svg);
+  background-position: center;
+  background-repeat: no-repeat;
+  margin-right: 4px;
+}
+
+.stars--1 .stars__star:nth-child(-n + 1) {
+  background-image: url(images/star-active.svg);
+}
+
+.stars--2 .stars__star:nth-child(-n + 2) {
+  background-image: url(images/star-active.svg);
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -1 +1,37 @@
...
+  margin-right: 4px;
+}
+
+.stars--1 .stars__star:nth-child(-n + 1) {
+  background-image: url(images/star-active.svg);
+}
+
+.stars--2 .stars__star:nth-child(-n + 2) {
+  background-image: url(images/star-active.svg);
+}
```
etojeDenys, line 25:
```suggestion
.stars--1 .stars__star:nth-child(-n + 1),
.stars--2 .stars__star:nth-child(-n + 2) {
  background-image: url(images/star-active.svg);
}
```
combine it all with a comma

src/style.css
```diff
@@ -1 +1,37 @@
...
+  display: flex;
+}
+
+.stars__star {
+  height: 16px;
+  width: 16px;
+  background-image: url(images/star.svg);
+  background-position: center;
+  background-repeat: no-repeat;
+  margin-right: 4px;
```
etojeDenys, line 16: remove the margin from the last star

etojeDenys (changes requested):

etojeDenys (approved):
good job

src/style.css
vorotilinas: .stars::after {
  content: '';
  margin-right: 0;
}

thx

PR #4318, opened 2024-03-21
Solution

Mykhalishere:
- [DEMO LINK](https://Mykhalishere.github.io/layout_stars/)
- [TEST REPORT LINK](https://Mykhalishere.github.io/layout_stars/report/html_report/)
- [X ] Yellow stars are added with container modifier + pseudo-selector (NO extra classes)
- [X ] Each BEM block has its own separate file
- [X ] All `Typical Mistakes` from `BEM` lesson theory are checked.
- [ (I think so but check please)] Code follows all the [Code Style Rules ❗️](./checklist.md)

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -1,11 +1,12 @@
 # Stars block
 
-Implement the [Stars Block](https://www.figma.com/file/ojkArVazq7vsX0nbpn9CxZ/Moyo-%2F-Catalog-(ENG)?node-id=11325%3A2960&mode=dev) used in a card and catalog.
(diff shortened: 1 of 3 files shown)


maxim2310 (changes requested):
the last star in every row must be without margin right 
image

maxim2310 (changes requested):
firstly you didnt deploy last version) and secondly EVERY last star in row must be without margin-right, but last time you tried to remove margin from last row
image

etojeDenys (approved):
lgtm

PR #4339, opened 2024-03-27
add task solution

Bogdankudrik:
- [DEMO LINK](https://Bogdankudrik.github.io/layout_stars/)
- [TEST REPORT LINK](https://Bogdankudrik.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -7,7 +7,6 @@ body {
   display: flex;
   align-items: center;
   justify-content: center;
-  width: auto;
 }
 
 .stars__star {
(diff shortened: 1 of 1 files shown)


src/style.css
```diff
@@ -1 +1,29 @@
...
+body {
+  margin: 0;
+  padding: 0;
+}
+
+.stars {
+  display: flex;
+  align-items: center;
+  justify-content: center;
+  width: auto;
```
nazarmatsevych, line 10: redundant width property, you can remove it

nazarmatsevych (changes requested):
Almost done, let's make some improvements:

Remove margin-right from the last stars in a row (use :last-child rule)
image

src/style.css
Bogdankudrik: if I type ( .stars__star:last-child {
   margin-right: 0;
}) then the tests do not pass.

Bogdankudrik: Sorry, all corrections)

anastasiiavorobiova (approved):
Awesome work!

PR #4344, opened 2024-03-28
added task solution

nshtykh:
- [x] Code follows all the [Code Style Rules ❗️](https://github.com/nshtykh/layout_stars/blob/master/checklist.md)


nazar-medushevskyi (approved):
Good job! Don't worry about adding an indent after the last line

PR #4360, opened 2024-04-03
add solution

Pjankusha:
[DEMO LINK]  (https://pjankusha.github.io/layout_stars/)
 [TEST REPORT LINK](https://pjankusha.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,1 +1,32 @@
 /* add styles here */
+body {
+  margin: 0;
+}
+
+.stars {
+  display: flex;
+}
+
+.stars__star {
+  background-image: url(/src/images/star.svg);
+  background-position: center;
+  background-repeat: no-repeat;
+  width: 16px;
+  height: 16px;
+}
+
+.stars__star:not(:first-child) {
+  margin-left: 4px;
+}
+
+.stars--1 :nth-child(1),
+.stars--2 :nth-child(-n + 2),
+.stars--3 :nth-child(-n + 3),
+.stars--4 :nth-child(-n + 4),
+.stars--5 :nth-child(-n + 5) {
+  background-image: url(/src/images/star-active.svg);
+  background-position: center;
+  background-repeat: no-repeat;
+  width: 16px;
+  height: 16px;
+}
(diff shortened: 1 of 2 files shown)


src/style.css
```diff
@@ -1 +1,32 @@
...
+.stars--1 :nth-child(1),
+.stars--2 :nth-child(-n + 2),
+.stars--3 :nth-child(-n + 3),
+.stars--4 :nth-child(-n + 4),
+.stars--5 :nth-child(-n + 5) {
+  background-image: url(/src/images/star-active.svg);
+  background-position: center;
+  background-repeat: no-repeat;
+  width: 16px;
+  height: 16px;
```
Viktor-Kost, line 31: no needs to repeat these styles since you've already added them to `.stars__star `

src/style.css
```diff
@@ -1 +1,32 @@
...
+
+.stars__star:not(:first-child) {
+  margin-left: 4px;
+}
+
+.stars--1 :nth-child(1),
+.stars--2 :nth-child(-n + 2),
+.stars--3 :nth-child(-n + 3),
+.stars--4 :nth-child(-n + 4),
+.stars--5 :nth-child(-n + 5) {
```
Viktor-Kost, line 26: it's better to specify where exactly you'd like to apply those styles
```suggestion
.stars--1 .stars__star:nth-child(1),
.stars--2 .stars__star:nth-child(-n + 2),
.stars--3 .stars__star:nth-child(-n + 3),
.stars--4 .stars__star:nth-child(-n + 4),
.stars--5 .stars__star:nth-child(-n + 5) {
```

Viktor-Kost (changes requested):
Almost perfect!)
Just some minor fixes need to be applied

src/style.css
Pjankusha: thank you for suggestion.

src/style.css
Pjankusha: thank you for suggestion.

Pjankusha:
made changes according mentor's  suggestions

PR #4379, opened 2024-04-11
task with the stasr

InnaPopenko:
- [DEMO LINK](https://InnaPopenko.github.io/layout_stars/)
- [TEST REPORT LINK](https://InnaPopenko.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,1 +1,28 @@
 /* add styles here */
+body {
+  margin: 0;
+}
+
+.stars {
+  display: flex;
+}
+
+.stars__star {
+  width: 16px;
+  height: 16px;
+  background-image: url(./images/star-active.svg);
+  display: inline-block;
+  justify-content: center;
+  background-position: center;
+  background-size: auto;
+  background-repeat: no-repeat;
+  margin-right: 4px;
+}
+
+.stars--0 :nth-child(n),
+.stars--1 :nth-last-child(-n + 4),
+.stars--2 :nth-last-child(-n + 3),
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -1 +1,28 @@
...
+.stars {
+  display: flex;
+}
+
+.stars__star {
+  width: 16px;
+  height: 16px;
+  background-image: url(./images/star-active.svg);
+  display: inline-block;
+  justify-content: center;
```
sTorba24, line 15: this property works only in flex-container

src/style.css
```diff
@@ -1 +1,28 @@
...
+  background-position: center;
+  background-size: auto;
```
sTorba24, line 17: can be removed

src/style.css
```diff
@@ -1 +1,28 @@
...
+
+.stars {
+  display: flex;
+}
+
+.stars__star {
+  width: 16px;
+  height: 16px;
+  background-image: url(./images/star-active.svg);
+  display: inline-block;
```
sTorba24, line 14: remove

sTorba24 (approved):
Good work!

src/style.css
InnaPopenko: ✅

src/style.css
InnaPopenko: ✅

src/style.css
InnaPopenko: ✅

PR #4409, opened 2024-04-22
Develop

ketinng:
- [DEMO LINK](https://ketinng.github.io/layout_stars/)
- [TEST REPORT LINK](https://ketinng.github.io/layout_stars/report/html_report/)

- [X] Code follows all the [Code Style Rules ❗️](./checklist.md)


src/index.html
```diff
@@ -12,7 +12,138 @@
...
     />
   </head>
-  <body>
-    <h1>Stars</h1>
+  <body class="page">
+    <div class="stars stars--0">
+      <a
+        href="#"
+        class="stars__star"
+      ></a>
```
GUSILLUS, line 20: Consider using `div` for the star element or `ul` and `li` combination

src/index.html
```diff
@@ -12,7 +12,138 @@
...
+      <a
+        href="#"
+        class="stars__star"
+      ></a>
+    </div>
+    <div class="stars stars--1">
```
GUSILLUS, line 38: add space between elements on one level of depths

GUSILLUS (changes requested):
GJ! But you need to use only two files 

1. style.css
2. stars.css

So you don't need to make file on each file

src/index.html
ketinng: I thought of using other elements, but a tag made sense to me because it could potentially be used to navigate straight to reviews

ketinng:
Please clarify if I can keep the extra stars folder or if it should be removed

src/index.html
```diff
@@ -12,7 +12,143 @@
...
+        class="stars__star"
+      ></a>
+      <a
+        href="#"
+        class="stars__star"
+      ></a>
+      <a
+        href="#"
+        class="stars__star"
+      ></a>
```
nazarmatsevych, line 51:
```suggestion
      <a href="#" class="stars__star"></a>
```
And you can use div instead of link

nazarmatsevych (changes requested):
A separate file for each star is redundant since these blocks are exactly the same. And you also need to use nth-child in a way that simplifies your code
image

if you have  any questions - feel free to ask them in fe_chat

ketinng:
I just wanted to write a comment explaining my decision regarding using a tag instead of div
I have had front end study experience at uni and it step school back in 2017 and we would always use div tags and while they were convenient, have everything packed in divs didn't look nice and impacted readability in a bad way, at least that's how I viewed it, that's why I tried to avoid using divs here while also after comments I have double-checked websites like rozetka.com.ua and star ratings there actually took me to reviews tab of an item being sold, most likely the a tag is applied for the whole block, I decided to take it a step further and imo using a tags for each separate star could also applied a filter to the review page that would show the reviews with the star rating that was selected

VitaliyBondarenko1982 (approved):
Looks good

PR #4443, opened 2024-04-30
add stars solution

FaiHamid:
- [DEMO LINK](https://FaiHamid.github.io/layout_stars/)
- [TEST REPORT LINK](https://FaiHamid.github.io/layout_stars/report/html_report/)

[x] All `Typical Mistakes` from `BEM` lesson theory are checked.
[x] Code follows all the [Code Style Rules ❗️](./checklist.md)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,1 +1,0 @@
-/* add styles here */
--- a/readme.md
+++ b/readme.md
@@ -21,8 +21,8 @@ Follow
 
 ❗️ Replace `<your_account>` with your Github username and copy the links to `Pull Request` description:
 
-- [DEMO LINK](https://<your_account>.github.io/layout_stars/)
-- [TEST REPORT LINK](https://<your_account>.github.io/layout_stars/report/html_report/)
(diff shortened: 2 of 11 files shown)


src/style.css
```diff
@@ -1 +0,0 @@
-/* add styles here */
```
roman-mirzoian, line 1: Styles should be here in one file.

roman-mirzoian (changes requested):
Be careful not to use too many different style files. In this task, you can put everything in one file using group selectors. This will combine multiple selectors with common styles.

src/style.css
FaiHamid: Good morning! One of checklist point was "Each BEM block has its own separate file". Should I move styles from rest.css to style.css ? If yes, where import BEM block files?

roman-mirzoian: @FaiHamid I wrote a clarification to your question in response to the QnA chat

roman-mirzoian (approved):
Well done!

PR #4444, opened 2024-04-30
completed all points of the task

StsKushnir:
- [DEMO LINK](https://StsKushnir.github.io/layout_stars/)
- [TEST REPORT LINK](https://StsKushnir.github.io/layout_stars/report/html_report/)

[x]Each BEM block has its own separate file
[x] Code follows all the [Code Style Rules ❗️]

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -21,8 +21,8 @@ Follow
 
 ❗️ Replace `<your_account>` with your Github username and copy the links to `Pull Request` description:
 
-- [DEMO LINK](https://<your_account>.github.io/layout_stars/)
-- [TEST REPORT LINK](https://<your_account>.github.io/layout_stars/report/html_report/)
+- [DEMO LINK](https://StsKushnir.github.io/layout_stars/)
+- [TEST REPORT LINK](https://StsKushnir.github.io/layout_stars/report/html_report/)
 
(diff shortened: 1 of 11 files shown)


roman-mirzoian (changes requested):
Be careful not to use too many different style files. In this task, you can put everything in one file using group selectors. This will combine multiple selectors with common styles.

Classes of stars have the same properties, so it is better to use a combination of selectors.
image

src/styles/stars.css
```diff
@@ -0,0 +1,19 @@
...
+
+/* блок неактивних зірок */
+
+.stars--0 > *,
+.stars--1 :nth-child(n + 2),
+.stars--2 :nth-child(n + 3),
+.stars--3 :nth-child(n + 4),
+.stars--4 :nth-child(n + 5) {
+  background-image: url(../images/star.svg);
+}
```
roman-mirzoian, line 19: No need to overcode, without imports in a separate css file, all of this can be contained in one css file it is better to immediately set the background to the class stars__star - star.svg and then change them to the active one.

roman-mirzoian (changes requested):
Please pay attention to the comment.

roman-mirzoian (approved):
Well done!

PR #4447, opened 2024-05-01
add tasl solution

V-Shut:
- [DEMO LINK](https://v-shut.github.io/layout_stars/)
- [TEST REPORT LINK](https://v-shut.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,1 +1,43 @@
-/* add styles here */
+html {
+  margin: 0;
+  padding: 0;
+
+  display: flex;
+  justify-content: center;
+  align-items: center;
+}
+
+body {
+  margin: 0;
+  width: 100px;
+  height: 100px;
+
+  display: flex;
+  flex-direction: column;
+}
+
+.stars {
+  width: 100%;
+  height: auto;
+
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -1 +1,43 @@
-/* add styles here */
+html {
+  margin: 0;
+  padding: 0;
+
+  display: flex;
+  justify-content: center;
+  align-items: center;
```
roman-mirzoian, line 7: There is no need to set these properties globally, a better approach is to apply flexbox to specific containers or components.

src/style.css
```diff
@@ -1 +1,43 @@
...
+}
+
+body {
+  margin: 0;
+  width: 100px;
+  height: 100px;
+
+  display: flex;
+  flex-direction: column;
```
roman-mirzoian, line 16: Please pay attention to the previous comment.

roman-mirzoian (changes requested):
Pay attention to this comments.

roman-mirzoian (approved):
Well done!

PR #4455, opened 2024-05-03
Develop

Yaroslavlesh:
- [DEMO LINK](https://Yaroslavlesh.github.io/layout_stars/)
- [TEST REPORT LINK](https://Yaroslavlesh.github.io/layout_stars/report/html_report/)


src/styles/stars.css
```diff
@@ -0,0 +1,37 @@
...
+.stars--4 > .stars__star:nth-child(-n + 2),
+.stars--4 > .stars__star:nth-child(-n + 3),
+.stars--4 > .stars__star:nth-child(-n + 4),
+.stars--5 > .stars__star:nth-child(-n + 1),
+.stars--5 > .stars__star:nth-child(-n + 2),
+.stars--5 > .stars__star:nth-child(-n + 3),
+.stars--5 > .stars__star:nth-child(-n + 4),
+.stars--5 > .stars__star:nth-child(-n + 5) {
+  background-image: url(../images/star-active.svg);
+}
```
dinaminahi, line 35: You don't have to write a selector for each star
`.stars--4 > .stars__star:nth-child(-n + 4)` means it is a selector for first four elements with such class
`.stars--3 > .stars__star:nth-child(-n + 3),`  means it is a selector for first three elements with such class

so you just need to write one selector per each row of stars and in the end you will receive only 5 strings of selectors

dinaminahi (changes requested):
Good job 👍 But lets approve it a bit

dinaminahi (approved):
Well done ⭐

PR #4459, opened 2024-05-05
add task solution

Maryan-Yashchenko:
- [x] Code follows all the [Code Style Rules ❗️](https://github.com/Maryan-Yashchenko/layout_stars/blob/master/checklist.md)


etojeDenys (approved):
lgtm

PR #4464, opened 2024-05-06
add task solution

wintsistrs:
- [DEMO LINK](https://wintsistrs.github.io/layout_stars/)
- [TEST REPORT LINK](https://wintsistrs.github.io/layout_stars/report/html_report/)
- [x] Code follows all the [Code Style Rules ❗️](https://github.com/wintsistrs/layout_stars/blob/master/checklist.md)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -22,77 +22,10 @@ body {
   margin-right: 0;
 }
 
-.stars--1 :nth-child(1) {
+.stars--1 .stars__star:nth-child(1),
+.stars--2 .stars__star:nth-child(-n + 2),
+.stars--3 .stars__star:nth-child(-n + 3),
+.stars--4 .stars__star:nth-child(-n + 4),
+.stars--5 .stars__star:nth-child(-n + 5) {
   background-image: url(images/star-active.svg);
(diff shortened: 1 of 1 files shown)


src/style.css
```diff
@@ -1 +1,98 @@
...
+  background-position: center;
+}
+
+.stars__star:last-child {
+  margin-right: 0;
+}
+
+.stars--1 :nth-child(1) {
+  background-image: url(images/star-active.svg);
+  background-repeat: no-repeat;
```
etojeDenys, line 27: it's redundant 'cause you specified that property above 
image
```suggestion

```

src/style.css
```diff
@@ -1 +1,98 @@
...
+}
+
+.stars--2 :nth-child(1) {
+  background-image: url(images/star-active.svg);
+  background-repeat: no-repeat;
+}
```
etojeDenys, line 33:
```
```
we can simplify it using the comma and `n` in `nth-child`

etojeDenys (changes requested):

etojeDenys (approved):
lgtm

PR #4495, opened 2024-05-24
add task solution

Fetrw:
- [DEMO LINK](https://Fetrw.github.io/layout_stars/)
- [TEST REPORT LINK](https://Fetrw.github.io/layout_stars/report/html_report/)


src/index.html
```diff
@@ -4,15 +4,82 @@
...
+      href="styles/stars--3.css"
+    />
+    <link
+      rel="stylesheet"
+      href="styles/stars--4.css"
+    />
+    <link
+      rel="stylesheet"
+      href="styles/stars--5.css"
     />
```
vadiimvooo, line 34: You don't need all of this files. Just define all styles in one file style.css

vadiimvooo (changes requested):
Good job, but there is one thing to fix

vadiimvooo (approved):
Well done

PR #4513, opened 2024-05-29
add task solution

VictorKomara:
- [x] Code follows all the [Code Style Rules ❗️]


VitaliyBondarenko1982 (approved):
Looks good. 
Do not forget provide demo links to pr description in next tasks.

PR #4532, opened 2024-06-09
made stars

kathetem:
[DEMO LINK](https://kathetem.github.io/layout_stars/)
[TEST REPORT LINK](https://kathetem.github.io/layout_stars/report/html_report/)

- [x] Code follows all the [Code Style Rules ❗️](https://github.com/kathetem/layout_stars/blob/master/checklist.md)Copy this

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,1 +1,74 @@
-/* add styles here */
+body {
+  margin: 0;
+}
+
+.stars {
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  width: auto;
+}
+
+.stars__star {
+  height: 16px;
+  width: 16px;
+  margin-right: 4px;
+  background-image: url(images/star.svg);
+  background-repeat: no-repeat;
+  background-position: center;
+}
+
+.stars--1 :nth-child(1) {
+  background-image: url(images/star-active.svg);
+  background-repeat: no-repeat;
+}
+
+.stars--2 :nth-child(1) {
+  background-image: url(images/star-active.svg);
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -1 +1,74 @@
...
+
+.stars--4 :nth-child(2) {
+  background-image: url(images/star-active.svg);
+  background-repeat: no-repeat;
+}
+
+.stars--4 :nth-child(3) {
+  background-image: url(images/star-active.svg);
+  background-repeat: no-repeat;
+}
```
anastasiiavorobiova, line 64:
```suggestion

```
Consider grouping selectors with commas. No need to repeat the `background-repeat: no-repeat;` property. It has been declared in the basic class. Also, consider refactoring the code a bit and using `-n` https://developer.mozilla.org/en-US/docs/Web/CSS/:nth-child

anastasiiavorobiova (changes requested):
Great work! Almost done!

src/style.css
```diff
@@ -1 +1,33 @@
...
+
+.stars--2 :nth-child(-n + 2),
+.stars--3 :nth-child(-n + 3),
+.stars--4 :nth-child(-n + 4) {
+  background-image: url(images/star-active.svg);
+}
+
+.stars--5 > .stars__star {
+  background-image: url(images/star-active.svg);
+}
```
etojeDenys, line 33: combine all these selectors using the ','

etojeDenys (changes requested):

etojeDenys (approved):
good job

PR #4539, opened 2024-06-11
Add stars block

Toltbcm:
- [DEMO LINK](https://toltbcm.github.io/layout_stars/)
- [TEST REPORT LINK](https://toltbcm.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -13,6 +13,52 @@
     />
   </head>
   <body>
-    <h1>Stars</h1>
+    <div class="block stars--0">
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+    </div>
+
+    <div class="block stars--1">
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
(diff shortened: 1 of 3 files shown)


src/index.html
```diff
@@ -13,6 +13,52 @@
     />
   </head>
   <body>
-    <h1>Stars</h1>
+    <div class="block stars--0">
```
BudnikOleksii, line 16:
```suggestion
    <div class="stars stars--0">
```
It should be the block class first and then the modifier. You can't use a modifier without a block. Fix in all places

src/style.css
```diff
@@ -1 +1,28 @@
-/* add styles here */
+body {
+  margin: 0;
+}
+
+.block {
```
BudnikOleksii, line 5:
```suggestion
.stars {
```

BudnikOleksii (changes requested):
Good job, just one moment regarding BEM should be fixed

src/index.html
Toltbcm: fixed

src/style.css
Toltbcm: fixed

BudnikOleksii (approved):
👍

PR #4553, opened 2024-06-15
add ls task solution

vadimurbanovich:
- [DEMO LINK](https://vadimurbanovich.github.io/layout_stars/)
- [TEST REPORT LINK](https://vadimurbanovich.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -13,6 +13,52 @@
     />
   </head>
   <body>
-    <h1>Stars</h1>
+    <div class="stars stars--0">
+      <div class="stars__star"></div>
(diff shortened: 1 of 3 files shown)


src/index.html
```diff
@@ -13,6 +13,52 @@
     />
   </head>
   <body>
-    <h1>Stars</h1>
+    <div class="stars stars--0">
```

DanilWeda (changes requested):
Great job! 
Please fix minor comment) 
Thanks)

BudnikOleksii (approved):
Good job 👍

PR #4556, opened 2024-06-15
Stars

timohahaaa:
- [DEMO LINK](https://timohahaaa.github.io/layout_stars/)
- [TEST REPORT LINK](https://timohahaaa.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,1 +1,44 @@
-/* add styles here */
+body {
+  margin: 0;
+}
+
+.stars {
+  display: flex;
+}
+
+.stars:not(:first-child) {
+  padding-top: 4px;
+}
+
+.stars__star {
+  background-image: url(images/star.svg);
+  height: 13px;
+  width: 13px;
+
+  background-size: contain;
+  background-repeat: no-repeat;
+}
+
+.stars__star:not(:last-child) {
+  margin-right: 8px;
+}
+
+.stars--1 .stars__star:nth-child(1) {
+  background-image: url(images/star-active.svg);
+}
+
+.stars--2 .stars__star:nth-child(-n + 2) {
+  background-image: url(images/star-active.svg);
+}
+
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -1 +1,40 @@
...
+  background-image: url(images/star-active.svg);
+}
+
+.stars--4 .stars__star:nth-child(-n + 4) {
+  background-image: url(images/star-active.svg);
+}
+
+.stars--5 .stars__star:nth-child(-n + 5) {
+  background-image: url(images/star-active.svg);
+}
```
volodymyr-soltys97, line 40: These selectors must be separated by commas

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's make your code better

src/style.css
timohahaaa: When you put commas between them, it makes no sense, cause the code starts to work in another way. Or do you mean that I should also change the way how I wrote it??

timohahaaa: oh btw I think I also can delete stars__star at all and leave only stars--n.

src/style.css
```diff
@@ -1 +1,40 @@
...
+
+.stars--4 :nth-child(-n + 4) {
+  background-image: url(images/star-active.svg);
+}
+
+.stars--5 :nth-child(-n + 5) {
+  background-image: url(images/star-active.svg);
+}
```
VitaliyBondarenko1982, line 40:
```suggestion
.stars--1 :nth-child(1),
.stars--2 :nth-child(-n + 2),
.stars--3 :nth-child(-n + 3),
.stars--4 :nth-child(-n + 4),
.stars--5 :nth-child(-n + 5) {
  background-image: url(images/star-active.svg);
}
```

it can be shorter

VitaliyBondarenko1982 (approved):
Looks good.

PR #4557, opened 2024-06-15
add solution

coldshein:
[DEMO LINK](https://colsdhein.github.io/layout_stars/)
[TEST REPORT LINK](https://coldshein.github.io/layout_stars/report/html_report/)

 [x] Code follows all the [Code Style Rules ❗️](https://github.com/Teollan/layout_stars/blob/master/checklist.md)


src/index.html
```diff
@@ -13,6 +13,47 @@
     />
   </head>
   <body>
-    <h1>Stars</h1>
+    <div class="stars stars--0">
```

DanilWeda (changes requested):
All great. 
Check small minor comments and please fix the demo link ( have 404 error) 
Thanks)

src/index.html
coldshein: ok, got it, thanks!

package-lock.json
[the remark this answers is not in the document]
coldshein: Thanks, got it!

PR #4561, opened 2024-06-17
add task solution

YaroslavKolesnyk:
- [DEMO LINK](https://YaroslavKolesnyk.github.io/layout_stars/)
- [TEST REPORT LINK](https://YaroslavKolesnyk.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -13,6 +13,52 @@
     />
   </head>
   <body>
-    <h1>Stars</h1>
+    <div class="stars--0 container">
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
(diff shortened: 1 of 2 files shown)


src/index.html
```diff
@@ -13,6 +13,52 @@
     />
   </head>
   <body>
-    <h1>Stars</h1>
+    <div class="stars--0 container">
```
Moroz-Dmytro, line 16:
```suggestion
    <div class="stars--0 stars">
```

It should be `starts` block, the same in other places

src/style.css
```diff
@@ -1 +1,28 @@
 /* add styles here */
```
Moroz-Dmytro, line 1: Don't leave comments in code

Moroz-Dmytro (changes requested):
Please, fix your DEMO link, it is broken 😢

PR #4569, opened 2024-06-18
create stars layout

0YehorVolkov0:
- [DEMO LINK](https://0YehorVolkov0.github.io/layout_stars/)
- [TEST REPORT LINK](https://0YehorVolkov0.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -13,47 +13,47 @@
     />
   </head>
   <body>
-    <li class="stars stars--0">
-      <div class="stars__star"></div>
-      <div class="stars__star"></div>
-      <div class="stars__star"></div>
-      <div class="stars__star"></div>
-      <div class="stars__star"></div>
-    </li>
-    <li class="stars stars--1">
-      <div class="stars__star"></div>
-      <div class="stars__star"></div>
-      <div class="stars__star"></div>
-      <div class="stars__star"></div>
-      <div class="stars__star"></div>
-    </li>
-    <li class="stars stars--2">
-      <div class="stars__star"></div>
-      <div class="stars__star"></div>
-      <div class="stars__star"></div>
-      <div class="stars__star"></div>
-      <div class="stars__star"></div>
-    </li>
-    <li class="stars stars--3">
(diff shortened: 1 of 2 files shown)


src/index.html
```diff
@@ -13,6 +13,47 @@
     />
   </head>
   <body>
-    <h1>Stars</h1>
+    <stars class="stars stars--0">
```
DanilWeda, line 16: Please use HTML tags. In this case, it is a better way to use a list element.

src/style.css
```diff
@@ -1 +1,37 @@
...
+  height: 16px;
+  width: 16px;
+  margin: 0 4px 0 0;
+  background-repeat: no-repeat;
+  background-position: center;
+}
+
+.stars--1 .stars__star:first-child {
+  background-image: url(images/star-active.svg);
+}
```
DanilWeda, line 21:
```suggestion
.stars--1 .stars__star:first-child,
.stars--2 .stars__star:nth-child(-n + 2) {
  background-image: url(images/star-active.svg);
}
```
You can unite selectors

DanilWeda (changes requested):
Good. 
Check some small comments. 
Many thanks)

src/index.html
```diff
@@ -13,6 +13,47 @@
     />
   </head>
   <body>
-    <h1>Stars</h1>
+    <li class="stars stars--0">
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
```
BudnikOleksii, line 21: `li` should be a child of the `ul` or `ol` element. https://developer.mozilla.org/en-US/docs/Web/HTML/Element/ul
So `stars` should be `ul` or `ol` element and `stars__star` - `li`

BudnikOleksii (changes requested):
Good work 👍 
Just one moment should be fixed

mrkovaliv (approved):
Nice work 👍

PR #4571, opened 2024-06-18
stars task 

piyopiyo0:
[DEMO LINK](https://piyopiyo0.github.io/layout_stars/)
[TEST REPORT LINK ](https://piyopiyo0.github.io/layout_stars/report/html_report/)

 [x]Yellow stars are added with container modifier + pseudo-selector (NO extra classes)
 [x]Each BEM block has its own separate file
 [x]All Typical Mistakes from BEM lesson theory are checked.
 [x]Code follows all the [Code Style Rules ❗️](https://github.com/mate-academy/layout_stars/blob/master/checklist.md)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,1 +1,27 @@
-/* add styles here */
+* {
+  margin: 0;
+}
+
+.stars {
+  display: flex;
+}
+
+.stars__star {
+  background-image: url(images/star.svg);
+  width: 16px;
+  height: 16px;
+  background-repeat: no-repeat;
+  background-position: center;
+}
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -1 +1,27 @@
-/* add styles here */
+* {
+  margin: 0;
+}
```
BudnikOleksii, line 3: Don't set margin to all elements, reset only what you actually need

BudnikOleksii (approved):
Good work 👍 
Approved, but check the comment as well

PR #4574, opened 2024-06-18
Stars

victoriasokolvak:
- [DEMO LINK](https://victoriasokolvak.github.io/layout_stars/)
- [TEST REPORT LINK](https://victoriasokolvak.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -13,6 +13,47 @@
     />
   </head>
   <body>
-    <h1>Stars</h1>
+    <div class="stars stars--0">
+      <div class="stars__star"></div>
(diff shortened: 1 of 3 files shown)


src/index.html
```diff
@@ -13,6 +13,47 @@
     />
   </head>
   <body>
-    <h1>Stars</h1>
+    <div class="stars stars--0">
```

DanilWeda (changes requested):
Good job. 
Check a small comment. 
Thanks)

ArturErmolenko (approved):
Well done 🔥

PR #4577, opened 2024-06-18
Stars task

MaxBorovique:
- [DEMO LINK](https://MaxBorovique.github.io/layout_stars/)
- [TEST REPORT LINK](https://MaxBorovique.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -13,6 +13,47 @@
     />
   </head>
   <body>
-    <h1>Stars</h1>
+    <div class="stars stars--0">
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+    </div>
+    <div class="stars stars--1">
+      <div class="stars__star"></div>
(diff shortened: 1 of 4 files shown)


src/index.html
```diff
@@ -13,6 +13,52 @@
...
   </head>
   <body>
-    <h1>Stars</h1>
+    <div class="stars stars--0">
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+    </div>
```
ArturErmolenko, line 22: <div class="stars__star"></div>
This element is repeated, so in this case it is better to use the **ul li** construction

ArturErmolenko (changes requested):
Great job 🔥 
Just a small improvement, let's use` ul li `

src/index.html
MaxBorovique: I`ll try to remember. Thx.

Moroz-Dmytro (approved):
LGTM 🥇

PR #4609, opened 2024-06-26
Finish task

WhyCame:
If you want to feel like a loser, just know that I spent 2 days doing an entire site out of design because I wasn't paying attention. When I got it completely done - I paid attention and saw that all I had to do was make stars  😆


etojeDenys (approved):
))0)

PR #4653, opened 2024-07-16
Develop

Yatsomo:
- [DEMO LINK](https://Yatsomo.github.io/layout_stars/)
- [TEST REPORT LINK](https://Yatsomo.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,1 +1,44 @@
-/* add styles here */
+body {
+  margin: 0;
+  padding: 0;
+  box-sizing: border-box;
+}
+
+.stars {
+  padding: 0;
+  margin: 0;
+  display: flex;
+  list-style: none;
+}
+
+.stars__star {
+  width: 16px;
+  height: 16px;
+  background-image: url(/images/star.svg);
+  background-repeat: no-repeat;
+  background-position: center;
+}
+
+.stars__star:not(:last-child) {
+  margin-right: 4px;
+}
+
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -1 +1,44 @@
...
+  background-image: url(./images/star-active.svg);
+}
+
+.stars--4 .stars__star:nth-child(-n + 4) {
+  background-image: url(./images/star-active.svg);
+}
+
+.stars--5 .stars__star:nth-child(-n + 5) {
+  background-image: url(./images/star-active.svg);
+}
```
natalia-klonowska, line 44: you can shorten it to:
```suggestion
.stars--1 .stars__star:nth-child(-n + 1),
.stars--2 .stars__star:nth-child(-n + 2),
.stars--3 .stars__star:nth-child(-n + 3),
.stars--4 .stars__star:nth-child(-n + 4),
.stars--5 .stars__star:nth-child(-n + 5) {
  background-image: url(./images/star-active.svg);
}
```

natalia-klonowska (changes requested):
Great work especially your use of list elements 😃

src/style.css
Yatsomo: I've shortened it and all is well! Do You have any additional tips?

natalia-klonowska: for now I don't but in future task you will use scss instead of css and reuse these stars so then you can try to create loop ;)

natalia-klonowska (approved):

PR #4656, opened 2024-07-18
add task solution

TomaszKrzywicki:
## Checklist

- [DEMO LINK](https://tomaszkrzywicki.github.io/layout_stars/)
- [TEST REPORT LINK](https://tomaszkrzywicki.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/styles/blocks/body.css
+++ b/src/styles/blocks/body.css
@@ -1,6 +1,4 @@
-body {
+* {
   margin: 0;
   padding: 0;
-  width: 96px;
-  height: 96px;
 }
--- a/src/styles/blocks/stars.css
+++ b/src/styles/blocks/stars.css
@@ -3,21 +3,16 @@
 }
 
 .stars__star {
-  display: flex;
   width: 16px;
   height: 16px;
   background-image: url('/src/images/star.svg');
-  background-position: center;
   background-repeat: no-repeat;
-  margin-right: auto;
+  background-position: center;
+  margin-right: 4px;
 }
 
 .stars__star.active {
   background-image: url('/src/images/star-active.svg');
-}
-
-.stars__star:last-child {
-  margin-right: auto;
 }
 
 .stars--0 .stars__star:nth-child(-n + 0),
@@ -27,6 +22,4 @@
 .stars--4 .stars__star:nth-child(-n + 4),
 .stars--5 .stars__star:nth-child(-n + 5) {
   background-image: url('/src/images/star-active.svg');
-  background-position: center;
-  margin: auto;
 }
--- a/package.json
+++ b/package.json
@@ -23,7 +23,7 @@
     "@mate-academy/backstop-config": "latest",
     "@mate-academy/bemlint": "latest",
     "@mate-academy/linthtml-config": "latest",
-    "@mate-academy/scripts": "^1.7.3",
+    "@mate-academy/scripts": "^1.8.5",
     "@mate-academy/stylelint-config": "latest",
     "backstopjs": "6.2.2",
     "jest": "^29.7.0",


src/style.css
```diff
@@ -1 +1,36 @@
...
+  display: flex;
+  position: relative;
+}
+
+.stars__star {
+  display: flex;
+  width: 16px;
+  height: 16px;
+  background-image: url('images/star.svg');
+  background-size: cover;
```
natalia-klonowska, line 16: don't use cover because as you can see star doesn't touch edges of it's container instead try to center it's position 
image

natalia-klonowska (changes requested):
you forgot to add margin between star elements

src/style.css
TomaszKrzywicki: I have done new commit however it still doesn't look like on test file ;( it's close but not perfect :)

src/styles/blocks/stars.css
```diff
@@ -0,0 +1,29 @@
+.stars {
+  display: flex;
+  position: center;
+}
+
+.stars__star {
```
Zibi95, line 6: What is see that you background-images does reapeat. There is property to avoid that `background-repeat: no-repeat;`.

src/styles/blocks/stars.css
```diff
@@ -0,0 +1,29 @@
+.stars {
+  display: flex;
+  position: center;
```
Zibi95, line 3: remove this property as it is invalid for this value.

src/styles/blocks/stars.css
```diff
@@ -0,0 +1,32 @@
...
+}
+
+.stars__star {
+  display: flex;
+  width: 16px;
+  height: 16px;
+  background-image: url('/src/images/star.svg');
+  background-position: center;
+  background-repeat: no-repeat;
+  margin-right: auto;
```
natalia-klonowska, line 12: don't use all of this styles with margin auto and instead set specific margins

src/styles/blocks/body.css
```diff
@@ -0,0 +1,6 @@
+body {
+  margin: 0;
+  padding: 0;
+  width: 96px;
+  height: 96px;
```
natalia-klonowska, line 5: [checklist](https://github.com/mate-academy/layout_stars/blob/master/checklist.md#css): Don't fix container size (if there is no such a requirement)

src/styles/blocks/stars.css
```diff
@@ -0,0 +1,32 @@
...
+
+.stars--0 .stars__star:nth-child(-n + 0),
+.stars--1 .stars__star:nth-child(-n + 1),
+.stars--2 .stars__star:nth-child(-n + 2),
+.stars--3 .stars__star:nth-child(-n + 3),
+.stars--4 .stars__star:nth-child(-n + 4),
+.stars--5 .stars__star:nth-child(-n + 5) {
+  background-image: url('/src/images/star-active.svg');
+  background-position: center;
+  margin: auto;
```
natalia-klonowska, line 31: you only need here background-image

src/styles/blocks/stars.css
```diff
@@ -0,0 +1,32 @@
...
+  height: 16px;
+  background-image: url('/src/images/star.svg');
+  background-position: center;
+  background-repeat: no-repeat;
+  margin-right: auto;
+}
+
+.stars__star.active {
+  background-image: url('/src/images/star-active.svg');
+}
```
natalia-klonowska, line 17: you don't have active class

natalia-klonowska (changes requested):

BogdanMaliuta (approved):
Solution is okay, but why you didn`t remove .active styles, you have no elements with this class

PR #4674, opened 2024-07-26
Develop

nex1994:
- [DEMO LINK](https://nex1994.github.io/layout_stars/)
- [TEST REPORT LINK](https://nex1994.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/star--0.css
+++ b/src/star--0.css
@@ -1,0 +1,10 @@
+.stars--0 .stars__star:nth-child(-n + 5) {
+  background-image: url(/src/images/star.svg);
+  background-size: 13px;
+  background-position: center;
+  background-repeat: no-repeat;
+}
+
+/* .stars--0 {
+  margin-right: 4px;
+} */
--- a/src/style.css
+++ b/src/style.css
@@ -1,1 +1,22 @@
-/* add styles here */
+.page {
+  margin: 0;
+  display: flex;
+  flex-direction: column;
+  justify-content: center;
+  align-items: center;
+}
+
+.stars {
+  display: flex;
+  align-items: center;
+}
+
+.stars__star {
+  height: 16px;
+  width: 16px;
+  margin-right: 4px;
+}
+
+/* .stars__star:not(:last-child) {
+  padding-right: 4px;
+} */
--- a/readme.md
+++ b/readme.md
@@ -26,7 +26,7 @@ Follow
 
 ❗️ Copy this `Checklist` to the `Pull Request` description after links, and put `- [x]` before each point after you checked it.
 
-- [ ] Yellow stars are added with container modifier + pseudo-selector (NO extra classes)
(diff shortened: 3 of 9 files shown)


src/star--0.css
```diff
@@ -0,0 +1,10 @@
+.stars--0 .stars__star:nth-child(-n + 5) {
+  background-image: url(/src/images/star.svg);
+  background-size: 13px;
+  background-position: center;
+  background-repeat: no-repeat;
```
Zibi95, line 5: You can just put this into `stars_star` selector. And then just overwrite the `background-image` property with active star image in selector with `nth-child` pseudo class.

src/star--0.css
Zibi95, line 1: Each **block** should have its own file. This one is **modifier** of block so it shouldn't have separate block.

src/style.css
```diff
@@ -1 +1,22 @@
-/* add styles here */
+.page {
+  margin: 0;
+  display: flex;
+  flex-direction: column;
+  justify-content: center;
+  align-items: center;
+}
```
Zibi95, line 7: You can style body and html elements with tag selector without problems.

Zibi95 (changes requested):
Hey. There are some things to improve. There is a lot of repeated code. You only need like around 25-27 lines of code for styles to finish this task. Try to achieve it. GL

src/stars.css
```diff
@@ -0,0 +1,75 @@
...
+
+/* #endregion */
+
+/* #region stars--5 */
+
+.stars--5 .stars__star:nth-child(-n + 5) {
+  background-image: url(/src/images/star-active.svg);
+}
+
+/* #endregion */
```
BogdanMaliuta, line 75: you are in al that selectors using same css rule, just combine this selectors to avoid code duplications, also remove regions after that

BogdanMaliuta (changes requested):
Nor bad, but remove code duplicating here

src/stars.css
```diff
@@ -0,0 +1,29 @@
...
+  background-position: center;
+  background-repeat: no-repeat;
+}
+
+.stars--0 .stars__star:nth-child(-n + 5),
+.stars--1 .stars__star:nth-last-child(-n + 4),
+.stars--2 .stars__star:nth-last-child(-n + 3),
+.stars--3 .stars__star:nth-last-child(-n + 2),
+.stars--4 .stars__star:nth-last-child(-n + 1) {
+  background-image: url(/src/images/star.svg);
```
Zibi95, line 20: You can just move this property to `.stars_star selector`. It would be then overtwritten by the below selectors with `nth-child` bcs it has bigger specificity.

Zibi95 (approved):
Good job. Left a comment for you what I would change and why. But this solution is okay as it meets all the req 👍

PR #4684, opened 2024-07-29
layout_stars

iteamfox:
- [DEMO LINK](https://iteamfox.github.io/layout_stars/)
- [TEST REPORT LINK](https://iteamfox.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,1 +1,49 @@
 /* add styles here */
+body {
+  margin: 0;
+  padding: 0;
+}
+
+.main {
+  display: flex;
+  flex-direction: column;
+  justify-content: space-between;
+}
+
+.stars {
+  display: flex;
+  justify-content: center;
+}
+
+.stars__star {
+  background-image: url('images/star.svg');
+  width: 16px;
+  height: 16px;
+  background-repeat: no-repeat;
+  background-position: center;
+  margin-right: 4px;
+}
+
+.stars--0 .stars__star:nth-child(-n + 0) {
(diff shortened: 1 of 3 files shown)


src/style.css
Zibi95, line 1: Each BEM block has its own separate file. So you should create stars.css and move all the selectors related to stars to that file, and main.css as you created additional block, what i propose is to remove the main block.

src/style.css
```diff
@@ -1 +1,49 @@
...
+  background-image: url('images/star-active.svg');
+}
+
+.stars--4 .stars__star:nth-child(-n + 4) {
+  background-image: url('images/star-active.svg');
+}
+
+.stars--5 .stars__star:nth-child(-n + 5) {
+  background-image: url('images/star-active.svg');
+}
```
Zibi95, line 49: The selectors can be written in a more compact form:

```suggestion
.stars--0 .stars__star:nth-child(-n + 0),
.stars--0 .stars__star:nth-child(-n + 1),
.stars--0 .stars__star:nth-child(-n + 2),
.stars--0 .stars__star:nth-child(-n + 3),
.stars--0 .stars__star:nth-child(-n + 4),
.stars--0 .stars__star:nth-child(-n + 5) {
background-image: url('images/star-active.svg');
}

```

Zibi95 (changes requested):
Hey. One req is not met. Move blocks to its own file and we are good to go 👍

Zibi95 (approved):
Good job!

PR #4702, opened 2024-08-06
Styles added

CamilleHaus:
- [DEMO LINK](https://camillehaus.github.io/layout_stars/)
- [TEST REPORT LINK](https:/camillehaus.github.io/layout_stars/report/html_report/)

- [ x] Each BEM block has its own separate file

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -13,7 +13,7 @@
     />
   </head>
   <body>
-    <section class="stars--0 section">
+    <section class="stars--0 stars">
       <div class="stars__star"></div>
       <div class="stars__star"></div>
       <div class="stars__star"></div>
@@ -21,7 +21,7 @@
       <div class="stars__star"></div>
     </section>
 
-    <section class="stars--1 section">
+    <section class="stars--1 stars">
       <div class="stars__star"></div>
       <div class="stars__star"></div>
       <div class="stars__star"></div>
@@ -29,7 +29,7 @@
       <div class="stars__star"></div>
     </section>
 
-    <section class="stars--2 section">
+    <section class="stars--2 stars">
       <div class="stars__star"></div>
       <div class="stars__star"></div>
       <div class="stars__star"></div>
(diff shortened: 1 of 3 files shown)


src/index.html
```diff
@@ -13,6 +13,52 @@
     />
   </head>
   <body>
-    <h1>Stars</h1>
+    <section class="stars--0 section">
```
pedro-ruas, line 16: When we create a _Block_ with a _Modifier_ we need to pass both names to the class.
In that manner, the correct usage here would be:
```
<section class="stars stars--0">
```
There is also no need for the class "section", you can replace it with the _Block_ name "stars".
Please apply this change for all blocks in this file

src/styles/section.css
```diff
@@ -0,0 +1,9 @@
+.section {
```
pedro-ruas, line 1: After the other changes, you will need to rename this to "stars"

src/styles/stars.css
```diff
@@ -0,0 +1,44 @@
...
+
+.stars--3 .stars__star:nth-child(-n + 3) {
+  background-image: url('../images/star-active.svg');
+}
+
+.stars--4 .stars__star:nth-child(-n + 4) {
+  background-image: url('../images/star-active.svg');
+}
+
+.stars--5 .stars__star:nth-child(-n + 5) {
```
pedro-ruas, line 42: All these operations add the same styling, so you can join then with a comma:
```
.stars--1 .stars__star:nth-child( -n + 1),
.stars--2 .stars__star:nth-child( -n + 2),
.stars--3 .stars__star:nth-child( -n + 3),
.stars--4 .stars__star:nth-child( -n + 4),
.stars--5 .stars__star {
    background-image: url('../images/star-active.svg');
}
```

src/styles/stars.css
```diff
@@ -0,0 +1,44 @@
...
+
+.stars__star:not(:last-child) {
+  margin-right: 4px;
+}
+
+.stars__star:nth-last-child(-n + 5) {
+  margin-bottom: 0;
+}
+
+.stars--0 .stars__star:nth-child(-n + 0) {
```
pedro-ruas, line 22: This style does not apply anywhere, because (-n + 0) don't exist, so you can remove it

pedro-ruas (changes requested):
Soma improvements can be done and there is one BEM mistake to correct

pedro-ruas (approved):
Nice job!!

PR #4741, opened 2024-08-14
Stars

D-Ptashevskyi:
- [DEMO LINK](https://D-Ptashevskyi.github.io/layout_stars/)
- [TEST REPORT LINK](https://D-Ptashevskyi.github.io/layout_stars/report/html_report/)

- [ x ] Yellow stars are added with container modifier + pseudo-selector (NO extra classes)
- [ x ] Each BEM block has its own separate file
- [ x ] All `Typical Mistakes` from `BEM` lesson theory are checked.
- [ x ] Code follows all the [Code Style Rules ❗️](./checklist.md)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,1 +1,30 @@
-/* add styles here */
+body {
+  margin: 0;
+  padding: 0;
+}
+
+.stars {
+  display: flex;
+  list-style: none;
+  margin: 0;
+  padding: 0;
+}
+
+.stars__star {
+  background-image: url(./images/star.svg);
+  background-repeat: no-repeat;
+  background-position: center;
+  margin-right: 4px;
+  width: 16px;
+  height: 16px;
+}
+
+.stars--1 :nth-child(-n + 1),
+.stars--2 :nth-child(-n + 2),
+.stars--3 :nth-child(-n + 3),
+.stars--4 :nth-child(-n + 4),
+.stars--5 :nth-child(-n + 5) {
+  background: url(./images/star-active.svg);
+  background-repeat: no-repeat;
+  background-position: center;
+}
--- a/readme.md
+++ b/readme.md
@@ -21,8 +21,8 @@ Follow
 
 ❗️ Replace `<your_account>` with your Github username and copy the links to `Pull Request` description:
 
-- [DEMO LINK](https://<your_account>.github.io/layout_stars/)
-- [TEST REPORT LINK](https://<your_account>.github.io/layout_stars/report/html_report/)
+- [DEMO LINK](https://D-Ptashevskyi.github.io/layout_stars/)
+- [TEST REPORT LINK](https://D-Ptashevskyi.github.io/layout_stars/report/html_report/)
 
 ❗️ Copy this `Checklist` to the `Pull Request` description after links, and put `- [x]` before each point after you checked it.
 
--- a/src/index.html
+++ b/src/index.html
@@ -13,6 +13,52 @@
     />
   </head>
   <body>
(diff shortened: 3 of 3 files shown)


src/style.css
```diff
@@ -1 +1,30 @@
...
+  display: flex;
+  list-style: none;
+  margin: 0;
+  padding: 0;
+}
+
+.stars__star {
+  background-image: url(./images/star.svg);
+  background-repeat: no-repeat;
+  background-position: center;
```
VolodymyrKirichenko, line 16: u can use background instead(https://developer.mozilla.org/en-US/docs/Web/CSS/background)

src/style.css
```diff
@@ -1 +1,30 @@
...
+}
+
+.stars--1 :nth-child(-n + 1),
+.stars--2 :nth-child(-n + 2),
+.stars--3 :nth-child(-n + 3),
+.stars--4 :nth-child(-n + 4),
+.stars--5 :nth-child(-n + 5) {
+  background: url(./images/star-active.svg);
+  background-repeat: no-repeat;
+  background-position: center;
```
VolodymyrKirichenko, line 29: you already use these properties on your class(`stars__star`), do you need to duplicate it?

VolodymyrKirichenko (changes requested):
Good work, look at my comments and fix the display of the DEMO link and tests.

D-Ptashevskyi:
when i try to delete this lines from the modifier, the picture breaks(

background-repeat: no-repeat;
background-position: center;


ср, 14 авг. 2024 г. в 14:44, VolodymyrKirichenko ***@***.***>:

src/style.css
D-Ptashevskyi: In both cases we need the picture to work properly, so of course we need these lines and it is not an unnecessary duplicate

src/style.css
```diff
@@ -1 +1,30 @@
...
+  display: flex;
+  list-style: none;
+  margin: 0;
+  padding: 0;
+}
+
+.stars__star {
+  background: url(./images/star.svg);
+  background-repeat: no-repeat;
+  background-position: center;
```
VolodymyrKirichenko, line 16: u can use just `background` instead ur properties
https://developer.mozilla.org/en-US/docs/Web/CSS/background

src/style.css
```diff
@@ -1 +1,30 @@
...
+}
+
+.stars {
+  display: flex;
+  list-style: none;
+  margin: 0;
+  padding: 0;
+}
+
+.stars__star {
```
VolodymyrKirichenko, line 13: the last star should have `margin: 0`

src/style.css
```diff
@@ -1 +1,30 @@
...
+  background: url(./images/star.svg);
+  background-repeat: no-repeat;
+  background-position: center;
+  margin-right: 4px;
+  width: 16px;
+  height: 16px;
+}
+
+.stars--1 :nth-child(-n + 1),
```
VolodymyrKirichenko, line 22:
```suggestion
.stars--1 .stars__star:nth-child(-n + 1),
```
etc...

src/style.css
VolodymyrKirichenko: yes, but we already have that behavior above on `stars__star` class)
u can do it without duplicated. Leave it for the end, edit all the comments, and try

VolodymyrKirichenko (changes requested):
well done, but let's make your code better and fix couple of mistakes

BudnikOleksii (approved):
Good work 👍

PR #4743, opened 2024-08-14
add task solution

Hobudo23:
- [DEMO LINK](https://Hobudo23.github.io/layout_stars/)
- [TEST REPORT LINK](https://Hobudo23.github.io/layout_stars/report/html_report/)

CheckList

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,1 +1,32 @@
-/* add styles here */
+body {
+  margin: 0;
+  padding: 0;
+}
+
+.stars {
+  display: flex;
+  margin: 0;
+  padding: 0;
+}
+
+.stars__star {
+  display: flex;
+  height: 16px;
+  width: 16px;
+  margin-right: 4px;
+
+  background-image: url(images/star.svg);
+  background-repeat: no-repeat;
+  background-position: center;
+  list-style: none;
+}
+
+.stars--1 :nth-child(-n + 1),
+.stars--2 :nth-child(-n + 2),
+.stars--3 :nth-child(-n + 3),
+.stars--4 :nth-child(-n + 4),
+.stars--5 :nth-child(-n + 5) {
+  background-image: url(images/star-active.svg);
+  background-repeat: no-repeat;
+  background-position: center;
+}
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -1 +1,32 @@
...
+
+.stars__star {
+  display: flex;
+  height: 16px;
+  width: 16px;
+  margin-right: 4px;
+
+  background-image: url(images/star.svg);
+  background-repeat: no-repeat;
+  background-position: center;
```
Moroz-Dmytro, line 20: You can use short version - https://developer.mozilla.org/ru/docs/Web/CSS/background

src/style.css
```diff
@@ -1 +1,32 @@
...
+}
+
+.stars--1 :nth-child(-n + 1),
+.stars--2 :nth-child(-n + 2),
+.stars--3 :nth-child(-n + 3),
+.stars--4 :nth-child(-n + 4),
+.stars--5 :nth-child(-n + 5) {
+  background-image: url(images/star-active.svg);
+  background-repeat: no-repeat;
+  background-position: center;
```
Moroz-Dmytro, line 31:
```suggestion
```

Not needed to add same styles one more time

src/style.css
```diff
@@ -1 +1,32 @@
...
+  display: flex;
+  margin: 0;
+  padding: 0;
+}
+
+.stars__star {
+  display: flex;
+  height: 16px;
+  width: 16px;
+  margin-right: 4px;
```
Moroz-Dmytro, line 16: Last star should not have `margin-right`

Moroz-Dmytro (changes requested):
Good work 👍 
Some small comments below

src/style.css
Hobudo23: > Last star should not have `margin-right`

only last children in .stars--5?

src/style.css
```diff
@@ -1 +1,36 @@
...
+  list-style: none;
+  background: center no-repeat url(images/star.svg);
+}
+
+.stars--0 :last-child,
+.stars--1 :last-child,
+.stars--2 :last-child,
+.stars--3 :last-child,
+.stars--4 :last-child,
+.stars--5 :last-child {
```
Moroz-Dmytro, line 26:
```suggestion
.stars :last-child {
```

Moroz-Dmytro (approved):
👍

PR #4755, opened 2024-08-16
add task solution

VolodymyrV7:
х Yellow stars are added with container modifier + pseudo-selector (NO extra classes)
 х Each BEM block has its own separate file
х  All Typical Mistakes from BEM lesson theory are checked.
 х Code follows all the Code Style Rules


etojeDenys (approved):
good job

PR #4756, opened 2024-08-16
add task solution

davicatarino:
- [DEMO LINK](https://davicatarino.github.io/layout_stars/)
- [TEST REPORT LINK](https://davicatarino.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,1 +1,28 @@
-/* add styles here */
+body {
+  margin: 0;
+}
+
+.stars {
+  display: flex;
+}
+
+.stars__star {
+  background-image: url('images/star.svg');
+  background-position: center;
+  background-repeat: no-repeat;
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -1 +1,28 @@
...
+  width: 16px;
+  height: 16px;
+  margin-right: 4px;
+}
+
+.stars :last-child {
+  margin: 0;
+}
+
+.stars--1 :nth-child(-n + 1),
```
pedro-ruas, line 22: It would be better to specify the child with `.stars--1 .stars__star:nth-child(-n + 1),`, omitting `stars__star` will apply this to `*:nth-child` .
Just a good practice

pedro-ruas (approved):
Nice! I left a comment for future improvement, but good job!

PR #4774, opened 2024-08-22
add task solution

Menwell:
- [DEMO LINK](https://Menwell.github.io/layout_stars/)
- [TEST REPORT LINK](https://Menwell.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,1 +1,27 @@
 /* add styles here */
+body {
+  margin: 0;
+  padding: 0;
+}
+
+.stars {
+  display: flex;
+}
+
+.stars__stars {
+  content: '';
+  background-image: url(images/star.svg);
+  background-repeat: no-repeat;
+  background-position: center;
+  width: 16px;
+  height: 16px;
+  margin-right: 4px;
+}
+
+.stars--1 .stars__stars:nth-child(-n + 1),
+.stars--2 .stars__stars:nth-child(-n + 2),
+.stars--3 .stars__stars:nth-child(-n + 3),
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -1 +1,27 @@
 /* add styles here */
```
volodymyr-soltys97, line 1:
```suggestion
```

src/style.css
```diff
@@ -1 +1,27 @@
...
+body {
+  margin: 0;
+  padding: 0;
```
volodymyr-soltys97, line 4: The body hasn't a padding
```suggestion
```

src/style.css
```diff
@@ -1 +1,27 @@
...
+}
+
+.stars {
+  display: flex;
+}
+
+.stars__stars {
+  content: '';
```
volodymyr-soltys97, line 12: Why do you use content here?
```suggestion
```

src/style.css
```diff
@@ -1 +1,27 @@
...
+  background-image: url(images/star.svg);
+  background-repeat: no-repeat;
+  background-position: center;
```
volodymyr-soltys97, line 15: Try to combine these only with `background` property 
image

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code
Need to remove margin right for the last stars 
image

src/style.css
Menwell: I corrected the errors, do you need to write about it? Or do you see my new commit for yourself?

lerastarynets (approved):
You're doing amazing, keep it up!

PR #4822, opened 2024-09-10
stars

KomodomFS:
- [DEMO LINK](https://KomodomFS.github.io/layout_stars/)
- [TEST REPORT LINK](https://KomodomFS.github.io/layout_stars/report/html_report/)


pedro-ruas (changes requested):
Ivo, looks good, but I see you did not pass any test.
The idea of using modifiers to light up the stars was done very well, but you also need to pass the styling test, there is spacing to fix, probably background-repeat to fix also.

pedro-ruas (changes requested):
Ivo, I still see failing tests, looks like a problem in spacing between each star and in the whole block.
Be sure to double check it and re-run your testing

KomodomFS:
Probably need to change the bg prosition to center.

pedro-ruas (approved):

PR #4828, opened 2024-09-13
add solution

osidanych:
- [DEMO LINK](https://osidanych.github.io/layout_stars/)
- [TEST REPORT LINK](https://osidanych.github.io/layout_stars/report/html_report/)
-[x] Yellow stars are added with container modifier + pseudo-selector (NO extra classes)
-[x] Each BEM block has its own separate file
-[x] All Typical Mistakes from BEM lesson theory are checked.
-[x] Code follows all the [Code Style Rules ❗️](https://github.com/osidanych/layout_stars/blob/master/checklist.md)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -13,6 +13,47 @@
     />
   </head>
   <body>
-    <h1>Stars</h1>
+    <div class="stars stars--0">
+      <div class="star"></div>
+      <div class="star"></div>
+      <div class="star"></div>
+      <div class="star"></div>
+      <div class="star"></div>
+    </div>
+    <div class="stars stars--1">
+      <div class="star"></div>
+      <div class="star"></div>
+      <div class="star"></div>
+      <div class="star"></div>
+      <div class="star"></div>
+    </div>
+    <div class="stars stars--2">
+      <div class="star"></div>
+      <div class="star"></div>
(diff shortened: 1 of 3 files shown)


src/index.html
```diff
@@ -13,6 +13,52 @@
     />
   </head>
   <body>
-    <h1>Stars</h1>
+    <div class="stars stars--0">
+      <div class="star"></div>
+      <div class="star"></div>
+      <div class="star"></div>
+      <div class="star"></div>
+      <div class="star"></div>
```
volodymyr-soltys97, line 21: You need to fix classes everywhere 
image
```suggestion
      <div class="stars__star"></div>
      <div class="stars__star"></div>
      <div class="stars__star"></div>
      <div class="stars__star"></div>
      <div class="stars__star"></div>
```

src/style.css
```diff
@@ -1 +1,29 @@
 /* add styles here */
```
volodymyr-soltys97, line 1:
```suggestion
```

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's fix classes everywhere

volodymyr-soltys97 (approved):
Great work 🔥

PR #4833, opened 2024-09-14
Stars

Dominik-Dudek-code:
- [DEMO LINK](https://Dominik-Dudek-code.github.io/layout_stars/)
- [TEST REPORT LINK](https://Dominik-Dudek-code.github.io/layout_stars/report/html_report/)


src/stars.css
```diff
@@ -0,0 +1,37 @@
...
+  background-image: url('images/star-active.svg');
+}
+
+.stars--4 > :nth-child(-n + 4) {
+  background-image: url('images/star-active.svg');
+}
+
+.stars--5 > :nth-child(-n + 5) {
+  background-image: url('images/star-active.svg');
+}
```
natalia-klonowska, line 37: you can combine it :D 
```suggestion
.stars--1 > :nth-child(-n + 1), 
.stars--2 > :nth-child(-n + 2),
.stars--3 > :nth-child(-n + 3),
.stars--4 > :nth-child(-n + 4),
.stars--5 > :nth-child(-n + 5) {
  background-image: url('images/star-active.svg');
}
```

src/stars.css
```diff
@@ -0,0 +1,37 @@
...
+  flex-direction: row;
+}
+
+.stars__star {
+  background-image: url('images/star.svg');
+  background-size: cover;
+  background-repeat: no-repeat;
+  height: 12px;
+  width: 13px;
+  margin: 2px 5px 2px 2px;
```
natalia-klonowska, line 12: it would be better to just follow design parameters :) 
change height and width to 16px and margin only to margin right 4px. 
Additionally you need to center the image so delete `background-size: cover` to keep original size and add `background-position: center` to center it 
image

natalia-klonowska (changes requested):
your demo and test raport links are not working so you need to make deploy :)

src/stars.css
Dominik-Dudek-code: I didn't know this could be done like that, now it looks even smoother! 👍

src/stars.css
Dominik-Dudek-code: I thought that I had to set the size of the stars to the same size as in the svg file, i.e. 13/12, because when I set the size to 16/16 it cut off a piece of the star arm, probably due to the lack of "background position".. now it works correctly!

natalia-klonowska (approved):
Great work 🔥

PR #4885, opened 2024-10-02
add task solution

GuilhermeSant:
Mesmo com as medidas do figma continua informando diferença no posicionamento.

Podem me auxiliar de como resolvo por favor ?

### DIFF
--- a/src/styles/blocks/main.css
+++ b/src/styles/blocks/main.css
@@ -1,0 +1,52 @@
+main {
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  height: 100vh;
+}
+
+.container {
+  display: flex;
+  flex-direction: column;
+  justify-content: center;
+  contain: content;
+  width: 96px;
+  height: 116px;
+  padding-top: 22px;
+  padding-left: 22px;
+  padding-bottom: 22px;
+}
+
+.stars {
+  display: flex;
+  justify-content: center;
+}
+
+.stars__star {
+  background-image: url(../../images/star-active.svg);
+  width: 16px;
+  height: 16px;
+  background-position: center;
+  padding-left: 4px;
+  background-repeat: no-repeat;
+}
+
+.stars--0 .stars__star {
+  background-image: url(../../images/star.svg);
+}
+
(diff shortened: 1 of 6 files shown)


pedro-ruas (changes requested):
Guilherme, I can see the code structure looks right, but you did not pass the tests.
Taking a look at the results, the problem seems to be the position of the container block itself. You should check it out.
The AI mentor has pointed out possible issues, regarding paddings margins and overall size limiting of the container that were not requested in the designs.
Also, all your stars have a padding-left, even the first one, which also causes the whole line to move to the right...
All your errors are positioning errors like this one.
First, try removing double size margins where not asked, like bottom and top margins on the same element.
Then remove this left padding from the first star and test again. Probably there will be less errors

src/styles/blocks/main.css
```diff
@@ -0,0 +1,50 @@
...
+.stars--2 .stars__star:nth-child(n + 3) {
+  background-image: url(../../images/star.svg);
+}
+
+.stars--3 .stars__star:nth-child(n + 4) {
+  background-image: url(../../images/star.svg);
+}
+
+.stars--4 .stars__star:nth-child(n + 5) {
+  background-image: url(../../images/star.svg);
```
jv-aquino, line 49: you could comprise all of these selectors in a single line - try to think of a way to do that

src/styles/blocks/main.css
```diff
@@ -0,0 +1,50 @@
...
+  height: 100vh;
+}
+
+.container {
+  display: flex;
+  flex-direction: column;
+  justify-content: center;
+  contain: content;
+}
+
```
jv-aquino, line 14: the above lines are unnecessary and may be crashing the tests

jv-aquino (changes requested):
There are still errors in your tests, I think it's because of the extra css - which would be realllly great in a normal project, but in this one it ends up messing the tests upklkkkkkk
Try to only style the start without any extra positioning

{PT} se ainda assim não for, manda msg no chat ou outro review que a gente aprova

PR #4938, opened 2024-10-15
add task solution

AsyaDev14:
- [x] Code follows all the [Code Style Rules ❗️](https://github.com/AsyaDev14/layout_stars/blob/master/checklist.md)


src/style.css
```diff
@@ -1 +1,124 @@
...
+  background-image: url('./images/star-active.svg');
+}
+
+.stars--4 .stars__star:nth-child(-n + 4) {
+  background-image: url('./images/star-active.svg');
+}
+
+.stars--5 .stars__star:nth-child(-n + 5) {
+  background-image: url('./images/star-active.svg');
+}
```
VolodymyrKirichenko, line 124: u don't need to duplicate it if u have the same properties for all cases. Change it
```suggestion
.stars--1 .stars__star:nth-child(1),
.stars--2 .stars__star:nth-child(-n + 2),
.stars--3 .stars__star:nth-child(-n + 3),
.stars--4 .stars__star:nth-child(-n + 4),
.stars--5 .stars__star:nth-child(-n + 5) {
  background-image: url(images/star-active.svg);
}
```

src/style.css
```diff
@@ -1 +1,124 @@
...
+  align-items: flex-start;
+  list-style: none;
+  font-family: Arial, sans-serif;
+}
+
+.stars__star {
+  width: 16px;
+  height: 16px;
+  background-image: url('./images/star.svg');
+  background-repeat: no-repeat;
```
VolodymyrKirichenko, line 98: add `background-position: center` and use only the `background` property instead of ur three
[background-doc](https://developer.mozilla.org/en-US/docs/Web/CSS/background)

VolodymyrKirichenko (changes requested):
Looks good, but let's improve ur code 👍

src/style.css
```diff
@@ -1 +1,104 @@
...
+b,
+u,
+i,
+center,
+dl,
+dt,
+dd,
+ol,
+ul,
+li,
```
Moroz-Dmytro, line 49: But better would be to reset styles only for tags that you really use in your code

src/style.css
```diff
@@ -1 +1,104 @@
...
+.stars {
+  display: flex;
+  list-style: none;
+}
+
+.stars__star {
+  width: 16px;
+  height: 16px;
+  background: url('./images/star.svg') center no-repeat;
+  margin-right: 4px;
```
Moroz-Dmytro, line 95: Last star in the list should not have `margin-right`

src/index.html
```diff
@@ -1,18 +1,62 @@
...
+      </ul>
+
+      <ul class="stars stars--5">
+        <li class="stars__star"></li>
+        <li class="stars__star"></li>
+        <li class="stars__star"></li>
+        <li class="stars__star"></li>
+        <li class="stars__star"></li>
+      </ul>
+
```
Moroz-Dmytro, line 59:
```suggestion
```

Moroz-Dmytro (changes requested):
Good work 👍 
But fix my comments below

VolodymyrKirichenko (approved):
Well done 🔥

PR #5065, opened 2024-11-17
add task solution

GabrielNunes-383:
- [DEMO LINK](https://GabrielNunes-383.github.io/layout_stars/)
- [TEST REPORT LINK](https://GabrielNunes-383.github.io/layout_stars/report/html_report/)


src/stars--n.css
```diff
@@ -0,0 +1,17 @@
...
+.stars--4 > .stars_star:nth-child(1),
+.stars--4 > .stars_star:nth-child(2),
+.stars--4 > .stars_star:nth-child(3),
+.stars--4 > .stars_star:nth-child(4),
+.stars--5 > .stars_star:nth-child(1),
+.stars--5 > .stars_star:nth-child(2),
+.stars--5 > .stars_star:nth-child(3),
+.stars--5 > .stars_star:nth-child(4),
+.stars--5 > .stars_star:nth-child(5) {
+  background-image: url(images/star-yellow.png);
```
raulriato, line 16: you should use the svg images that were on the project, instead of the images you added.

src/stars.css
```diff
@@ -0,0 +1,4 @@
+.stars {
+  display: flex;
+  justify-content: center;
```
raulriato, line 3: Not a mistake, just a reminder, you don't need to use this justify-content on this project. It's not wrong, you simply don't need it.

src/stars_star.css
```diff
@@ -0,0 +1,7 @@
+.stars_star {
+  background-image: url(images/star-grey.png);
```
raulriato, line 2: Again, it's better to use the svg given by the project. The .png does not work on the tests

src/stars_star.css
```diff
@@ -0,0 +1,7 @@
...
+  width: 16px;
+  height: 16px;
+  display: flex;
+  margin-right: 20px;
```
raulriato, line 6: This margin is wrong, Figma shows a margin of 4px between stars.

Also, you should prevent the last star from having that margin, using `stars__star:last-child`

src/stars_star.css
```diff
@@ -0,0 +1,7 @@
+.stars_star {
```
raulriato, line 1: following the BEM methodology, You need two underscores (`__`), not just one.

raulriato (changes requested):
Your code is fine, but need some changes.

PR #5172, opened 2024-12-18
Develop

ochepurna:
- [DEMO LINK](https://ochepurna.github.io/layout_stars/)
- [TEST REPORT LINK](https://ochepurna.github.io/layout_stars/report/html_report/)


src/stars--N.css
```diff
@@ -0,0 +1,15 @@
...
+.stars--2 .stars__star:nth-child(-n + 2),
+.stars--3 .stars__star:nth-child(-n + 3),
+.stars--4 .stars__star:nth-child(-n + 4),
+.stars--5 .stars__star:nth-child(-n + 5) {
+  background-image: url(images/star-active.svg);
+}
+
+.stars--5 {
+  margin-bottom: 0;
+}
```
etojeDenys, line 15: we don't need that

src/style.css
```diff
@@ -1 +1,22 @@
-/* add styles here */
+body {
+  margin: 0;
+}
+
+.stars {
+  display: flex;
+  margin-bottom: 0;
+  margin-top: 0;
```
etojeDenys, line 8: and this too

etojeDenys (approved):

PR #5249, opened 2025-01-11
add solution

MartaKurdylo:
- [DEMO LINK](https://MartaKurdylo.github.io/layout_stars/)
- [TEST REPORT LINK](https://MartaKurdylo.github.io/layout_stars/report/html_report/)
- Yellow stars are added with container modifier + pseudo-selector (NO extra classes)
-  Each BEM block has its own separate file
-  All Typical Mistakes from BEM lesson theory are checked.
- Code follows all the Code Style Rules

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,1 +1,47 @@
 /* add styles here */
+body {
+  margin: 0;
+  padding: 0;
+  box-sizing: border-box;
+}
+
+.stars {
+  display: flex;
+  gap: 4px;
+}
+
+.stars__star {
+  width: 16px;
+  height: 16px;
+  background-position: center;
+  background-image: url(./images/star.svg);
+  background-repeat: no-repeat;
+}
+
+.stars__star:last-child {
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -1 +1,45 @@
...
+  background-image: url('./images/star-active.svg');
+}
+
+.stars--1 .stars__star:nth-child(-n + 1) {
+  background-image: url('./images/star-active.svg');
+}
+
+.stars--2 .stars__star:nth-child(-n + 2) {
+  background-image: url('./images/star-active.svg');
+}
```
etojeDenys, line 31:
```suggestion
.stars--1 .stars__star:nth-child(-n + 1),
.stars--2 .stars__star:nth-child(-n + 2) {
  background-image: url('./images/star-active.svg');
}
```
combine it all with a comma

etojeDenys (changes requested):

PR #5350, opened 2025-01-30
add task solution

Olehgoon093:
- [DEMO LINK](https://Olehgoon093.github.io/layout_stars/)
- [TEST REPORT LINK](https://Olehgoon093.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,1 +1,54 @@
-/* add styles here */
+body {
+  margin: 0;
+}
+
+.stars {
+  display: flex;
+}
+
+.stars__star {
+  width: 16px;
+  height: 16px;
+  background-image: url(images/star.svg);
+  background-repeat: no-repeat;
+  margin-right: 4px;
+}
+
+.stars--1 > .stars__star:first-child {
+  width: 16px;
+  height: 16px;
+  background-image: url(images/star-active.svg);
+  background-repeat: no-repeat;
+  margin-right: 4px;
+}
+
+.stars--2 > .stars__star:nth-child(-n + 2) {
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -1 +1,54 @@
...
+  margin-right: 4px;
+}
+
+.stars--5 > .stars__star:nth-child(-n + 5) {
+  width: 16px;
+  height: 16px;
+  background-image: url(images/star-active.svg);
+  background-repeat: no-repeat;
+  margin-right: 4px;
+}
```
alexander-ignatow, line 54: Too much duplication here
1. There's no need to set width and height for each style, once it is already set for the `stars__star`
2. You can style `.stars--1..5 > .stars__star:nth-child...` with a single expression

```suggestion
.stars__star {
  width: 16px;
  height: 16px;
  background-image: url(images/star.svg);
  background-repeat: no-repeat;
  margin-right: 4px;
}

.stars--1 > .stars__star:first-child,
.stars--2 > .stars__star:nth-child(-n + 2),
.stars--3 > .stars__star:nth-child(-n + 3) ,
.stars--4 > .stars__star:nth-child(-n + 4) ,
.stars--5 > .stars__star:nth-child(-n + 5) {
  background-image: url(images/star-active.svg);
  background-repeat: no-repeat;
  margin-right: 4px;
}
```

alexander-ignatow (changes requested):
Good job, but please, consider fixing some issues

1. See the comments in the code
2. If it won't take too much time try to fix you code so it passes the test

PR #5465, opened 2025-03-05
layout-stars is done

Timur-Velichko:


src/style.css
```diff
@@ -1 +1,30 @@
-/* add styles here */
+* {
+  box-sizing: border-box;
+}
+
+body {
+  margin: 0;
+  display: block;
```
volodymyr-soltys97, line 7: Default value
```suggestion
```

volodymyr-soltys97 (changes requested):
Hi, there is no demo link. Also you need to add demo link to the PR description 
image

etojeDenys (approved):
lgtm

PR #5575, opened 2025-04-03
stars

DenisNosov22:
- [DEMO LINK](https://DenisNosov22.github.io/layout_stars/)
- [TEST REPORT LINK](https://DenisNosov22.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,1 +1,38 @@
-/* add styles here */
+/* #region page */
+.page__body {
+  margin: 0;
+
+  display: flex;
+
+  width: 96px;
+  height: 116px;
+  flex-direction: column;
+  align-items: center;
+  justify-content: center;
+}
+
+/* #endregion */
+
+/* #region stars */
+.stars {
+  display: flex;
+  justify-content: center;
+  margin: 0;
+  padding: 0;
+}
+
+.stars__star {
+  margin: 2px;
+  width: 16px;
+  height: 16px;
+  border-radius: 0.5px;
+  background-image: url('./images/star.svg');
+  background-size: cover;
+  display: inline-block;
+}
+
+.stars__star--active {
+  background-image: url('./images/star-active.svg');
+}
+
+/* #endregion */
--- a/readme.md
+++ b/readme.md
@@ -22,8 +22,8 @@ Hold `Alt` key (`Option` on MacOS) to measure distances in Figma.
 
 ❗️ Replace `<your_account>` with your Github username and copy the links to `Pull Request` description:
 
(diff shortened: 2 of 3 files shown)


src/style.css
```diff
@@ -1 +1,38 @@
...
+/* #region stars */
+.stars {
+  display: flex;
+  justify-content: center;
+  margin: 0;
+  padding: 0;
+}
+
+.stars__star {
+  margin: 2px;
```
AndrewSRGN, line 25: .stars__star:not(:last-child) {
  margin-right: 4px;
}
You make a margin of 2 px on all sides. You only need to margin between the stars

src/style.css
```diff
@@ -1 +1,38 @@
...
+  width: 16px;
+  height: 16px;
+  border-radius: 0.5px;
+  background-image: url('./images/star.svg');
+  background-size: cover;
```
AndrewSRGN, line 30: The star.svg image is 16px by 16px. All you need to do is background-repeat: no-repeat;
delete background-size: cover;
delete border-radius: 0.5px;

src/style.css
```diff
@@ -1 +1,38 @@
...
+  display: inline-block;
```
AndrewSRGN, line 31: The “display: inline-block;” property is redundant. You are using a div block with the default “display: block;” property, but the parent of this block has the “display: flex” property, which by default has the “flex-direction: row” property, which will arrange the child elements in a row

src/style.css
```diff
@@ -1 +1,38 @@
...
+}
+
+.stars__star--active {
```
AndrewSRGN, line 34: You must use a pseudo class.
For example:
.stars--1 .stars__star:nth-child(-n + 1) {
}

src/style.css
```diff
@@ -1 +1,38 @@
...
+.page__body {
+  margin: 0;
+
+  display: flex;
+
+  width: 96px;
+  height: 116px;
+  flex-direction: column;
+  align-items: center;
+  justify-content: center;
```
AndrewSRGN, line 11: display: flex;
   width: 96px;
   height: 116px;
   flex-direction: column;
   align-items: center;
   justify-content: center;
   Are redundant properties. You only need to properly configure the .stars and .stars_star classes

src/style.css
```diff
@@ -1 +1,38 @@
...
+}
+
+/* #endregion */
+
+/* #region stars */
+.stars {
+  display: flex;
+  justify-content: center;
```
AndrewSRGN, line 19: justify-content: center; is reduntant
Your block with class .stars will adjust to the size of child elements and they will follow each other, it remains to add margin for child elements

PR #5780, opened 2025-06-04
Develop

danie010830:
- [DEMO LINK](https://danie010830.github.io/layout_stars/)
- [TEST REPORT LINK](https://danie010830.github.io/layout_stars/report/html_report/)


src/style.css
```diff
@@ -1 +1,46 @@
...
+  background-image: url('./images/star.svg');
+  background-repeat: no-repeat;
+  background-position: center;
+}
+
+.stars__star:last-child {
+  margin-right: 0;
+}
+
+.stars--0 .stars__star:nth-child(-n + 0) {
```
pedro-ruas, line 23: You could join all declarations with the same style using a comma `,`:
```
    background-image: url('./images/star-active.svg');
}
```

pedro-ruas, line 23: And there is no need to style something that won't apply. `.stars--0 .stars__star:nth-child(-n + 0)`, this selector will never select anything (There is no nth-child with index less than 0)

pedro-ruas (changes requested):
Small change, I would approve it, but I think it is worth that you read the comments

PR #5781, opened 2025-06-05
add task solution

Nika-FD:
- [DEMO LINK](https://Nika-FD.github.io/layout_stars/)
- [TEST REPORT LINK](https://Nika-FD.github.io/layout_stars/report/html_report/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -13,7 +13,6 @@
     />
   </head>
   <body>
-    <h1>Stars</h1>
     <div class="stars stars--0">
       <div class="stars__star"></div>
       <div class="stars__star"></div>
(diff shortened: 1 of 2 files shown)


anastasiiavorobiova (changes requested):
Great work!
Make sure that your solution is similar to the design 
Screenshot 2025-06-05 at 18 41 36

etojeDenys (changes requested):
1. fix the tests
2. check if star size is the same

if you have problems, feel free to ask for help in chat

src/index.html
```diff
@@ -14,5 +14,52 @@
   </head>
   <body>
     <h1>Stars</h1>
```
etojeDenys, line 16: you don't need that

etojeDenys (changes requested):
your demo links still showing that you need to pass the tests

PR #5931, opened 2025-07-19
apliquei as soluções de estilizações

NilsonDev:
- [DEMO LINK](https://github.com/NilsonDev/layout_stars.git/)
- [TEST REPORT LINK](https://NilsonDev.github.io/layout_stars/report/html_report/)

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -2,7 +2,7 @@
 
 Implement the [Stars Block](https://www.figma.com/file/ojkArVazq7vsX0nbpn9CxZ/Moyo-%2F-Catalog-(ENG)?node-id=11325%3A2960) used in a card and catalog.
 
-Hold `Alt` key (`Option` on MacOS) to measure distances in Figma. 
+Hold `Alt` key (`Option` on MacOS) to measure distances in Figma.
 
 > Here are the [Layout Tasks Instructions](https://mate-academy.github.io/layout_task-guideline)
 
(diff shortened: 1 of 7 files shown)


readme.md
```diff
@@ -22,8 +22,8 @@ Hold `Alt` key (`Option` on MacOS) to measure distances in Figma.
 
 ❗️ Replace `<your_account>` with your Github username and copy the links to `Pull Request` description:
 
-- [DEMO LINK](https://<your_account>.github.io/layout_stars/)
-- [TEST REPORT LINK](https://<your_account>.github.io/layout_stars/report/html_report/)
+- [DEMO LINK](https://github.com/NilsonDev/layout_stars.git/)
```
raulriato, line 25: You should not change the url structure. you should just replace `<your_account>` by `NilsonDev`, making the url `https://NilsonDev.github.io/layout_stars/`

src/style.css
```diff
@@ -1 +1,56 @@
...
+  background-image: url('./images/star-active.svg');
+}
+
+.stars--5 .stars__star:nth-child(-n + 5) {
+  background-image: url('./images/star-active.svg');
+}
+
+.stars--0:last-child {
+  margin-right: 0;
+}
```
raulriato, line 56: I think this style is not working as you intended it to.

When you use `.stars--0:last-child`, the style you code to it is applied directly to the last `<div>` thaat has the `class="stars--0"`, not to the last `.star` that is a "child" of the `.stars--0`.

raulriato (approved):
Good job! Just left some notes there for next projects

PR #6094, opened 2025-08-30
add task solution

Mirlas123:
- [DEMO LINK](https://Mirlas123.github.io/layout_stars/)
- [TEST REPORT LINK](https://Mirlas123.github.io/layout_stars/report/html_report/)


src/images/Cover_photo.png
vitalii-fedusov, line 1: why we need images from landing in stars repository?

vitalii-fedusov (changes requested):
image
please remove margin-right for last star in row

TarasHoliuk (approved):

PR #6399, opened 2025-12-10
stranger things

HokkaidoDev:
- [DEMO LINK](https://hokkaidodev.github.io/layout_stars/)
- [TEST REPORT LINK](https://hokkaidodev.github.io/layout_stars/report/html_report/)

Each BEM block has its own separate file. What's that mean exactly? CSS files for every BEM block? Can you clarify?
I also have a problem with deployment. This link [DEMO LINK](https://hokkaidodev.github.io/layout_stars/) was deployed successfully, but the page is empty. What is the reason it can be? [TEST REPORT LINK] is working very well. Every test was passed.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -2,7 +2,7 @@
 
 Implement the [Stars Block](https://www.figma.com/file/ojkArVazq7vsX0nbpn9CxZ/Moyo-%2F-Catalog-(ENG)?node-id=11325%3A2960) used in a card and catalog.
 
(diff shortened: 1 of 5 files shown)


HokkaidoDev:
@mateacademy-ai-mentor Do I need to redeploy the project after my changes using the npm run deploy command? The page is still invisible by [DEMO LINK](https://hokkaidodev.github.io/layout_stars/).
And why should we leave the last line empty in our .css or .html files?

PR #6419, opened 2025-12-20
add stars

DenisNosov22:
- [DEMO LINK](https://DenisNosov22.github.io/layout_stars/)
- [TEST REPORT LINK](https://DenisNosov22.github.io/layout_stars/report/html_report/)


src/style.css
```diff
@@ -1 +1,42 @@
...
+  background-image: url(images/star-active.svg);
+}
+
+.stars--4 .stars__star:nth-child(-n + 4) {
+  background-image: url(images/star-active.svg);
+}
+
+.stars--5 .stars__star:nth-child(-n + 5) {
+  background-image: url(images/star-active.svg);
+}
```
2pasha: don't repeat yourself

you can join all this rules like this:
```
...
.stars--2 .stars__star:nth-child(-n + 5) {
}
```

src/style.css
```diff
@@ -1 +1,42 @@
...
+  width: 96px;
+}
+
+.stars {
+  display: flex;
+}
+
+.stars__star {
+  width: 14px;
+  height: 14px;
```
2pasha: in figma i can see that stars is 16x16px

2pasha (changes requested):
good job! 

please take a look at improvements bellow

PR #6475, opened 2026-01-16
Stars

yuriyorobchuk090:


Denys-Kravchuk9988 (changes requested):
Good Job!

Could you please share the Demo Link and the Test Report Link?

As I noticed, the star styles are currently placed in `src/images/src/blocks/stars.css`, which makes them a bit hard to discover in the project structure. Could we move them to `src/styles/stars.css` so all styling is kept in one consistent location?

Also, once the styles are imported from `src/styles/stars.css`, the star-related rules in `style.css` become redundant, so we can remove them to avoid duplication.

PR #6749, opened 2026-05-06
stars

Phantom-coder0:
- [DEMO LINK](https://Phantom-coder0.github.io/layout_stars/)
- [TEST REPORT LINK](https://Phantom-coder0.github.io/layout_stars/report/html_report/)


Anton-Kuchmasov (changes requested):
Good work, but please, re-use provided images instead of importing `.png` images by ourself - check out comment below to proceed
It's better to pass all tests because you'll re-use this stars in next module task

Anton-Kuchmasov:
> Good work, but please, re-use provided images instead of importing `.png` images by ourself - check out comment below to proceed It's better to pass all tests because you'll re-use this stars in next module task

In this task you should use `star.svg` and `star-active.svg` which are already provided in the `src/images` dir
image

2pasha (changes requested):
you should not delete `style.css` and use `starts.css` instead

you should in `index.html` still import `style.css` and inside `style.css` you should import `stars.css`

Anton-Kuchmasov (approved):
Good job!

PR #6825, opened 2026-06-07
add task solution

Ahavrush:
https://Ahavrush.github.io/layout_stars/
https://Ahavrush.github.io/layout_stars/report/html_report/


brespect (approved):
Well done!

PR #6840, opened 2026-06-15
add task solution

yukie-kameoka:
- [DEMO LINK](https://yukie-kameoka.github.io/layout_stars/)
- [TEST REPORT LINK](https://yukie-kameoka.github.io/layout_stars/report/html_report/)


DanielCamposDevX (approved):
The implementations is 100%, but the layout is diverging a little bit for the tests remember to always take a look at the tests and make then all pass

PR #6923, opened 2026-07-21
Develop

skuzmenk:


src/index.html
```diff
@@ -7,12 +7,65 @@
...
+
   <body>
-    <h1>Stars</h1>
+    <div class="stars stars--0">
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+      <div class="stars__star"></div>
+    </div>
```
Moroz-Dmytro, line 29: Would be good to use semantic `u'/li` here

src/stars.css
```diff
@@ -0,0 +1,26 @@
...
+
+.stars__star {
+  width: 16px;
+  height: 16px;
+  margin-right: 4px;
+
+  background-image: url(images/star.svg);
+  background-repeat: no-repeat;
+  background-position: center;
+  background-size: 13px;
```
Moroz-Dmytro, line 13: You can use just one property `background` for that

Moroz-Dmytro (approved):
Looks like you forgot to do deploy

src/index.html
skuzmenk: I used it but Luke said to change it

src/stars.css
skuzmenk: ok

skuzmenk: thanks

